PR #64, opened 2015-03-16
Reduce integration test time.

### LINKED ISSUE #62: Reduce time required to run tests
The tests take a long time to run. Some of this overhead is unavoidable with the integration tests. However, there's probably some low-hanging fruit in ClusterTestHarness that might make a big difference:

* Default number of brokers is 3. This is probably overkill for almost every test.
* Might be able to pull up/tear down brokers in parallel, although it would be worth checking how long they take to start up/tear down before trying to make the parallel version work.
* Schema registry only needs to be started if we're going to use Avro. Only a couple of the tests need Avro support.

ewencp:
Reduce integration test time by using fewer brokers, one most of the time, and
only starting the schema registry when required. Fixes #62.


granders:
Compiles, tests pass and code LGTM

ewencp:
Thanks for the review @granders.

PR #67, opened 2015-03-17
Handle conflicting IDs and separate IDs used in the REST proxy and by Ka...

### LINKED ISSUE #65: Consumer read request hangs if you try to create two consumer instances with the same ID
If you:
1. Create consumer with an ID
2. Read from a topic
3. Create consumer with same ID (don't delete previous one)
4. Read from topic
the second create call succeeds and the read hangs. The stack trace shows it stuck trying to create a zookeeper node in a method that can also handle conflicts, which we'd expect since the IDs are the same.

Turns out that we ended up conflating the meaning of "ID", which is used both to name the consumer nicely so you have readable consumer URLs and it got passed in as the ID for the underlying old Kafka consumer. The latter should really only be happening for debugging purposes since a random UUID (generated automatically for the consumer if the ID is omitted) is way more reliable.

Things get a bit more confusing when we consider moving to the new consumer, which has a separate client ID that it uses to make logging on the brokers clearer and traceable to specific applications.

A fix for this has to be careful about compatibility. I think the right thing to do is add a new parameter ("name") which will be used in the URL. If the "id" is specified and "name" isn't, "name" will default to "id". The underlying Kafka "consumer.id" will only be set if "id" was explicitly set. Then, the name is only relevant to the single instance of the REST proxy since consumer URLs are specific to a single process. We can return a 409 if we see a naming conflict during creation, which will prevent the subsequent hang during read. Finally, we'll want to deprecate and deemphasize the "id" field in the docs since "name" will now be the field they really should be using.

ewencp:
...fka's consumer implementation.

Add a check for duplicate IDs before creating a consumer and adding it to the
map in ConsumerManager. Also add a "name" field to ConsumerInstanceConfig to
allow control over the naming of the instance config within the single REST
proxy without setting the ID of the underlying Kafka consumer, since this can be
useful if a limited set of consumer instances within a group is desired. Update
the documentation to mark the "id" field as deprecated since it probably doesn't
provide the behavior users expect and without this fix it can cause problems if
the user ends up with duplicates. Kafka doesn't expect it to be used this way
except for testing anyway. Also add the name field and update all the example
cURL requests to use the recommended default where both "id" and "name" are
omitted, and consumers are deleted properly when they are no longer needed.

Fixes #65.

@junrao Maybe you could review this? I want to be especially careful here since this affects the existing API rather than just adding to it, so backwards compatibility is particularly important.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,26 +36,34 @@ the REST Proxy running using the default settings and some topics already create
       {"value_schema_id":0,"offsets":[{"partition":0,"offset":0}]}
 
     # Create a consumer for binary data, starting at the beginning of the topic's
-    # log. Then consume some data from a topic.
+    # log. Then consume some data from a topic using the base URL in the first response.
+    # Finally, close the consumer with a DELETE to make it leave the group and clean up
+    # its resources.
     $ curl -X POST -H "Content-Type: application/vnd.kafka.v1+json" \
-          --data '{"id": "my_instance", "format": "binary", "auto.offset.reset": "smallest"}' \
+          --data '{"format": "binary", "auto.offset.reset": "smallest"}' \
           http://localhost:8082/consumers/my_binary_consumer
-      {"instance_id":"my_instance","base_uri":"http://localhost:8082/consumers/my_binary_consumer/instances/my_instance"}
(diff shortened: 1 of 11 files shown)


ewencp:
retest this please

ewencp:
retest this please

ewencp:
@junrao Can you review this?

Issue #72, opened 2015-03-27
Avro producer needs tests for more schema types


ewencp:
See https://groups.google.com/forum/#!topic/confluent-platform/RTc1u6F68JY where plain string keys are failing as key schemas. The cause is really in the serializer (confluentinc/schema-registry#110), but since the REST proxy is doing a bunch of extra translation when it gets the request, we should at least have one test for each schema type, including all the primitive types.

PR #89, opened 2015-06-16
JSON message support

hachikuji:
This request adds support for JSON as an embedded message format. The new embedded content type is "application/vnd.kafka.json.v1+json." A few points worth mentioning:

- To preserve backwards compatibility, the content type "application/json" will still assume the binary embedded format on produce requests. 
- To consume messages with the embedded format, you must use "application/vnd.kafka.json.v1+json" in the Accept header.
- Only UTF-8 encoding of JSON is supported.


ewencp:
add to whitelist

ewencp:
retest this please

ewencp:
retest this please

ewencp:
retest this please

ewencp:
retest this please

src/main/java/io/confluent/kafkarest/serializers/KafkaJsonSerializer.java
```diff
@@ -0,0 +1,50 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.kafka.common.errors.SerializationException;
+import org.apache.kafka.common.serialization.Serializer;
+
+import java.util.Map;
+
+/**
+ * Serialize JsonNode to UTF-8 JSON.
+ */
+public class KafkaJsonSerializer implements Serializer<Object> {
```
ewencp: Would it make sense to just move the serializer implementations into a package in schema-registry, along with the Avro ones? Obviously they're pretty trivial right now and won't get any real schema registry support unless we add JSON support to the schema registry (in which case we'd need to figure out how to pass along that schema info anyway), but is it better to just house them there and depend on them here?

src/test/java/io/confluent/kafkarest/integration/ConsumerJsonTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.List;
+import java.util.Properties;
+
+public class ConsumerJsonTest extends AbstractConsumerTest {
+
+  private static final String topicName = "topic1";
+  private static final String groupName = "testconsumergroup";
+
+  private final List<ProducerRecord<Object, Object>> recordsWithKeys = Arrays.asList(
+      new ProducerRecord<Object, Object>(topicName, "key", "value"),
```
ewencp: Should we test anything besides strings? At least covering a few types with complex structure in one test would be a good idea, e.g. make sure integers, floats, strings, arrays, boolean, and null as fields in an object work.

src/test/java/io/confluent/kafkarest/integration/JsonProducerTest.java
```diff
@@ -0,0 +1,125 @@
...
+        new Properties());
+  }
+
+  @Override
+  protected String getEmbeddedContentType() {
+    return Versions.KAFKA_V1_JSON_JSON;
+  }
+
+  private final List<JsonTopicProduceRecord> topicRecordsWithKeys = Arrays.asList(
+      new JsonTopicProduceRecord("key", "value", 0),
```
ewencp: Same as in consumer test, we should probably test other types. We ran into this issue already (see #72) so it's probably best to make sure we have this covered in the JSON case with this initial commit.

ewencp:
@hachikuji Looks good so far. A couple things are probably missing:

1. Docs. `docs/api.rst` is probably the most important. `docs/intro.rst` should also probably be updated. The quickstart is starting to get overcrowded though, we might need to split some of that off into a sort of user guide and keep the quickstart to be very minimal.

2. A recent PR added simple consumer support. SimpleConsumerManager has the code that should handle that case, and then you just need to chain it back up to PartitionsResource to map the right content type to the EmbeddedFormat.

src/test/java/io/confluent/kafkarest/integration/ConsumerJsonTest.java
hachikuji: Yeah, that's a good idea. I'll update the tests.

src/main/java/io/confluent/kafkarest/serializers/KafkaJsonSerializer.java
hachikuji: If there's a good chance that we'll support JSON schema in the future, then we may as well move it now. Otherwise, I'd just leave it here.

ewencp:
@hachikuji LGTM. One last missing piece is to update the Content Types section of api.rst to include json as an embedded format. After that, feel free to merge.

dalassi:
Hi guys,
I have the latest version deployed (1.0.1). It doesn't support the JSON format without base64. I tried it with the above-mentioned Content-Type (application/vnd.kafka.json.v1+json) but I get an HTTP 415 error, here is the raw HTTP response back:
HTTP/1.1 415 Unsupported Media Type
Content-Length: 62
Content-Type: application/vnd.kafka.v1+json
Server: Jetty(8.1.16.v20140903)

{"error_code":415,"message":"HTTP 415 Unsupported Media Type"}

Is there anything I am missing to get it to work?

ewencp:
@dalassi That's correct, the JSON support has not been released yet. It'll be in the next version of CP (or you can always build the REST proxy yourself if you're comfortable building and using code from master).

dalassi:
Thanks @ewencp  for the quick reply! I don't mind doing that. However, I wasn't able to build using as there were external dependencies that Maven (Common, Rest Utils, and Schema Registry). Would you be able to direct me to  documentation on how to do the build? Thanks!

ewencp:
@dalassi There's a brief [development](https://github.com/confluentinc/kafka-rest#development) section in the README. You basically just have to build all the dependencies in order: (maybe Kafka), confluentinc/common, confluentinc/rest-utils, confluentinc/schema-registry. For each, if you `mvn install`, it will install them in your local Maven repository. 

Then, if you `mvn package` in kafka-rest, it will also build compiled versions with all the necessary jars pulled in, much like our released packages (as well as standalone jars). Those will be under `target/kafka-rest-<version>-<mode>` directories. The ones you'll probably find most useful are the ones ending in  `-development` (pulls in copies of all jars, including transitive dependencies). If you want an uber jar, you can `mvn -Pstandalone package` and it'll be located in `ls target/kafka-rest-<version>-standalone.jar`.

You may not want to build directly off master right now because we've shifted over to relying on the unreleased Kafka 0.9.0.0-SNAPSHOT builds. You may want to pick an earlier version. You'll need to do the same for the schema registry as well.

dalassi:
Cool, thanks Ewen!

PR #101, opened 2015-07-16
rest-utils updates

ijuma:
Updates the documentation and JUnit version.

I am not sure how we want to format deprecations, so this is just something to start the conversation.

Should follow https://github.com/confluentinc/rest-utils/pull/21

### DIFF
--- a/docs/monitoring.rst
+++ b/docs/monitoring.rst
@@ -24,8 +24,9 @@ Global Metrics
   ``connections-active``
     Total number of active TCP connections.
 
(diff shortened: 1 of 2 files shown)


ewencp:
@ijuma I think the way you marked the deprecation is fine -- it'd be nicer if there were just a way to add an annotation and have this maintained automatically, but I don't think there's a good way to do that with metrics.

This LGTM, feel free to merge after merging confluentinc/rest-utils#21.

PR #102, opened 2015-07-22
swagger documentation prototype

jendap:
DO NOT MERGE!!! at least not yet!

See the comment in confluentinc/rest-utils#22 in confluentinc/rest-utils repo.

Try it out and tell me what you think! Thanks!


jendap:
What a cute @ConfluentCLABot !

It should come with a link. Googling "confluent cla" does not find it. I found link in a history of public activity of the cla bot. http://clabot.confluent.io/cla ??? It would be nice to put it back to the message.

BTW: otherwise very good way of doing cla, great github integration

ewencp:
@jendap Odd, the message format used to be different -- it had an explanation and a link as well. I'll check on why this has changed. The link you found is the correct one.

ewencp:
See confluentinc/rest-utils#22

Issue #94, opened 2015-06-29
Rename Main class


ewencp:
This is an inconvenient name since jps just lists it as Main instead of as kafka-rest, and it gets confusing if you're also running schema-registry.

We'll need to maintain the Main class for compatibility, but we should change all usages (e.g. in scripts) to the new class.

PR #103, opened 2015-07-27
Issue 94 rename main

granders:
@ewencp 
Very similar to https://github.com/confluentinc/schema-registry/pull/203


ewencp:
@granders LGTM, thanks for taking care of these.

FYI, if you put "Fixes #X" into commit messages, it'll close them automatically upon merge so, in this case, you wouldn't have to close #94 manually.

PR #104, opened 2015-07-27
Allowing Null Values

scbrady:
Allows value to be null even if a value schema has been defined for the request. This allows someone to have multiple inserts and deletes in the same request.

Currently the only way to have a null value is to not pass a value schema or schema id. If someone wanted to send a request with a large number of records, some records being inserts, some deletes; they could not just pass a key and a null value for the deletes, because a value schema would be defined. This allows them to send a null value for the deletes and still have a value schema in place for the inserts.

### DIFF
--- a/src/main/java/io/confluent/kafkarest/AvroRestProducer.java
+++ b/src/main/java/io/confluent/kafkarest/AvroRestProducer.java
@@ -94,7 +94,7 @@ public class AvroRestProducer implements RestProducer<JsonNode, JsonNode> {
         // if there isn't a schema. Validation will have already checked that all the keys/values
         // were NullNodes.
         Object key = (keySchema != null ? AvroConverter.toAvro(record.getKey(), keySchema) : null);
-        Object value = (valueSchema != null
+        Object value = ((valueSchema != null && !record.getValue().isNull())
                         ? AvroConverter.toAvro(record.getValue(), valueSchema) : null);
         Integer recordPartition = partition;
         if (recordPartition == null) {
--- a/src/main/java/io/confluent/kafkarest/AvroRestProducer.java
+++ b/src/main/java/io/confluent/kafkarest/AvroRestProducer.java
@@ -94,7 +94,7 @@ public class AvroRestProducer implements RestProducer<JsonNode, JsonNode> {
         // if there isn't a schema. Validation will have already checked that all the keys/values
         // were NullNodes.
         Object key = (keySchema != null ? AvroConverter.toAvro(record.getKey(), keySchema) : null);
-        Object value = ((valueSchema != null && !record.getValue().isNull())
+        Object value = (valueSchema != null
                         ? AvroConverter.toAvro(record.getValue(), valueSchema) : null);
         Integer recordPartition = partition;
         if (recordPartition == null) {
--- a/src/main/java/io/confluent/kafkarest/converters/AvroConverter.java
+++ b/src/main/java/io/confluent/kafkarest/converters/AvroConverter.java
@@ -66,6 +66,8 @@ public class AvroConverter {
 
   public static Object toAvro(JsonNode value, Schema schema) {
     try {
+      if (value.isNull())
+         return null;
       ByteArrayOutputStream out = new ByteArrayOutputStream();
       jsonMapper.writeValue(out, value);
       DatumReader<Object> reader = new GenericDatumReader<Object>(schema);


src/main/java/io/confluent/kafkarest/AvroRestProducer.java
```diff
@@ -94,7 +94,7 @@ public void produce(ProduceTask task, String topic, Integer partition,
         // if there isn't a schema. Validation will have already checked that all the keys/values
         // were NullNodes.
         Object key = (keySchema != null ? AvroConverter.toAvro(record.getKey(), keySchema) : null);
-        Object value = (valueSchema != null
```
ewencp: Does `AvroConverter.toAvro` not currently support this? If the schema supports null as a type (i.e. it is always null or is a union with null as one of the types), shouldn't the `toAvro` conversion just work, converting the `null` `JsonNode` to a `null` `Object` value? A unit test would be helpful here.

In any case, it seems like that'd be a better location for this check since it would centralize the conversion and makes it unit-testable. Centralizing seems useful since we would have missed the previous line, which presumably should also get the same treatment.

scbrady:
@ewencp The `AvroConverter.toAvro` function was not working the way I thought it should, but maybe my situation was just very specific. Here is some background:

My key schema was:
```json
{
    "type": "record",
    "name": "PK_HomePageList",
    "fields": [
        {
            "name": "HomePageListId",
            "type": [
                "null",
                "int"
            ]
        }
    ]
}
```

And my value schema was:
```json
{
    "type": "record",
    "name": "HomePageList",
    "fields": [
        {
            "name": "HomePageListId",
            "type": [
                "null",
                "int"
            ]
        },
        {
            "name": "LibraryKey",
            "type": [
                "null",
                "string"
            ]
        },
        {
            "name": "ListId",
            "type": [
                "null",
                "string"
            ]
        },
        {
            "name": "ListType",
            "type": [
                "null",
                "string"
            ]
        },
        {
            "name": "ListName",
            "type": [
                "null",
                "string"
            ]
        },
        {
            "name": "ListJson",
            "type": [
                "null",
                "string"
            ]
        },
        {
            "name": "ListOrder",
            "type": [
                "null",
                "int"
            ]
        },
        {
            "name": "LastUpdateDateUtc",
            "type": [
                "null",
                "string"
            ]
        }
    ]
}
```

It was not letting me make a request like this:
```json
{
    "key_schema_id": 3,
    "value_schema_id": 4,
    "records": [
        {
            "key": {
                "HomePageListId": {
                    "int": 1
                }
            },
            "value": {
                "HomePageListId": {
                    "int": 1
                },
                "LibraryKey": {
                    "string": "test"
                },
                "ListId": {
                    "string": "test"
                },
                "ListType": {
                    "string": "test"
                },
                "ListName": {
                    "string": "test"
                },
                "ListJson": {
                    "string": "test"
                },
                "ListOrder": {
                    "int": 2
                },
                "LastUpdateDateUtc": {
                    "string": "test"
                }
            }
        },
        {
            "key": {
                "HomePageListId": {
                    "int": 1
                }
            },
            "value": null
        }
    ]
}
```

But with my modifications it now works like I expect. I just figured I would submit the pull request in case anyone else had run into an issue like this. But I agree, a better place to move this fix would be in the `toAvro` function.

scbrady:
@ConfluentCLABot When trying to submit a completed CLA, I just get a Gateway Timeout Error.

ewencp:
@scbrady Sorry about that, not sure why it was getting stuck. I bounced the service, you can try submitting again.

scbrady:
[clabot:check]

ewencp:
@scbrady We have a test for passing `null`, and this actually causes it to fail:
```
AvroConverterTest.testPrimitiveTypesToAvro:111 » Conversion Failed to convert ...
```
The issue is that your schemas don't allow using `null`. Both key and value schemas are `record` types. Each of the *fields* can be `null` (via a `union` of the `null` type and another primitive type such as `string`). However, the type of the entire schema is only `record`. If you want to be able to set the entire key or value to `null`, you need to use a `union` schema. For example, your key schema would be:

```
["null", {
    "type": "record",
    "name": "PK_HomePageList",
    "fields": [
        {
            "name": "HomePageListId",
            "type": [
                "null",
                "int"
            ]
        }
    ]
}]
```

This is a common confusion because Avro handles `null` as a separate type (like some programming languages type systems do) rather than using an `optional` flag on schemas to allow the field to be `null`.

Issue #75, opened 2015-04-22
Add TLS support


mastersingh24:
I know we can run kafka-rest behind a secure proxy / load balancer, but in many organizations TLS is also required between the secure proxy / load balancer and the actual endpoint.  Also, in some simple deployments there may not be a front-end gateway / load balancer.  

I believe this would need to be implemented in rest-utils but would need to be exposed here

kjvalencik:
:+1: Although, I would be satisfied with just the ability to listen on a unix socket instead of a TCP port. This would allow a simple nginx proxy to provide TLS without any network traffic.

ewencp:
Fixed by #200

PR #106, opened 2015-08-06
Support kafka-rest.base_uri configuration that defines a custom base URI for all the URLs returned by API

huyphan:
When we run kafka-rest behind an HTTP Web Server (e.g. Nginx or Apache), we would expect it to use the public facing URL instead of the local one. That can be achieved by having this configuration and giving it higher priority than scheme, host, and port configurations.

This PR also partially solved the issue #75: one can setup SSL at the Web Server (which runs in front of Kafka Rest API) and use an URI with `https` scheme for `kafka-rest.base_uri` configuration.

### DIFF
--- a/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -43,6 +43,13 @@ public class KafkaRestConfig extends RestConfig {
       + " hostname is used";
   public static final String HOST_NAME_DEFAULT = "";
 
+  public static final String BASE_URI_CONFIG = "kafka-rest.base_uri";
+  private static final String BASE_URI_DOC =
+      "The base uri used to generate absolute URLs in responses. Defaults to empty. If this value is "
+      + "set, host.name and port configuration will be ignored. This configuration is useful when kafka-rest "
+      + "is running behind another HTTP web server.";
+  public static final String BASE_URI_DEFAULT = "";
+
   public static final String ZOOKEEPER_CONNECT_CONFIG = "zookeeper.connect";
   private static final String
       ZOOKEEPER_CONNECT_DOC =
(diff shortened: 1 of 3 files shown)


huyphan:
[clabot:check]

ewencp:
I think this new setting somewhat duplicates the existing `host.name` setting, but with a different use case that requires overriding the scheme and port as well as the hostname. The goal of the `host.name` setting was to handle cases where the automatically determined FQDN isn't the one you want to use (which can easily happen if two host names resolve to the same machine).

In this case, might it be a better solution to have the URI builder code support `X-Forwarded-*` HTTP headers, and maybe as part of confluentinc/rest-utils since that's pretty generic utility code? I think this is the common practice if you expect to be proxying requests, and it gives more flexibility. For example, you don't have to choose your scheme up front, so you can generate URLs with `https` or `http` based on any policy the proxy can implement. This would also avoid settings that somewhat overlap like this current solution does. We still want `host.name` for the case where you're not proxying, need a fixed host name to use, and the automatically determined hostname is not the one you want to use. But `X-Forwarded-*` headers would still override this and give complete flexibility to the proxy in helping construct absolute URLs.

huyphan:
@ewencp thanks for your comment.

I agree that making use of `X-Forwarded-*` is a better solution, do you have any plan to implement that feature in `confluentinc/rest-utils` ? Proxying kafka-rest behind an HTTP Web Server is a popular use case though. 

I haven't looked at the rest-utils code yet, but I don't mind creating a patch for it if you don't really have a plan yet.

ewencp:
@huyphan Definitely feel free to create a patch, we didn't have an issue filed for this yet. I think it shouldn't be too bad since you already have similar code, it's just a matter of checking the request headers instead of a configuration option.

ewencp:
Closing this since we agreed the `X-Forwarded-*` option is the better solution.

Issue #92, opened 2015-06-22
Docs: Unclear default partitioning behavior


randa1:
When using the POST /topics/{topic} endpoint to produce a message:
1. If record.partition is not explicitly passed, is the record.key used as partitioning key?
2. Which partitioning scheme is used when sending the message to Kafka? Is it the default Kafka partitioner (hash(key)%partitions)? 
3. Is there a way to override the partitioning behavior other than specifying record.partition?

I'm sorry if these questions are trivial, I've couldn't find a conclusive statement in the docs and I may have some terminology gaps as I'm only getting into Kafka and Confluent.

hachikuji:
1. Yes, record.key is the partitioning key in the absence of record.partition.
2. In the absence of record.partition, we delegate partition assignment to KafkaProducer with the default partitioner.
3. Other than exposing record.partition, we don't currently provide a way to change the partition assignment strategy. Do you have a use case where this feature would be helpful?

And I'll take a look at updating the docs to make the partitioning strategy clearer.

randa1:
Regarding #3 - I can't really think of anything right now, record.partition is probably enough.
Thanks for the quick reply.

miguno:
@hachikuji: Have we checked whether the docs could be improved?  If so, can we close this ticket?

hachikuji:
@miguno Sorry, forgot about this. Just opened GH-108 to fix.

PR #108, opened 2015-08-19
Clarify partitioning behavior for produce requests

hachikuji:
This fixes GH-92.


hachikuji:
retest this please

ewencp:
LGTM. The test failure seems unrelated, though a bit concerning that it just suddenly appeared...

My only comment is that partitioning is only round-robin for leaders that are available, right? Or at least that's the behavior on trunk in Kafka? I doubt we need that level of detail, but if there is anything in the regular Kafka docs we could link to that gives the full explanation, that could be a nice addition.

PR #117, opened 2015-10-20
Update to Kafka 0.9.0.0-SNAPSHOT and make adjustments to work with updated ZkUtils.

ewencp:


ewencp:
@hachikuji I know you've at least worked with this code base a bit, mind looking it over to spread the review load a bit? It depends on confluentinc/schema-registry#237 so the PR builder will probably fail until that one lands.

ewencp:
retest this please

ewencp:
retest this please

The failure looks like an unrelated transient issue, filed #119 to follow up on that issue.

ewencp:
retest this please

docs/api.rst
```diff
@@ -89,6 +89,8 @@ per-request basis.
           * Error code 50002 -- Kafka error.
           * Error code 50003 -- Retriable Kafka error. Although the operation failed, it's
             possible that retrying the request will be successful.
+          * Error code 50101 -- Only SSL endpoints were found for the specified broker, but
```
granders: should there be a corresponding tiny doc patch?

ewencp: Not sure I understand. This is a doc patch, right?

src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -86,37 +117,54 @@ public ClusterTestHarness() {
...
+    this.numBrokers = numBrokers;
+    this.withSchemaRegistry = withSchemaRegistry;
+
+    schemaRegProperties = new Properties();
+    restProperties = new Properties();
+  }
 
-    configs = new Vector<Properties>();
-    bootstrapServers = "";
+  public Properties overrideBrokerProperties(int i, Properties props) {
```
granders: Is this method needed?

src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -51,19 +57,45 @@
...
+      return ports;
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * Choose an available port
+   */
+  public static int choosePort() {
```
granders: Is there any danger that successive calls to `choosePort` will return the same port?

It looks like before the change you allocated all ports at once, and then handed them out from that pool.

src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -86,37 +117,54 @@ public ClusterTestHarness() {
...
-      Properties props = TestUtils.createBrokerConfig(i, port, false);
-      // Turn auto creation *off*, unlike the default. This lets us test errors that should be
-      // generated when brokers are configured that way.
-      props.put("auto.create.topics.enable", "false");
+      final Option<java.io.File> noFile = scala.Option.apply(null);
+      final Option<SecurityProtocol> noInterBrokerSecurityProtocol = scala.Option.apply(null);
+      Properties props = TestUtils.createBrokerConfig(
+          i, zkConnect, false, false, TestUtils.RandomPort(), noInterBrokerSecurityProtocol,
+          noFile, true, false, TestUtils.RandomPort(), false, TestUtils.RandomPort(), false,
+          TestUtils.RandomPort());
```
granders: Just a comment for my understanding. I think in this pattern, `TestUtils.RandomPort()` returns a thing which signals `createBrokerConfig` to make a call to its own random port method?

src/test/java/io/confluent/kafkarest/integration/ProducerTest.java
```diff
@@ -163,7 +154,7 @@ public void testProducerConfigOverrides() {
     overrides.setProperty("buffer.memory", "1");
     // Note separate ProducerPool since the override should only be for this test, so
     // getProducerPool doesn't work here
-    ProducerPool pool = new ProducerPool(this.restConfig, this.bootstrapServers, overrides);
+    ProducerPool pool = new ProducerPool(this.restConfig, this.brokerList, overrides);
```
granders: Just curiosity - I see you've generally opted to rename `bootstrapServers` as `brokerList`. Any particular reason for the change?

docs/api.rst
granders: Just checking because I forgot which things are a source of autogenerated docs (confluent platform docs for the website), and which have to be manually updated

src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
ewencp: Yes, there is some danger. In particular, we're more likely to run into issues if tests are run in parallel. We changed this in Kafka, but it's harder to fix in these because of how the ports are used -- last I looked into this, there wasn't a convenient way to use a random port and extract the bound port. However, at the moment since we're not using a parallel test runner, this approach works ok in practice.

src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
ewencp: It's a hook for subclasses to allow them to override broker properties only for their tests. See, e.g. `ProducerTopicAutoCreationTest`.

src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
ewencp: Yes, correct. `RandomPort` is just `0`.

src/test/java/io/confluent/kafkarest/integration/ProducerTest.java
ewencp: I think that fell out of copying from schema-registry as a base. Somehow these classes had diverged. I'm happy with either name.

granders:
@ewencp Ok thanks for answering my questions. LGTM

PR #122, opened 2015-11-04
Use x.y.z versioning scheme (i.e. 2.0.0-SNAPSHOT)

miguno:


miguno:
Note: The Jenkins build above fails because the associated PR for the versioning changes of `common`, `rest-utils`, and `schema-registry` aren't merged yet:

https://github.com/confluentinc/common/pull/23
https://github.com/confluentinc/rest-utils/pull/25
https://github.com/confluentinc/schema-registry/pull/243

Because of this the kafka-rest build fails to find the various 2.0.0-SNAPSHOT jar files of its dependencies.

PR #127, opened 2015-11-24
Fix to ConsumerManager tests AND then handling of max bytes

gawth:
Hi,

whilst looking into a bug related to long polling I discovered that the consumer manager tests have a flaw in them.

The tests rely on assertions within a callback method.  However, assertion fails are hidden and the test code continues without reporting a test failure.

I have fixed the tests but left an assert that should clearly fail as a demo of the issue (see testConsumerNormalOps, first test case).

I'm not a Java dev by trade so you may want to provide further guidance on how you'd prefer to restructure the assertions.  My basic approach is to save off the results of the callback to variables that are accessible outside the callback and then carry out assertion tests (as per existing sawCallback checks).

Having fixed the flaw in the tests one of the tests (testConsumerExceptions) now fails.  Not sure if this is a code or a test problem - happy to fix it but need some guidance on what is required.
 
Thanks,
Alan


marekstachura:
:+1:

gawth:
Hi,

I've attempted to sign up to the CLA at http://clabot.confluent.io/cla but I'm getting a "Please try again. Something went wrong" error every time I try

Thanks,
Alan

gawth:
CLA now submitted.
Thanks,
Alan

gawth:
I've updated my fork with changes to handling of max bytes.

Previously the code correctly stopped consuming messages when max bytes was reached but when it came to checking whether to finish processing it failed to take the last message in to consideration.  This resulted in the code waiting for the consumer timeout to expire even though no more messages would be read.

To fix this I have also had to change the MockTime implementation and any timing related tests in ConsumerManagerTests.  The aim of MockTime seems to be to strictly control timer handling however it did not fit with the way ConsumerReadTask worked.  The consumer read task does not use sleep or waitOn (methods in the mock that incremented time) so would never trigger the termination of the loop.

src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
```diff
@@ -104,6 +104,7 @@ public boolean doPartialRead() {
       }
 
       boolean backoff = false;
+      long roughMsgSize = 0;
```
gawth: Moved this outside the loop for use later on

src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
```diff
@@ -116,8 +117,8 @@ public boolean doPartialRead() {
         while (iter.hasNext()) {
           MessageAndMetadata<KafkaK, KafkaV> msg = iter.peek();
           ConsumerRecordAndSize<ClientK, ClientV> recordAndSize = parent.createConsumerRecord(msg);
-          long roughMsgSize = recordAndSize.getSize();
-          if (bytesConsumed + roughMsgSize > maxResponseBytes) {
+          roughMsgSize = recordAndSize.getSize();
+          if (bytesConsumed + roughMsgSize >= maxResponseBytes) {
```
gawth: Changed to use >= to be consistent with check later on

src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
```diff
@@ -143,7 +144,9 @@ public boolean doPartialRead() {
           started + parent.getConfig().getInt(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG);
       waitExpiration = Math.min(backoffExpiration, requestExpiration);
 
-      if (elapsed >= requestTimeoutMs || bytesConsumed >= maxResponseBytes) {
+      // Including the rough message size here ensures processing finishes if the next
+      // message exceeds the maxResponseBytes
+      if (elapsed >= requestTimeoutMs || bytesConsumed + roughMsgSize >= maxResponseBytes) {
```
gawth: Include the check on the roughMsgSize here so that we break out if it has been exceeded.

src/test/java/io/confluent/kafkarest/mock/MockTime.java
```diff
@@ -25,11 +25,18 @@
 
   @Override
   public long milliseconds() {
+    // Incrementing currentMs prevents tests from locking up
+    // "waiting" for the time checks in the code.  Without this
+    // the ConsumerManager never completed (it never calls sleep or waitOn)
+    currentMs++;
```
gawth: Because the ConsumerReadTask only calls the milliseconds method failure to do something with the currentMs results in an infinite loop.  Incrementing here is somewhat arbitrary but at least the code doesn't get stuck forever...

src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
```diff
@@ -172,38 +177,52 @@ public void testConsumerNormalOps() throws InterruptedException, ExecutionExcept
...
-    // the per-request timeout (because the timeout perfectly coincides with a scheduled
-    // iteration when using the default settings).
-    assertEquals(config.getInt(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG)
-                 + config.getInt(KafkaRestConfig.CONSUMER_ITERATOR_TIMEOUT_MS_CONFIG),
-                 config.getTime().milliseconds());
+    // the per-request timeout (attempts to tie this down to the ms not possible so assumes that greater than is "good enough")
+    String msg = "Time taken (" + Long.toString(config.getTime().milliseconds()) + ") to process message should be greater than the timeout " +
+        Integer.toString(config.getInt(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG)
+                 + config.getInt(KafkaRestConfig.CONSUMER_ITERATOR_TIMEOUT_MS_CONFIG)) ;
+    assertFalse(msg, config.getInt(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG)
```
gawth: Because the MockTime now relies on an increment inside the milliseconds method its no longer possible to be exact on time comparison...

src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
```diff
@@ -216,10 +235,10 @@ public void testConsumerMaxBytesResponse() throws InterruptedException, Executio
     // response, not all of it is returned.
     final List<ConsumerRecord<byte[], byte[]>> referenceRecords
         = Arrays.<ConsumerRecord<byte[], byte[]>>asList(
-        new BinaryConsumerRecord(null, new byte[512], 0, 0),
-        new BinaryConsumerRecord(null, new byte[512], 1, 0),
-        new BinaryConsumerRecord(null, new byte[512], 2, 0),
-        new BinaryConsumerRecord(null, new byte[512], 3, 0)
+        new BinaryConsumerRecord(null, new byte[511], 0, 0), // Don't use 512 as this happens to fall on boundary
```
gawth: This test passed when using 512 because the max bytes was 1024.  The code in ConsumerReadTask used a >= to check so 2 messages triggered the exit.  Dropping to 511, two messages should still trigger the exit (because the next message takes it over the max bytes) but the original code didn't exit (and in fact it locked up due to the mock time bug).

gawth:
@ewencp any chance you could take a look at this change? I'd love to get the fix to max bytes in to the main repository so I don't have to run off my fork in production.

Thanks,
Alan

ewencp:
ok to test

src/test/java/io/confluent/kafkarest/mock/MockTime.java
ewencp: This shouldn't be necessary and it actually makes writing tests with MockTime a lot harder to reason about -- the point of MockTime is to let the test get some control over how time proceeds but now it is highly dependent on how many times `milliseconds()` is called. And the inconsistency between `milliseconds()` and `nanoseconds()` is confusing.

What was the test that got hung up when this wasn't incrementing? We probably just need to address the way that test works. The `ConsumerManager` may never directly wait/sleep, but something else should (e.g. consumer iterators).

src/test/java/io/confluent/kafkarest/mock/MockTime.java
```diff
@@ -40,6 +47,10 @@ public void sleep(long ms) {
 
   @Override
   public void waitOn(Object on, long ms) throws InterruptedException {
+    // Introduced a thread.sleep to allow context switching in consumer worker.
+    // Without this the worker appears to get stuck in a busy loop and actually
+    // takes longer  to process the messages...
+    Thread.sleep(1);
```
ewencp: I'm not sure I understand what this is supposed to accomplish. Context switching doesn't rely on `Thread.sleep()` being called.

src/test/java/io/confluent/kafkarest/mock/MockTime.java
gawth: The short answer - you're right, I'll revert the change.  

Before I fixed the max bytes bug this was an issue.  With the maxbytes bug fixed the tests no longer lock up.

src/test/java/io/confluent/kafkarest/mock/MockTime.java
gawth: I'll remove this - don't think its needed now

gawth:
The build failed due to a dependency issue.  I'll fix that and revert the mock time changes and then let you know.

ewencp:
@gawth Sounds good. Re: jenkins failure, it looks like it's an intermittent bug we've been seeing on our Jenkins where it sometimes fails to properly fetch one of the upstream jars. If it passes locally, just let me know and I'll test manually before merging.

gawth:
@ewencp I've made changes as discussed, specifically to revert previous changes to MockTime and then amend tests back to where they were.

So the remaining changes are two fold:
1) Changes to tests to fix a problem with failed assertions being smothered by exceptional handling in callbacks
2) Max Bytes ug fix to ConsumerReadTask 

All tests pass locally.
Thanks,
Alan

src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
```diff
@@ -172,38 +177,50 @@ public void testConsumerNormalOps() throws InterruptedException, ExecutionExcept
     String cid = consumerManager.createConsumer(
         groupName, new ConsumerInstanceConfig(EmbeddedFormat.BINARY));
     sawCallback = false;
+    actualException = null;
```
ewencp: @gawth I've also used a special test base class to handle background exceptions. See https://github.com/apache/kafka/blob/trunk/connect/runtime/src/test/java/org/apache/kafka/connect/util/ThreadedTest.java for example. Would this approach work too so you could just do things like `assertNull(e)` in some of the callbacks and still get them propagated to the main test? The obvious drawback is that the failure becomes a bit disconnected from the origin of the error, but that's true in both cases.

I'm not sure if it is worth the change here since these cases need to record `actualRecords` too, unless we can just keep all assertions in the event handlers if we propagate exceptions back to the main test correctly. What do you think?

gawth: @ewencp What would be reported if a test failed?  If its enough to identify the assert that failed then it is worth a go but if you get what amounts to an unexpected exception in the test I'd probably advise against it.  At least with the current approach you can setup asserts clearly to identify what went wrong.

ewencp: @gawth It should throw an `AssertionError` where the cause is the exception that occurred in the other thread: https://github.com/apache/kafka/blob/trunk/connect/runtime/src/test/java/org/apache/kafka/connect/util/TestBackgroundThreadExceptionHandler.java#L35 That should mean anything that logs the exception stacktrace should also provide the original exception's stack trace, which should make it easy to track down the source of the issue. You do have to scan far enough in the message to see the second stacktrace, but we could also just add a message, something like `"A background thread threw an exception during the test, see the cause exception for details."` to make it obvious what happened.

gawth: @ewencp I've taken a quick look  at the implementation in Kafka of ThreadedTest.

In summary, I think we would need:

- I'll need to create a new abstract class that extends Thread and includes a static UncaughtExceptionHandler.
- ConsumerWorker will need to then extend that new class
- Tests will have to extend ThreadedTest class to swap in the uncaught exception handler for the tests

I did take a look if I could add a few references and then use the ShutdownableThread class from Kafka but that hooks in a bunch of other stuff that I don't think is needed.

So I guess the question becomes - do you want me to do the work to insert that abstract class OR are you happy to go with the tests as is?

Ta,
Alan

gawth:
Hi @ewencp any further feedback on this?  Please see my above response re use of test exception handler.
Thanks,
Alan

ewencp:
retest this please

ewencp:
Sorry about the delay, completely lost track of this PR. LGTM, thanks!

gawth:
No probs - happens to us all.

PR #128, opened 2015-11-26
update kafka-rest-run-class

jacksu:
I hope like other software.- for the letter,-- for the word.


jacksu:
CLA now submitted.

ewencp:
It may not be consistent with many Unix tools, but this approach is consistent with the way java flags work. For kafka-rest and other Confluent tools, we're staying consistent with the approach used by Kafka, which also matches the Java conventions.

PR #130, opened 2015-11-30
REST-116: Catching exception due to NONE from ZkUtils in Kafka.

fpj:
Adding code to catch the exception. My understanding is that we should be adding no partition if we get NONE in the `getLeaderAndIsrForPartition` call, but if that's an error and we should be throwing an exception, then we should at least transform it into a more meaningful message.

### DIFF
--- a/src/main/java/io/confluent/kafkarest/MetadataObserver.java
+++ b/src/main/java/io/confluent/kafkarest/MetadataObserver.java
@@ -180,20 +180,24 @@ public class MetadataObserver {
 
       Partition p = new Partition();
       p.setPartition(partId);
-      LeaderAndIsr leaderAndIsr =
-          zkUtils.getLeaderAndIsrForPartition(topic, partId).get();
-      p.setLeader(leaderAndIsr.leader());
-      scala.collection.immutable.Set<Integer> isr = leaderAndIsr.isr().toSet();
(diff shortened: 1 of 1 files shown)


fpj:
The failure is due to a dependency issue:

```
Failed to execute goal on project kafka-rest: Could not resolve dependencies for project io.confluent:kafka-rest:jar:2.0.0-SNAPSHOT: Could not find artifact io.confluent:rest-utils-test:jar:2.0.0-20151126.125612-28 in jenkins
```

src/main/java/io/confluent/kafkarest/MetadataObserver.java
```diff
@@ -180,20 +181,24 @@ public int getLeaderId(final String topicName, final int partitionId) {
...
+          int broker = (Integer) brokerObj;
+          PartitionReplica
+                  r =
+                  new PartitionReplica(broker, (leaderAndIsr.leader() == broker), isr.contains(broker));
+          partReplicas.add(r);
+        }
+        p.setReplicas(partReplicas);
+        partitions.add(p);
+      } catch (NoSuchElementException e) {
+        log.warn("No leader and ISR information for partition {}", partId);
```
ewencp: It looks like the result of this is that we just omit the partition if it's missing leader and ISR info? Is this what we want to happen wrt REST API output? Or should the leader & ISR be null, but the `Partition` object is still returned?

Issue #105, opened 2015-08-03
Race condition when consuming while topic is being created


kjvalencik:
If you make a call to `POST /consumers/(string: group_name)/instances/(string: instance)/offsets`
while a topic is in the process of still being created you may get an exception due to this line:

https://github.com/confluentinc/kafka-rest/blob/d899a1eb1bc83d67cc9ebe0ec74195b0f11f6346/src/main/java/io/confluent/kafkarest/MetadataObserver.java#L185

The error is that this `Option` is empty.

In practice, this is likely not a concern, but it makes unit testing difficult because consumers may not be started until a topic has fully been created.

hachikuji:
@kjvalencik Good catch. We definitely should be checking the option.

ewencp:
I think there might be a couple of other cases of this same issue in that class as well. Should be easy to find them all with an IDE to search for uses of `Option.get()`

rekhajoshm:
https://github.com/confluentinc/kafka-rest/pull/144

michaelkirschbaum:
Has this been resolved?

ewencp:
@michaelkirschbaum Yes, should be for at least this case since #144 was merged. If you're seeing another case where we're not checking an `Option` properly, please let us know.

PR #144, opened 2015-12-25
KAFKA-REST:105 Fix for Race condition when consuming while topic is being created

rekhajoshm:
#105: Fix for Race condition when consuming while topic is being created

### DIFF
--- a/src/main/java/io/confluent/kafkarest/MetadataObserver.java
+++ b/src/main/java/io/confluent/kafkarest/MetadataObserver.java
@@ -114,16 +114,18 @@ public class MetadataObserver {
     // shouldn't be common, so we just grab all of them to keep this simple
     Map<String, Properties> configs = AdminUtils.fetchAllTopicConfigs(zkUtils);
     for (String topicName : JavaConversions.asJavaCollection(topicNames)) {
-      Map<Object, Seq<Object>> partitionMap = topicPartitions.get(topicName).get();
-      List<Partition> partitions = extractPartitionsFromZKData(partitionMap, topicName, null);
-      if (partitions.size() == 0) {
-        continue;
+      if(!topicPartitions.get(topicName).isEmpty()) {
+        Map<Object, Seq<Object>> partitionMap = topicPartitions.get(topicName).get();
+        List<Partition> partitions = extractPartitionsFromZKData(partitionMap, topicName, null);
+        if (partitions.size() == 0) {
+          continue;
+        }
+        Option<Properties> topicConfigOpt = configs.get(topicName);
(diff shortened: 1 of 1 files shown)


ewencp:
ok to test

src/main/java/io/confluent/kafkarest/MetadataObserver.java
```diff
@@ -180,20 +185,21 @@ public int getLeaderId(final String topicName, final int partitionId) {
...
-      p.setLeader(leaderAndIsr.leader());
-      scala.collection.immutable.Set<Integer> isr = leaderAndIsr.isr().toSet();
-      List<PartitionReplica> partReplicas = new Vector<PartitionReplica>();
-      for (Object brokerObj : JavaConversions.asJavaCollection(part.getValue())) {
-        int broker = (Integer) brokerObj;
-        PartitionReplica
-            r =
-            new PartitionReplica(broker, (leaderAndIsr.leader() == broker), isr.contains(broker));
-        partReplicas.add(r);
+      if(!zkUtils.getLeaderAndIsrForPartition(topic, partId).isEmpty()) {
```
ewencp: In the other cases, you looked up the items twice. When they were just map lookups that's not really a problem since its a fast, local operation. However, in this case you're calling `zkUtils.getLeaderAndIsrForPartition` twice, which requires making a request to zookeeper. And in fact, may return different results if something changes between the two calls.

Could we instead just call it once, store it into a variable of type `Option[LeaderAndIsr]` and then do the `isEmpty`/`get` on that variable?

ewencp:
@rekhajoshm Thanks for the patch! Waiting on tests, but it looks good to me other than the minor comment I left on avoiding making two requests to ZK.

rekhajoshm:
Passes locally.
```
/.m2/repository/io/confluent/kafka-rest/2.0.0-SNAPSHOT/kafka-rest-2.0.0-SNAPSHOT-tests.jar
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 04:07 min
[INFO] Finished at: 2015-12-25T11:50:25-08:00
[INFO] Final Memory: 30M/348M
[INFO] ------------------------------------------------------------------------
```

rekhajoshm:
retest this please

rekhajoshm:
Thanks @ewencp for the review.updated.please check.thanks.

ewencp:
LGTM, tests are passing locally, and the failure on Jenkins was due to a known transient issue with the server. Thanks for the fix, @rekhajoshm!

Issue #132, opened 2015-12-02
KAFKA REST allows me create a consumer with illegal names but can not delete it


lukemao:
Can create a consumer with id: my/Consumer£$%^&*#

The create was successful, however when I delete it I get the error:
java.lang.RuntimeException: java.lang.IllegalArgumentException: URLDecoder: Illegal hex characters in escape (%) pattern - For input string: "^&"
        at com.mashape.unirest.http.HttpClientHelper.prepareRequest(HttpClientHelper.java:186)
        at com.mashape.unirest.http.HttpClientHelper.request(HttpClientHelper.java:134)
        at com.mashape.unirest.request.BaseRequest.asJson(BaseRequest.java:68)

PR #145, opened 2016-01-06
Created patch for Issue #132 - Invalid URL Characters in Consumer Name

KFarmer3:
Created a patch for Issue #132 , added validation for the consumer name and then added a test to check that creating a consumer no longer allows invalid URL Characters in the name/id.


ewencp:
ok to test

ewencp:
@KFarmer3 Seems like something may have gone wrong with the patch -- maybe some issue with the line endings or reformatting or something? It's showing entire files changed which makes it difficult to review the changes. Any ideas what happened?

Issue #154, opened 2016-02-03
GET on consumers with several hundreds topics created 


chrisramond:
After running some load tests using the kafka-rest api , i found responses times from 3s till 30s when trying to get messages from a topic using a the uri of consumer previously created.

25 topics tested among 800 created 
20000 messages of 3,5k posted to each of the 25 topics
A consumer is created after 10000th message is posted for each topic .
The uri returned on the consumer creation is used to get the messages of each topics. ( max_bytes = 1000000 )

Setting the consumer.request.timeout.ms to 100 and consumer.threads to 25 does not change anything.

If the unused topics are deleted ( 25 topics kept only ) , the response time is around 1s.

ewencp:
I think this is due to trying to load all the metadata when checking if the topic exists. I think there are further improvements that could be made (e.g. some caching of topics in `MetadataObserver`), but #202 should have improved this substantially. I'm going to close this out, but if you reevaluate and find the performance is still poor, please reopen and we can work on it from there.

PR #155, opened 2016-02-08
Check that at least one message has been consumed before breaking to …

kjvalencik:
…prevent hanging on messages that are larger than max_bytes

### DIFF
--- a/src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
+++ b/src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
@@ -117,7 +117,7 @@ class ConsumerReadTask<KafkaK, KafkaV, ClientK, ClientV>
           MessageAndMetadata<KafkaK, KafkaV> msg = iter.peek();
           ConsumerRecordAndSize<ClientK, ClientV> recordAndSize = parent.createConsumerRecord(msg);
           long roughMsgSize = recordAndSize.getSize();
-          if (bytesConsumed + roughMsgSize > maxResponseBytes) {
+          if (bytesConsumed > 0 && bytesConsumed + roughMsgSize > maxResponseBytes) {
             break;
           }
 


kjvalencik:
@ewencp Is this something that is acceptable to merge?

ewencp:
@kjvalencik This is a parameter that the normal Java clients actually enforce and refuse to return data if it exceeds the configured size. In fact, brokers are also supposed to filter messages that are too large, such that you need to coordinate settings across producer, broker, and consumer to get larger messages to make it through the full pipleine.

This can potentially be pretty significant. If I configure the broker (or one topic on the broker) to accept larger messages, I can do things like produce a 1GB message. Then consumers that might not be expecting such large messages can easily run into memory issues. Of course the 1GB example is arbitrary, if you remove the size limit in order to return at least one message, then the client has to accept arbitrarily large messages.

I'm not necessarily opposed to merging a change like this, but `max_bytes` definitely expresses an intent that no more than that amount of data (even if REST proxy can only enforce it roughly) should be returned. There's another issue filed that suggests providing a `min_bytes` parameter to allow for more push-like behavior. Would setting `max_bytes` to a large value and combining it with a `min_bytes` parameter to get more prompt responses to consume requests be a better solution? In that case you would explicitly be allowing for much larger messages (or still be able to cap their size!), but be explicitly requesting a faster response if only a smaller amount of data is available.

kjvalencik:
The issue I am having is that I have a couple of topics with wildly
different message sizes that each take roughly the same amount of time.

With the normal Java consumer, this is okay because health checks are not
tightly coupled to pulling data. It is possible to set a value that is
larger than your largest message. Then if this results in a very large
number of messages, you can continue to heartbeat while you drain the
queue.

But, this isn't possible in kafka rest because pulling messages is the only
way to heartbeat. If you are unable to process your local queue in time,
the consumer will time out, a rebalance will occur and you won't be able to
commit offsets. This starts the cycle over.

There are a few possible solutions:

* Allow returning a single message that exceeds the REST max bytes, but not
exceeding the internal consumer max bytes (what this PR does)
* Add an endpoint that allows heartbeats without pulling messages
* Allow committing arbitrary offsets. This helps the least because you
still have unnecessary rebalances.

Thanks!

gawth:
Having looked into this area as part of my pull request I agree with @ewencp that it would be a bit of a significant change.

With regards heartbeat, we're using the max bytes to our advantage.  We have a heartbeat thread that makes a request to kafka-rest for messages with max bytes set to 1byte.  This will never return a message (non of our messages are that small!) but keeps the consumer active.  Would that work for you?

Alan

kjvalencik:
I had considered that work around. The issue with using that approach is it is a very expensive way to do a heartbeat (possibly due to one of these: https://github.com/confluentinc/kafka-rest/issues/154, https://github.com/confluentinc/kafka-rest/issues/162). Although, I haven't delved in too deeply.

We are already having significant cpu usage with the current load (consuming two CPU cores all of the time, with 15 seconds of lagtime between batches) and I am concerned that the additional load would be unreasonable.

In the meantime, I'm content continuing to use a forked branch with this change. In regards to @gawth, I think my preferred change would be a dedicated health check endpoint. This could also potentially decrease load because we would be able to have fewer calls to fetch.

gawth:
We've got a low number of topics at the moment so not been an issue for us for the number of health checks we need to do.

How many topics/partitions are you dealing with?
How frequently would you be carrying out the health check?
What's your current request timeout on kafka-rest?
What's your consumer timeout configured to?

Ta,
Alan

kjvalencik:
* **How many topics/partitions are you dealing with?** ~50 topics, 30 partitions each
* **How frequently would you be carrying out the health check?** We are using the default consumer timeout (5 minutes), so I don't see a reason to do it more than once a minute.
* **What's your current request timeout on kafka-rest?** Default 1s.
* **What's your consumer timeout configured to?** Default 5 minutes.

Also, running 9 instances of `kafka-rest`.

kjvalencik:
@gawth How would you feel about this change if it was updated to make `maxBytes=0` a special case?

Logic:
* If `msg bytes + read bytes > max bytes`
    * if `max bytes = 0` then `return`
    * if `read bytes > 0` then `return`
    * else `continue reading`
* else `continue reading`

alecswan:
We just spent two days troubleshooting a slightly different problem caused by the same code. Yesterday we observed that the consumer lag on some partitions kept constantly growing. The REST client was requesting 32MB max_bytes, the REST proxy consumer was configured to read up to 64MB and the message that caused the problem was 40MB in size.

The code change in this PR request would have fixed our issue. I also understand the concern that this code change may cause some existing clients to experience higher memory usage. However, as it stands right now the client has no way to determine if the topic has no more messages or it's just stuck on a larger message than max_bytes. At the very minimum an exception should be thrown and converted to 429 or some 5xx response code if (bytesConsumed == 0 && roughMsgSize > maxResponseBytes).

PR #163, opened 2016-02-17
Doc change to remove the expectation that a broker is running on localhost:9092.

alexlod:
The docs currently say that a broker is expected to be running on localhost:9092. This isn't true. A broker doesn't need to be running on localhost:9092, and a bootstrap list of brokers doesn't need to be configured, either.


ewencp:
LGTM. When you merge, and you also merge into master?

alexlod:
See https://github.com/confluentinc/kafka-rest/pull/164 for the merge into master.

PR #165, opened 2016-03-02
Fix for issue confluentinc/kafka-rest/issues/118

luinix:
This is a quick fix, not intended to be the best solution, for solving the issue with string partition keys in avro messages.

When we try to convert the [key to avro](https://github.com/confluentinc/kafka-rest/blob/v2.0.1/src/main/java/io/confluent/kafkarest/AvroRestProducer.java#L96), we end up using the `JsonDecoder` class in Apache Avro. From that class we return an `org.apache.avro.util.Utf8` that later on [we send to the KafkaProducer](https://github.com/confluentinc/kafka-rest/blob/v2.0.1/src/main/java/io/confluent/kafkarest/AvroRestProducer.java#L109).

The problem is, Kafka-rest is [injecting a SchemaRegistry's `KafkaAvroSerializer` class instance](https://github.com/confluentinc/kafka-rest/blob/v2.0.1/src/main/java/io/confluent/kafkarest/ProducerPool.java#L109-L114) for decoding, which ends up using [this code](https://github.com/confluentinc/schema-registry/blob/v2.0.1/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroSerDe.java#L92):

```java
protected Schema getSchema(Object object) {
    if (object == null) {
      return primitiveSchemas.get("Null");
    } else if (object instanceof Boolean) {
      return primitiveSchemas.get("Boolean");
    } else if (object instanceof Integer) {
      return primitiveSchemas.get("Integer");
    } else if (object instanceof Long) {
      return primitiveSchemas.get("Long");
    } else if (object instanceof Float) {
      return primitiveSchemas.get("Float");
    } else if (object instanceof Double) {
      return primitiveSchemas.get("Double");
    } else if (object instanceof String) {
      return primitiveSchemas.get("String");
    } else if (object instanceof byte[]) {
      return primitiveSchemas.get("Bytes");
    } else if (object instanceof GenericContainer) {
      return ((GenericContainer) object).getSchema();
    } else {
      throw new IllegalArgumentException(
          "Unsupported Avro type. Supported types are null, Boolean, Integer, Long, " +
          "Float, Double, String, byte[] and IndexedRecord");
    }
  }
```
The class `org.apache.avro.util.Utf8` is not inheriting from `java.lang.String`, so we end up with an `IllegalArgumentException` error.

This PR just casts the `org.apache.avro.util.Utf8` to a `java.lang.String` as a fast fix. Probably the proper fix should be done in other place.

### DIFF
--- a/src/main/java/io/confluent/kafkarest/AvroRestProducer.java
+++ b/src/main/java/io/confluent/kafkarest/AvroRestProducer.java
@@ -94,6 +94,9 @@ public class AvroRestProducer implements RestProducer<JsonNode, JsonNode> {
         // if there isn't a schema. Validation will have already checked that all the keys/values
         // were NullNodes.
         Object key = (keySchema != null ? AvroConverter.toAvro(record.getKey(), keySchema) : null);
+        if (key instanceof org.apache.avro.util.Utf8) {
+          key = key.toString();
+        }
         Object value = (valueSchema != null
                         ? AvroConverter.toAvro(record.getValue(), valueSchema) : null);
         Integer recordPartition = partition;
--- a/src/test/java/io/confluent/kafkarest/integration/AvroProducerTest.java
+++ b/src/test/java/io/confluent/kafkarest/integration/AvroProducerTest.java
@@ -64,7 +64,8 @@ public class AvroProducerTest extends ClusterTestHarness {
 
   // This test assumes that AvroConverterTest is good enough and testing one primitive type for
   // keys and one complex type for records is sufficient.
-  private static final String keySchemaStr = "{\"name\":\"int\",\"type\": \"int\"}";
+  private static final String intKeySchemaStr = "{\"name\":\"int\",\"type\": \"int\"}";
+  private static final String stringKeySchemaStr = "{\"name\":\"string\",\"type\": \"string\"}";
   private static final String valueSchemaStr = "{\"type\": \"record\", "
                                                + "\"name\":\"test\","
                                                + "\"fields\":[{"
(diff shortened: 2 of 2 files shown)


ewencp:
@luinix If the issue is the `Utf8` class (an irritating quirk of Avro unfortunately...), then perhaps the more natural fix is to support `Utf8` in the serializer classes in confluentinc/schema-registry? The current patch is, as you say, pretty hacky since it fixes a very specific case of a more general problem.

luinix:
@ewencp I agree, but I'm not familiar enough with Avro and these projects' code to know the implications of that change. That's why I just coded a hotfix and waited for someone with more knowledge to propose the right place for the fix :smile:

luinix:
Also, should we hardcode the `org.apache.avro.util.Utf8` class in that serializer, or maybe make that class inherit from `java.langString`? Or make the serializer check for `CharSequence` implementations?

ewencp:
@luinix Great point, `CharSequence` is actually the right choice since it covers all the string types that Avro knows how to serialize. I think Avro made `Utf8` implement `CharSequence` hoping that people would start to prefer `CharSequence` interface, but most people continue to expect `String` instead. `CharSequence` is more general and often a the sufficient interface, but unfortunately not what people default to in general.

If we adjusted the serializers to use `CharSequence` instead, I think this would probably resolve the issue you're seeing. It'd be great to get some confirmation if it really does address the problem or not.

luinix:
Closing this PR, the fix for this issue is in confluentinc/schema-registry/pull/297

PR #166, opened 2016-03-04
Bump version to 3.0.0-SNAPSHOT and Kafka dependency to 0.10.0.0-SNAPSHOT

ewencp:


ewencp:
@miguno @ijuma @gwenshap @norwood Updating versions to clean up issues with system tests to address some issues in confluentinc/muckrake#90. Personally verified on that PR that it fixes issues w/ incompatible changes. These changes will be required for release anyway, modulo the SNAPSHOT suffix. Note that this upgrades all deps to 0.10.0.0-SNAPSHOT. Previous discussion left it unclear whether we might want to depend on older versions in some cases, but it simplifies things for now if we maintain one version across all packages.

Depends on confluentinc/schema-registry#298

src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
```diff
@@ -172,7 +172,7 @@ public void testConsumerNormalOps() throws InterruptedException, ExecutionExcept
     String cid = consumerManager.createConsumer(
         groupName, new ConsumerInstanceConfig(EmbeddedFormat.BINARY));
     sawCallback = false;
-    consumerManager.readTopic(
+    Object result = consumerManager.readTopic(
```
ijuma: Was this an existing but that you fixed?

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -210,9 +213,10 @@ private JsonConsumerRecord createJsonConsumerRecord(final MessageAndOffset messa
                                                       final String topicName,
                                                       final int partitionId) {
     final MessageAndMetadata<Object, Object> messageAndMetadata =
-        new MessageAndMetadata<Object, Object>(topicName, partitionId,
-            messageAndOffset.message(), messageAndOffset.offset(),
-            jsonDecoder, jsonDecoder);
+        new MessageAndMetadata<>(topicName, partitionId,
+                                 messageAndOffset.message(), messageAndOffset.offset(),
+                                 0, TimestampType.CREATE_TIME,
```
ijuma: What is the implication of always passing `0` instead of current time here?

ijuma:
A couple of questions, LGTM otherwise.

src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
ewencp: Nope, oversight from some debugging which I've removed in an updated patch. Thanks for catching that!

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
ewencp: Tbh, these extra timestamp parameters have no impact on the existing tests since they are completely new. The exact values didn't matter.

ijuma: Fair enough (I don't know how these classes are actually used :)).

src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
ijuma: Thanks, looks good.

PR #167, opened 2016-03-04
increase producer instances to improve produce throughput

xingfeng2510:


xingfeng2510:
@ConfluentJenkins @hachikuji Fristly, I am very glad to receive yours feedbacks，But....So sorry to overwrite your comments by mistake, could you review it again?

xingfeng2510:
@ConfluentCLABot I have posted my Confluent CLA submission via http://clabot.confluent.io/cla, wish to pass yours license. Thanks a lot!

src/main/java/io/confluent/kafkarest/ProducerPool.java
```diff
@@ -37,17 +37,21 @@
...
 /**
  * Shared pool of Kafka producers used to send messages. The pool manages batched sends, tracking
- * all required acks for a batch and managing timeouts. Currently this pool only contains one
- * producer per serialization format (e.g. byte[], Avro).
+ * all required acks for a batch and managing timeouts. Currently this pool contains three
+ * producers per serialization format (e.g. byte[], Avro).
  */
 public class ProducerPool {
 
+  private static int PRODUCER_COUNT_PER_FORMAT = 3;
```
cpuchakayaka: This should be made configurable

ewencp:
@xingfeng2510 Do you have benchmarks or something to show improvement here? The new producer usually does quite well in terms of speed, and using more producers can actually *hurt* performance if you don't have the right workload.

Agreed with @cpuchakayaka that the # of producers would need to be configurable. The current approach to selecting a producer also doesn't preserve ordering that clients might expect since it is spreading the load randomly across producers. We'd need a solution that can consistently select the same producer for a given client, which probably means some identifier for clients that can persist across HTTP requests and TCP connections (e.g., an HTTP header, maybe a cookie).

kpocius:
We've managed to go from ~10k req/s to ~25k req/s on a single instance of rest proxy by playing around with the producer count. We can provide testing scripts and infrastructure description if that's relevant.

If this is not getting implemented, at least the documentation of producer.threads should be updated because it's very misleading.

Regarding ordering, this information could easily be added to the documentation of the parameter (once it's actually implemented). Currently, the documentation states that it's easy to scale rest proxy by adding more instances behind some sort of DNS load balancer - wouldn't that mess with the ordering as well? So if we can increase the number of producers rather than spinning up more and more instances, that would be much better.

PR #169, opened 2016-03-07
.NET Sample consuming the Kafka Rest Proxy.

selvendranayyaswamy:
this is a sample .NET applicatiomn that consumes the Kafka to consume and produce the messages to the existing Topic.

### DIFF
--- a/Kafka-sample-dotnet
+++ b/Kafka-sample-dotnet
@@ -1,0 +1,317 @@
+namespace RestClient 
+{
+    public enum HttpVerb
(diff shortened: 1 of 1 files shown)


cricket007:
I feel like this code would be more useful as a blog or just [in your repo](https://github.com/selvendranayyaswamy/Kafka_Rest_Proxy_Example) rather than as an addition to this project. 

In other words, it should be enough to document the API in this project, not include all possible external HTTP client examples.

PR #205, opened 2016-05-23
CP-303: Rearrange quickstart and use topics from earlier steps in requests for metadata so the example output will exactly match real output when starting from an empty cluster.

ewencp:


blueck:
LGTM

PR #215, opened 2016-06-23
CP-140: add version.txt to share/doc

cotedm:


ewencp:
LGTM. @cotedm I don't know that it makes sense to review all of these directly as they are all basically the same, so consider this a LGTM on all of them :)

PR #216, opened 2016-06-27
REST-204 Add support for new Kafka Consumer

GRpro:
- New consumer for topics resource.
- New consumer for partitions resource
- Offset caching support for simple-consumer
  (partition resource)
- Tests


config/kafka-rest.properties
```diff
@@ -17,3 +17,4 @@
 #id=kafka-rest-test-server
 #schema.registry.url=http://localhost:8081
 #zookeeper.connect=localhost:2181
+#bootstrap.servers=localhost:9092
```
alexlod: Please also add this configuration parameter to `docs/config.rst`.

src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -55,9 +56,9 @@
 
   private final KafkaRestConfig config;
   private final Time time;
-  private final String zookeeperConnect;
+  private final String bootstrapServers;
   private final MetadataObserver mdObserver;
-  private final int iteratorTimeoutMs;
+//  private final int iteratorTimeoutMs;
```
alexlod: Please delete any commented code. Better to not check this in and instead let version control track files that have been deleted/added.

src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -145,13 +146,15 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
 
+      Properties props;
       // Note the ordering here. We want to allow overrides, but almost all the
       // consumer-specific settings don't make sense to override globally (e.g. group ID, consumer
       // ID), and others we want to ensure get overridden (e.g. consumer.timeout.ms, which we
       // intentionally name differently in our own configs).
-      Properties props = (Properties) config.getOriginalProperties().clone();
-      props.setProperty("zookeeper.connect", zookeeperConnect);
+      props = (Properties) config.getOriginalProperties().clone();
+      props.setProperty("bootstrap.servers", bootstrapServers);
```
alexlod: You can use `KafkaRestConfig.BOOTSTRAP_SERVERS_CONFIG` here instead of `bootstrap.servers`.

src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
```diff
@@ -108,37 +110,50 @@ public boolean doPartialRead() {
...
           }
 
-          iter.next();
-          messages.add(recordAndSize.getRecord());
-          bytesConsumed += roughMsgSize;
-          // Updating the consumed offsets isn't done until we're actually going to return the
-          // data since we may encounter an error during a subsequent read, in which case we'll
-          // have to defer returning the data so we can return an HTTP error instead
+          while (iter.hasNext()) {
+            org.apache.kafka.clients.consumer.ConsumerRecord<KafkaK, KafkaV> record = iter.next();
```
alexlod: May be good to rename the `ConsumerRecord` class in the REST proxy now that you're using the `ConsumerRecord` in the "new" consumer. That feels cleaner to me than dealing with a name collision. Let me know if you disagree.

src/main/java/io/confluent/kafkarest/ConsumerSubscriptionState.java
```diff
@@ -0,0 +1,165 @@
+/**
+ * Copyright 2015 Confluent Inc.
```
alexlod: Please use `2016` instead of `2015`. The same is true for the other licenses agreements below. Thanks!

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -57,6 +57,18 @@
       + "the connection string as hostname1:port1,hostname2:port2,hostname3:port3/chroot/path.";
   public static final String ZOOKEEPER_CONNECT_DEFAULT = "localhost:2181";
 
+  public static final String BOOTSTRAP_SERVERS_CONFIG = "bootstrap.servers";
+  private static final String
+      BOOTSTRAP_SERVERS_DOC =
```
alexlod: Might be worth calling out that the only protocol currently supported is PLAINTEXT, hence why `bootstrap.servers` doesn't take a protocol (eg `PLAINTEXT://localhost:9092`).

src/main/java/io/confluent/kafkarest/SimpleConsumerRecordsCache.java
```diff
@@ -0,0 +1,306 @@
...
+        return res;
+      } else {
+        return null;
+      }
+    }
+
+
+    /**
+     * Adds new record to the cache if it has grater priority
+     * (grater offset) than all existing records or if the cache
```
alexlod: Nitpick: type. `grater` -> `greater`.

alexlod:
@GRpro, thanks for contributing this! I took a first, quick pass and made a few comments. I'm not familiar with the REST proxy codebase so I'll need more time with your code to give a full review. I'll continue to make comments. In the meantime, feel free to address my existing comments by pushing new commits, but please don't rebase so I can see your changes. Let's see who else chimes in with a review, too. Thanks again!

config/kafka-rest.properties
ewencp: It's worth noting here that this can be autogenerated by running the `main()` in `KafkaRestConfig`

src/main/java/io/confluent/kafkarest/AvroConsumerState.java
```diff
@@ -35,35 +38,42 @@
...
-    props.setProperty("schema.registry.url",
-                      config.getString(KafkaRestConfig.SCHEMA_REGISTRY_URL_CONFIG));
-    decoder = new KafkaAvroDecoder(new VerifiableProperties(props));
+  public AvroConsumerState(KafkaRestConfig config, ConsumerInstanceId instanceId,
+                           Properties consumerProperties,
+                           ConsumerManager.ConsumerFactory consumerFactory) {
+    super(config, instanceId, consumerProperties, consumerFactory);
+  }
+
+  private Deserializer<Object> initDeserializer() {
```
hachikuji: Any particular reason this was moved out of the constructor?

src/main/java/io/confluent/kafkarest/BinaryConsumerState.java
```diff
@@ -17,42 +17,44 @@
...
 
   @Override
-  protected Decoder<byte[]> getValueDecoder() {
-    return decoder;
+  protected Deserializer<byte[]> getValueDeserializer() {
+    return deserializer;
   }
 
   @Override
   public ConsumerRecordAndSize<byte[], byte[]> createConsumerRecord(
```
hachikuji: This name makes a little less sense with the type change. Maybe we could change it to `convertRecord` or something like that?

src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -145,13 +146,15 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
     try {
       log.debug("Creating consumer " + name + " in group " + group);
 
+      Properties props;
```
hachikuji: nitpick: this change seems unnecessary.

src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -160,49 +163,57 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+
       try {
-        if (consumerFactory == null) {
-          consumer = Consumer.createJavaConsumerConnector(new ConsumerConfig(props));
+        ConsumerFactory factory;
+        if (consumerFactory != null) {
+          factory = consumerFactory;
         } else {
-          consumer = consumerFactory.createConsumer(new ConsumerConfig(props));
+          factory = new ConsumerFactory() {
```
hachikuji: Any reason we don't assign this to `consumerFactory` in the constructor instead of setting it to null?

src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -160,49 +163,57 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+            @Override
+            public <K, V> KafkaConsumer<K, V> createConsumer(Properties props,
+                                                             Deserializer<K> keyDeserializer,
+                                                             Deserializer<V> valueDeserializer) {
+              return new KafkaConsumer<K, V>(props, keyDeserializer, valueDeserializer);
+            }
+          };
+        }
+        ConsumerState state;
+        switch (instanceConfig.getFormat()) {
```
hachikuji: minor: a slightly more elegant option would be to add a factory method to the `EmbeddedFormat` enum to create the right `ConsumerState` depending on the instance.

src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
```diff
@@ -108,37 +110,50 @@ public boolean doPartialRead() {
...
-          long roughMsgSize = recordAndSize.getSize();
-          if (bytesConsumed + roughMsgSize > maxResponseBytes) {
-            break;
+        while (parent.getConfig().getTime().milliseconds() < endTime) {
+
+          if (iter == null || !iter.hasNext()) {
+            // The consumer timeout should be set very small, so the expectation is that even in the
+            // worst case, num_messages * consumer_timeout << request_timeout, so it's safe to only
+            // check the elapsed time once this loop finishes.
+            ConsumerRecords<KafkaK, KafkaV> records = consumer.poll(itBackoff);
```
hachikuji: I think we might need to think through this a bit. The old consumer did most of its work (including rebalancing and fetching) in the background, but the new consumer is designed to work as an event loop with the user calling poll() repeatedly and all of the work done in the foreground inside poll(). When you stop calling poll(), the consumer stops sending heartbeats and the broker reassigns its partitions to another member. So I think we may need to do a little more work to ensure that consumers stay in the group as long as users are fetching data. 

Thinking outloud... One option I can think of would be to change `ConsumerWorker` to work only with a single instance of `KafkaConsumer`. Since each worker gets its own thread, it can execute the poll loop for that consumer instance and feed records into a queue, which can then be accessed asynchronously. Then we could have `ConsumerManager` maintain a map from the `ConsumerInstanceId` to the `ConsumerWorker`. This might even simplify expiration since we can have the worker itself shutdown when we see no reads from the consumer after some time.

src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
hachikuji: I'm actually sort of hoping it might be possible to delete the `ConsumerRecord` we have here and use the one in Kafka only. I think we would just need to add a custom Jackson module to ensure that it serializes correctly. I don't think we necessarily need to do it here though.

src/main/java/io/confluent/kafkarest/AvroConsumerState.java
GRpro: In case of constructor initialization the superclass constructor is invoked before subclass and we get an NPE when getValue/KeyDeserializer() are called in superclass.

config/kafka-rest.properties
GRpro: Done. https://github.com/confluentinc/kafka-rest/pull/216/commits/b8a2cf58273d8038ffe30a630f52d1490b163002
Thanks.

src/main/java/io/confluent/kafkarest/ConsumerManager.java
GRpro: Done. https://github.com/confluentinc/kafka-rest/pull/216/commits/b8a2cf58273d8038ffe30a630f52d1490b163002
Thanks.

src/main/java/io/confluent/kafkarest/ConsumerManager.java
GRpro: Done. https://github.com/confluentinc/kafka-rest/pull/216/commits/b8a2cf58273d8038ffe30a630f52d1490b163002
Thanks.

src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
GRpro: @alexlod I've enamed it to AbstractConsumerRecord here https://github.com/confluentinc/kafka-rest/pull/216/commits/b8a2cf58273d8038ffe30a630f52d1490b163002
Thanks.

src/main/java/io/confluent/kafkarest/ConsumerSubscriptionState.java
GRpro: Done. https://github.com/confluentinc/kafka-rest/pull/216/commits/b8a2cf58273d8038ffe30a630f52d1490b163002
Thanks.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
GRpro: Added PLAINTEXT://localhost:9092 as a default value. 
https://github.com/confluentinc/kafka-rest/pull/216/commits/b8a2cf58273d8038ffe30a630f52d1490b163002
Thank you.

src/main/java/io/confluent/kafkarest/SimpleConsumerRecordsCache.java
GRpro: Done https://github.com/confluentinc/kafka-rest/pull/216/commits/b8a2cf58273d8038ffe30a630f52d1490b163002

src/main/java/io/confluent/kafkarest/BinaryConsumerState.java
GRpro: Done

src/main/java/io/confluent/kafkarest/ConsumerManager.java
GRpro: Done

src/main/java/io/confluent/kafkarest/ConsumerManager.java
GRpro: Moved initialization to constructor. https://github.com/confluentinc/kafka-rest/pull/216/commits/b8a2cf58273d8038ffe30a630f52d1490b163002
Thanks.

src/main/java/io/confluent/kafkarest/ConsumerManager.java
GRpro: Done. https://github.com/confluentinc/kafka-rest/pull/216/commits/b8a2cf58273d8038ffe30a630f52d1490b163002
Thanks.

src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
GRpro: @hachikuji my understanding is when records systematically appear in a topic and nobody retrieves them from REST proxy there is a risk of OutOfMemoryError with such approach. We can only bound records by the expiration timeout which gives us not much flexibility in configuration.
Could you please write a confirmation if I need to implement this approach.
Thank you!

cpuchakayaka: The 0.8 Kafka Consumer has two settings,  one is the num.consumer.fetchers and the other is queued.max.message.chunks, configuring these settings approach greatly improves the performance in the REST approach as once a request has been served for a consumer, it starts fetching messages in the background, it already has some messages fetched by the time another request comes in for the consumer instance. With 0.9 Kafka Consumer, we are kinda of losing this which will definitely reduce the performance/throughput. I would suggest to create a wrapper around the KafkaConsumer, that maintains a blocking queue (configurable) and the keeps polling for new messages in the background rather than waiting for a request to arrive, which also solves the problem with heartbeats/rebalancing that @hachikuji was talking about.  This wrapper maintains internal state of offsets to be committed, similar to the 0.8 Kafka Consumer

cpuchakayaka: The 0.8 Kafka Consumer has two settings, one is the num.consumer.fetchers and the other is queued.max.message.chunks, configuring these settings approach greatly improves the performance in the REST approach as once a request has been served for a consumer, it starts fetching messages in the background, it already has some messages fetched by the time another request comes in for the consumer instance. With 0.9 Kafka Consumer, we are kinda of losing this which will definitely reduce the performance/throughput. I would suggest to create a wrapper around the KafkaConsumer, that maintains a blocking queue (configurable) and the keeps polling for new messages in the background rather than waiting for a request to arrive, which also solves the problem with heartbeats/rebalancing that @hachikuji was talking about.

GRpro: Wouldn't it better to give each `ConsumerState` a blocking queue and to have a separate thread that periodically performs poll(0) on each `KafkaConsumer` object? Retrieved records are put into that queue. When a user asks consumer for records:
1) Records from queue are retrieved first and it becomes empty. If the queue is not empty go to #3
2) poll with some timeout is invoked (We can still use Round-Robin strategy to poll for records by invoking `doPartitialRead()` many times). Then these records are appended to those retrieved in #1.

3) return records to the user.

With such a scenario we do not need to have as many threads as consumers.

cpuchakayaka: I think having separate thread for each consumer that does the poll in the background is better having a single thread that does the iterate over consumers, which would create contention or performance concerns. It would be good to have a configurable BlockingQueue in _ConsumerState_ and a background thread that polls and puts messages into the BlockingQueue , this BlockingQueue will contain chunks of messages , each poll will give a chunk of messages.

src/main/java/io/confluent/kafkarest/ConsumerState.java
```diff
@@ -15,114 +15,240 @@
...
+      lock.unlock();
+    }
+  }
+
+  /**
+   * To keep a Consumer subscribed the poll must be invoked periodically
+   * to send heartbeats to the kafka back-end
+   * This method should not block for a long period.
+   */
+  public void sendHeartbeat() {
```
cpuchakayaka: I think it would good to put this code and the code in **ConsumerReadTask.doPartialRead()** in one class, so we have one place where the _consumer.poll_ is done and records are put in the queue

cpuchakayaka: It might be worth the wait for https://issues.apache.org/jira/browse/KAFKA-3888 to be done, as the heartbeat is being separated from the _poll_ method and will done in a separate background thread,  it would make this code much cleaner.

cpuchakayaka: Any ETA on when this pull request will be merged?

hachikuji:
@GRpro Thanks a lot for your effort (and patience) on this. I think there's really two parts to this patch. The first is the conversion of the "high-level consumer" usage. This probably requires some more design work since the new consumer exposes a very different API and threading model. For example, the current APIs reflect the topic-centric nature of the old consumer, but the new consumer doesn't expose a way to fetch topics independently unless you change the subscription to just that topic, which sort of defeats any kind of prefetching that it can do. 

The second part of this patch is conversion of the "simple consumer" usage. Since this is much more straightforward, I'm wondering if it would make sense to try and split it into a separate PR? We should be able to get it merged fairly quickly if it's not coupled with the rest. What do you think?

cpuchakayaka:
@hachikuji not sure I fully understand the topic centric nature of the old consumer you were talking about, the new consumer is still topic(s) centric, but it allows you to change subscription after initial subscription of topic(s). With the new consumer prefetching can be done until the subscription changes. The prefetching config could something similar to the old consumer where we put a limit on how many messages are prefetched.Also need to add the ability to change the subscription of topic(s) thru the rest API (which is not a necessary requirement for the initial porting to new consumer). The implementation of heartbeats for new consumer is changing ( https://issues.apache.org/jira/browse/KAFKA-3888), anything we do now with heartbeats is going to change, maybe @ewencp can tell us when Kafka 0.10.1.0 will be released to get a better understanding of when the port to new consumer will be done.

hachikuji:
@cpuchakayaka What I mean is that it's more natural to have topic-centric APIs with the old consumer since you can get a separate message stream for each topic and the consumer hides the fetching. With the new consumer, you subscribe to a set of topics and there is no (easy) way to fetch from only one of those topics. Changing the subscription of a consumer to answer a single API request is a pretty tough sell since the consumer rebalances on subscription changes. A better option would be to use pause/resume, but my guess is that this is going to be pretty complex.

What I'm imagining is an API which more closely mirrors the new consumer. For example, we might have something like this:

1. Subscribe to topics: POST /consumers/{id}/subscription
2. Fetch from subscribed topics: GET /consumers/{id}/records

The addition of KAFKA-3888 (which I am implementing) may give us some more flexibility in the implementation (I'm not too sure), but I think it's largely orthogonal to this issue. My hunch is that we will still need to have a separate thread for each consumer to maintain the poll loop since otherwise we won't be able to respond quickly to rebalances.

cpuchakayaka:
@hachikuji  makes sense. I think the rest API need to support the basic use case(s) initially (endpoints you were proposing) and later on pausing and resuming functionality. It would also be good to expose the seek functionality in the rest API. I totally agree with the separate thread for each consumer, which is what I was proposing in one of my previous comments. But doing the heartbeats with current consumer api (with out KAFKA-3888) could be tricky if the consumer poll is done in a separate thread, so was wondering what the timeline for releasing KAFKA-3888 would be.

alexlod:
@GRpro: @ewencp, @hachikuji, and I have been working on a new consumer REST API. We have a draft document here:

https://github.com/confluentinc/kafka-rest/wiki/New-Consumer-API-Design

The draft document should be enough to get you unblocked, and should also be a good starting point to start thinking about other API changes, including but not limited to the APIs that the document has not addressed yet (eg `assign`).

Take a look at the doc, feel free to discuss it with us and propose changes, and hopefully you can also continue working on this PR.

Thanks everyone!

hachikuji:
@cpuchakayaka The plan is for KAFKA-3888 to go in 0.10.1 (or whatever the next kafka release happens to be). I think the review is nearing completion, just a few more wrinkles to iron out. I agree that we'll have to think through heartbeating and the consumer's liveness a bit. One of the difficulties of adding APIs like seek() is that they're tied to the consumer's current assignment. That suggests we might need some hook into the rebalance logic, and I'm not sure how we'll do that over REST. Overall, since this is clearly a large effort, it seems best to split it into separate pieces where it makes sense. The first is getting the simple consumer APIs converted. Maybe the next is getting the basic subscribe/poll functionality worked out. Finally, we can think about more advanced APIs like pause/resume.

cpuchakayaka:
@hachikuji I agree that its a large effort. I am willing pitch in to make the basic subscribe/poll functionality done quickly.

groghkov:
@hachikuji I will separate changes for simple consumer into a separate PR. I agree that splitting the task into subtasks gives a better understanding of each problem.

GRpro:
@hachikuji I've created a PR: https://github.com/confluentinc/kafka-rest/pull/234

hachikuji:
@GRpro Awesome. I'll help review the patch this week. Have you had a chance to take a look at the  document posted by @alexlod? We may need to iterate on the design depending what people think. I think @cpuchakayaka was also willing to help out with this effort, so you might consider whether there are reasonable ways to divide the work.

bharatviswa504:
@hachikuji Could you please let us know when the new consumer will be integrated?

ewencp:
@bharatviswa504 It's actively being worked on, should be in the next release.

jeffwidman:
Is this PR still relevant or was there a newer one that supercedes this? I'm thinking I saw the 3.2 release took care of this...

cpuchakayaka:
There is new consumer in 3.2, but I don't think it works :), there was an issue with reading messages with new v2 rest api.

jeffwidman:
Is it possible to get a status update here?

Trying to figure out if/when the migration to the new consumer is happening, and what the API might look like as we are writing a quick-n-dirty pseudo API endpoint internally at my day job, and we'd like to match the new API so whenever rest proxy ships with the new consumer/new APIs, that we only have to change the http endpoint. 

Bump @hachikuji / @ewencp

ewencp:
Work on the new consumer support took a few iterations, but it has been added as of 3.2.0 in the [v2 API](https://docs.confluent.io/current/kafka-rest/docs/api.html#consumers).

I'm going to close this out as it was a first iteration but ultimately ended up going a bit of a different direction.

Currently both v1 and v2 APIs are still supported, so you can use both still. However, there is not an online upgrade path between the two due to the nature of partition assignment, so upgrading an application would require bringing down the current instances and then bring up a new set of instances that have switched to the new API (assuming you are doing offset commits via Kafka so the new consumer instances pick up the same offsets the old consumer instances committed upon shutdown).

Issue #229, opened 2016-08-01
Add Documentation on Jetty Metrics


kevinconaway:
The rest-utils project now exposes the JMX metrics from the Jetty server (like thread pool management, session usage, servlet handler metrics etc) via https://github.com/confluentinc/rest-utils/issues/46

We should update the rest proxy documentation to reflect that these metrics are now exposed.

ewencp:
Fixed by #230

PR #230, opened 2016-08-01
Issue #229: Add blurb on Jetty jmx metrics

kevinconaway:
Commit for #229

I'm not of the correct way to document these metrics.  There isn't really any public documentation available on the specifics on what Jetty is actually exposing so I'm not sure if this is the right place to do it.  It exposes a lot of different beans and attributes so it would be cumbersome to list them out here.

I took the tack of providing a hint for users that the data is indeed there and can be discovered via jconsole/visualvm/whatever.


ewencp:
LGTM, thanks!

Btw, if you're just submitting a PR anyway, don't feel like you also need to file a separate issue.

PR #234, opened 2016-08-10
REST-233 Simple consumer with KafkaConsumer 0.9+

GRpro:


docs/config.rst
```diff
@@ -11,20 +11,69 @@ provided as configuration options, and finally falls back to the default values
...
+  * Importance: high
+
 ``id``
   Unique ID for this REST server instance. This is used in generating unique IDs for consumers that do not specify their ID. The ID is empty by default, which makes a single server setup easier to get up and running, but is not safe for multi-server deployments where automatic consumer IDs are used.
 
   * Type: string
   * Default: ""
   * Importance: high
 
+``listeners``
```
hachikuji: I'm not sure why this is included. It doesn't look like we use the new configuration anywhere.

docs/config.rst
```diff
@@ -11,20 +11,69 @@ provided as configuration options, and finally falls back to the default values
...
+  * Importance: high
+
 ``schema.registry.url``
   The base URL for the schema registry that should be used by the Avro serializer.
 
   * Type: string
   * Default: "http://localhost:8081"
   * Importance: high
 
+``ssl.key.password``
```
hachikuji: I guess you were intending to use these to support TLS for the proxy itself? If so, maybe we could move them into a separate patch?

docs/config.rst
```diff
@@ -62,13 +111,76 @@ Java Kafka clients.
...
   * Importance: medium
 
+``simpleconsumer.cache.max.records``
+  Maximum number of records that can be stored for a specific topic-partition combination. Records with higher offsets replace records with lower ones Must be greater that 0.
+
+  * Type: int
+  * Default: 1000
+  * Importance: medium
+
+``simpleconsumer.max.caches.num``
```
hachikuji: This configuration doesn't seem too useful. I guess the idea is to bound the total memory devoted to the cache, but that's difficult to do anyway knowing the size of the cached records. In practice, it seems simpler to use only per-partition cache settings and let users scale the number of proxy instances as needed. What do you think?

docs/config.rst
```diff
@@ -174,9 +286,51 @@ Java Kafka clients.
   * Default: 1000
   * Importance: low
 
+``simpleconsumer.max.poll.records``
```
hachikuji: Would there ever be any reason for this to be different from `simpleconsumer.cache.max.records`?

docs/config.rst
```diff
@@ -174,9 +286,51 @@ Java Kafka clients.
...
+  Maximum number of records that can be fetched by a single consumer poll request. Since desired number of fetched records can't be defined for every poll request the consumer can poll more records thannecessary (extra records are stored in cache.). The value of this property defines maximum number of excess records fetched by a single user request. The grater value means grater network overhead. It may be reasonable if user retrieves records from specified partition increasing offset sequentially. The smaller value suits better when the user consumes records in a random fashion.Use 0 for no max value allowed by kafka consumer 2147483647
+
+  * Type: string
+  * Default: "100"
+  * Importance: low
+
+``simpleconsumer.max.poll.time``
```
hachikuji: This seems like something users shouldn't have any reason to adjust.

GRpro: I agree that it is verbose to use `simpleconsumer.max.poll.records` as there is already `simpleconsumer.cache.max.records`.

docs/config.rst
GRpro: We can remove it and use `simpleconsumer.cache.max.records` instead

docs/config.rst
GRpro: I agree that it is hard to calculate maximum size of RAM needed for rest proxy in the worst case. Even with number of caches property as we don't know size of each record. I will remove number of caches bound.

docs/config.rst
GRpro: Typo

docs/config.rst
GRpro: That is my typo during cherry-pick from another branch.

GRpro:
@hachikuji I've rebased commit as I figured out that the previous version contained diff that I included by mistake when cherry-picking. This version has changes that you've mentioned. All following changes I will perform as new commits on top of existing. Please, take a look again and let me know if there is anything I can help with. Thanks!

src/main/java/io/confluent/kafkarest/SimpleConsumerFactory.java
```diff
@@ -59,16 +58,40 @@ private String nextClientId() {
...
 
     final String clientId = nextClientId();
 
-    log.debug("Creating SimpleConsumer with id " + clientId + " (host: " + host + ", port: " + port + ")");
-    return new SimpleConsumer(
-        host, port,
-        simpleConsumerConfig.socketTimeoutMs(),
-        simpleConsumerConfig.socketReceiveBufferBytes(),
-        clientId);
+    log.debug("Creating SimpleConsumer with id " + clientId);
```
hachikuji: nitpick: can we use interpolation? i.e.
```java
log.debug("Creating SimpleConsumer with client id {}", clientId);
```

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -124,157 +90,101 @@ public void consume(final String topicName,
...
+    final String topicName,
+    final int partitionId) {
+
+    return new JsonConsumerRecord(
+      deserializeJson(consumerRecord.key()),
+      deserializeJson(consumerRecord.value()),
+        topicName, partitionId, consumerRecord.offset());
+  }
+
+  private Object deserializeJson(byte[] data) {
```
hachikuji: nitpick: it's a trivial function, but perhaps we can move this to a `JsonConverter` similar to `AvroConverter`?

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -124,157 +90,101 @@ public void consume(final String topicName,
...
-            messageAndOffset.message(), messageAndOffset.offset(),
-            binaryDecoder, binaryDecoder);
-    return new BinaryConsumerRecord(messageAndMetadata.key(), messageAndMetadata.message(),
-        partitionId, messageAndOffset.offset());
+  private BinaryConsumerRecord createBinaryConsumerRecord(
+    final org.apache.kafka.clients.consumer.ConsumerRecord<byte[], byte[]> consumerRecord,
+    final String topicName,
+    final int partitionId) {
+
+    // KafkaConsumer instances are created with ByteArrayDeserializer so
```
hachikuji: nitpick: comment seems redundant given the type information on ConsumerRecord.

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -124,157 +90,101 @@ public void consume(final String topicName,
                       final EmbeddedFormat embeddedFormat,
                       final ConsumerManager.ReadCallback callback) {
 
-    List<ConsumerRecord> records = null;
+    List<ConsumerRecord> records = new ArrayList<>();
```
hachikuji: A minor optimization would be to move this into the `else` branch below so that we can allocate with the exact number of records. Otherwise we can use `Collections.emptyList()`.

src/main/java/io/confluent/kafkarest/SimpleConsumerFactory.java
```diff
@@ -59,16 +58,40 @@ private String nextClientId() {
...
+    Properties properties = (Properties) simpleConsumerProperties.clone();
+
+    properties.setProperty("client.id", clientId);
+    properties.setProperty("max.poll.records", String.valueOf(
```
hachikuji: Sorry for the long comment, but this is probably the main work to sort out in this patch.

There is some subtlety with the interplay of max.poll.records and the cache that you've included below. Although you request only `max.poll.records` from the consumer, it will fetch an amount which is bound by `max.partition.fetch.bytes`, which may contain considerably more records than what was asked for. It will then feed out this data in batches of size max.poll.records when poll() is called.

This makes the configuration `cache.max.records` a little misleading since it doesn't account for the data cached in the consumer. In fact, there is an advantage to leaving the data in the consumer as long as possible. If the fetched data is compressed, then we will not decompress it until it is ready to be returned to the user, which helps to keep memory usage lower.

The other interesting thing is the impact of prefetching. Once the consumer's cached records are depleted, it will send a new round of fetches prior to returning from a poll() call. If we end up seeking to another offset before polling again, those fetches will have to be discarded. In the worst case, with multiple users fetching the same topic partition with the same consumer instance, we'll have to discard fetch requests from every call to poll(). So it might be nice to have more stickiness between the user and the KafkaConsumer instance. We could potentially do this by remembering the position of the consumer in the cache. Then instead of mapping to a consumer instance only by the topic partition, we can take into account the requested offset. For each topic partition, then we would have multiple consumers and we'd only need to do a seek when we've reached some limit.

Thoughts?

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -64,13 +49,13 @@
   private final MetadataObserver mdObserver;
   private final SimpleConsumerFactory simpleConsumerFactory;
```
hachikuji: Seems we don't need this anymore since the factory was moved into SimpleConsumerPool.

hachikuji: By the way, it seems rather excessive to have `SimpleConsumerManager`, `SimpleConsumerPool`,  and a `SimpleConsumerFactory`. Do you think it would make sense to at least consolidate `SimpleConsumerPoll` and `SimpleConsumerManager` into a single class or do you think it's better to keep the separation?

src/main/java/io/confluent/kafkarest/SimpleConsumerFactory.java
GRpro: My understanding is that Consumer object that is assigned for a single partition cannot store more records than `max.partition.fetch.bytes` (Please correct me it that is not right) so we can try to bound cache by size of records, say `max.partition.fetch.bytes` (not by number of records). In the worst case there would be number of records equal to `2 * max.partition.fetch.bytes` stored in memory for single TopicPartition (In case when the cache is full but no required offset is found in cache, consumer calls seek and poll and the second part of records is fetched.)
I will add stickiness between Cache and offset in the next commit.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -57,6 +57,18 @@
       + "the connection string as hostname1:port1,hostname2:port2,hostname3:port3/chroot/path.";
   public static final String ZOOKEEPER_CONNECT_DEFAULT = "localhost:2181";
 
+  public static final String BOOTSTRAP_SERVERS_CONFIG = "bootstrap.servers";
+  private static final String
+      BOOTSTRAP_SERVERS_DOC =
+      "A list of host/port pairs to use for establishing the initial connection to the Kafka cluster. "
+      + "The client will make use of all servers irrespective of which servers are specified here for bootstrapping—this "
+      + "list only impacts the initial hosts used to discover the full set of servers. This list "
+      + "should be in the form host1:port1,host2:port2,.... Since these servers are just used for "
```
ewencp: Was going to say we should clean this up since the `host1:port1` form doesn't cover the full format, though now I realize that this looks verbatim from the same Kafka config. We should probably update both since one of the major benefits of switching to the new consumer will be support for secure transports.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -129,6 +141,20 @@
...
+    SIMPLE_CONSUMER_MAX_POLL_TIME_DOC =
+    "Maximum amount of time to poll for records by a consumer.";
+  public static final int SIMPLE_CONSUMER_MAX_POLL_TIME_DEFAULT = 500;
+
+  public static final String SIMPLE_CONSUMER_CACHE_MAX_RECORDS_CONFIG = "simpleconsumer.cache.max.records";
+  private static final String
+    SIMPLE_CONSUMER_CACHE_MAX_RECORDS_DOC =
+    "Maximum number of records that can be stored for a specific topic-partition combination."
+     + " Records with higher offsets replace records with lower ones"
+     + " Must be greater that 0.";
```
ewencp: typo: that -> than

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -25,33 +25,18 @@
...
-import kafka.javaapi.FetchResponse;
-import kafka.javaapi.message.ByteBufferMessageSet;
-import kafka.message.MessageAndMetadata;
-import kafka.message.MessageAndOffset;
-import kafka.serializer.Decoder;
-import kafka.serializer.DefaultDecoder;
-import kafka.utils.VerifiableProperties;
-
-import org.apache.kafka.common.protocol.SecurityProtocol;
+import jersey.repackaged.com.google.common.collect.Maps;
```
ewencp: Is this considered a public and stable copy of Guava?

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
ewencp: Related to that might be the ability to override deserializer settings. We're not doing this at the moment, but maybe we should be?

src/main/java/io/confluent/kafkarest/SimpleConsumerPool.java
```diff
@@ -35,7 +36,7 @@
   private final Time time;
 
   private final SimpleConsumerFactory simpleConsumerFactory;
-  private final Map<String, SimpleConsumer> simpleConsumers;
+  private final Map<String, Consumer<byte[], byte[]>> simpleConsumers;
```
ewencp: Hmm, I'm honestly not sure why this was previously implemented this way with a separate map and list of IDs. Think we could just make simpleConsumers just a Queue and get rid of available consumers? Although with the suggestion about caching consumers by topic partition for more effective reuse, we might need to further adjust this.

src/main/java/io/confluent/kafkarest/SimpleConsumerPool.java
```diff
@@ -45,26 +46,40 @@ public SimpleConsumerPool(int maxPoolSize, int poolInstanceAvailabilityTimeoutMs
     this.time = time;
     this.simpleConsumerFactory = simpleConsumerFactory;
 
-    simpleConsumers = new HashMap<String, SimpleConsumer>();
+    simpleConsumers = new HashMap<String, Consumer<byte[], byte[]>>();
```
ewencp: We can clean this up with Java 7 diamond syntax instead of the actual types.

src/main/java/io/confluent/kafkarest/TPConsumerState.java
```diff
@@ -15,29 +15,36 @@
...
 
-  public FetchResponse fetch(final FetchRequest request) {
-    return consumer.fetch(request);
+  public Consumer<byte[], byte[]> consumer() {
+    return consumer;
   }
 
   public void close() throws Exception {
+    // release partition
+    consumer.unsubscribe();
```
ewencp: I think we might not want to unsubscribe here because of the way the consumer works. It's probably likely that we'll see another request for the same topic partition and if we unsubscribe, the consumer will always redo the metadata fetch. Since assignments just replace previous ones anyway, we can never call unsubscribe and we'll get bit of an optimization.

To fully take advantage of it, we'd need to also keep track of the partition previously assigned in the pool of consumers so they can be reused as effectively as possible.

src/main/java/io/confluent/kafkarest/SimpleConsumerRecordsCache.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ConcurrentMap;
+
+
+
+public class SimpleConsumerRecordsCache {
+
+  private final KafkaRestConfig config;
+  private final int maxPollTime;
```
ewencp: Seems `maxPollTime` and `time` are not used?

src/main/java/io/confluent/kafkarest/SimpleConsumerRecordsCache.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.Queue;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ConcurrentMap;
+
+
+
+public class SimpleConsumerRecordsCache {
```
ewencp: I'm curious about the need for a cache here. What's the usage pattern that would make this effective? I'm curious because adding a cache necessarily adds configs (e.g. you at least need to be able to specify the size) and I'm not sure that consumers in general should expect to benefit from caching the data they read.

Is there a use case you're thinking of specifically which would benefit from this caching? Of course, generically there is any case where there are multiple readers of the same data, but should we expect that to be especially common? Perhaps a bit more common with simple consumers than with consumer groups, but enough to warrant a cache?

GRpro: The cache should improve performance in case end user sequentially fetches limited batches of records from a specific partition. The extra records that were fetched are stored within cache. Say topic "events" has 3 partitions. The last record of partition 0 has offset 9999.
The `simpleconsumer.cache.max.records=1000`. User requests `/topic/events/partitions/0/messages&offset=5000?count=1000` and the `KafkaConsumer` fetches internally 5000 records. Then records from offset 6000 to 6999 are stored within the cache. When the user will request records sequentially for the next time we minimize invocation of `KafkaConsumer.poll()`.

GRpro:
I've try to fit the patch according to the above comments. For now I removed cache implementation to make the code more clear remaining a door for future implementation (possible in this PR).
Please take a look and provide a feedback.

docs/config.rst
```diff
@@ -62,8 +69,15 @@ Java Kafka clients.
   * Default: ""
   * Importance: medium
 
+``simpleconsumer.cache.max.records``
```
ewencp: I think this needs to be regenerated since this is a leftover cache setting.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -129,6 +140,12 @@
           + " Use 0 for no timeout";
   public static final String SIMPLE_CONSUMER_POOL_TIMEOUT_MS_DEFAULT = "1000";
 
+  public static final String SIMPLE_CONSUMER_MAX_POLL_TIME_CONFIG = "simpleconsumer.max.poll.time";
```
ewencp: Do we want to continue using the "simple consumer" nomenclature? These setting is user-facing and I think might be the only user-facing place we mention the simple consumer. The reason I am wondering is that the simple/high-level naming is based on the old consumers and has always been pretty confusing for users as the names are really from the perspective of Kafka developers (SimpleConsumer is anything but simple for users).

More generally, will these just end up being normal consumer settings as well when we convert the rest of the code to the new consumer? In which case maybe we can just prefix them with `consumer`? Or come up with some other naming which more clearly explains the difference between the manual partition assignment and group-based consumers?

/cc @hachikuji

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -15,108 +15,83 @@
...
 
-    // Load decoders
-    Properties props = new Properties();
-    props.setProperty("schema.registry.url",
-        config.getString(KafkaRestConfig.SCHEMA_REGISTRY_URL_CONFIG));
-    avroDecoder = new KafkaAvroDecoder(new VerifiableProperties(props));
+    ConsumerFactory<byte[], byte[]> consumerFactory = simpleConsumerFactory;
 
-    binaryDecoder = new DefaultDecoder(new VerifiableProperties());
+    if (consumerFactory == null) {
```
ewencp: It seems that this code path is never used? This constructor is only instantiated from one place and it passes in a non-null `simpleConsumerFactory`

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -15,108 +15,83 @@
...
 
-  private SimpleFetcher getSimpleFetcher(final Broker broker) {
-    // When upgrading to Java 1.8, use simpleConsumersPools.computeIfAbsent() instead
-    SimpleConsumerPool pool = simpleConsumersPools.get(broker);
-    if (pool == null) {
-      simpleConsumersPools.putIfAbsent(broker, createSimpleConsumerPool());
-      pool = simpleConsumersPools.get(broker);
+        @Override
+        public Consumer<byte[], byte[]> createConsumer() {
+          return new KafkaConsumer<byte[], byte[]>(consumerProperties,
```
ewencp: nit: can use diamond syntax and omit the generic types

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -15,108 +15,83 @@
...
-    }
+    simpleConsumersPool =
+      new SimpleConsumerPool(maxPoolSize,
+          poolInstanceAvailabilityTimeoutMs, maxPollTime, time, consumerFactory);
+
+    // Load deserializers
+    Map<String, String> props = new HashMap<>();
+    props.put("schema.registry.url", config.getString(KafkaRestConfig.SCHEMA_REGISTRY_URL_CONFIG));
+    avroDeserializer = new KafkaAvroDeserializer();
+    avroDeserializer.configure(props, true);
```
ewencp: It doesn't have any impact today, but you might want to use separate deserializers for key and value just to make sure this code is future-proof in case the deserializer ever *does* start relying on the `isKey` flag.

src/main/java/io/confluent/kafkarest/SimpleConsumerManager.java
```diff
@@ -126,155 +101,93 @@ public void consume(final String topicName,
...
     List<ConsumerRecord> records = null;
     RestException exception = null;
-    SimpleFetcher simpleFetcher = null;
 
-    try {
-      final Broker broker = mdObserver.getLeader(topicName, partitionId);
-      simpleFetcher = getSimpleFetcher(broker);
+    if (!mdObserver.topicExists(topicName)) {
+      exception = Errors.topicNotFoundException();
+    } else
```
ewencp: Code formatting -- the way this `else if` is broken up is misleading

src/main/java/io/confluent/kafkarest/SimpleConsumerPool.java
```diff
@@ -34,37 +35,192 @@
...
+
+    private Consumer<byte[], byte[]> consumer;
+    private final SimpleConsumerPool ownerPool;
+    private final long maxPollTime;
+    private boolean initialized;
+    private Time time;
+    private final TopicPartition topicPartition;
+
+    RecordsFetcher(TopicPartition topicPartition) {
+      this.ownerPool = SimpleConsumerPool.this;
```
ewencp: nit: Given that these are always available to non-`static` nested classes, is there really a need for storing them separately as fields?

src/main/java/io/confluent/kafkarest/SimpleConsumerPool.java
```diff
@@ -34,37 +35,192 @@
...
+      this.time = SimpleConsumerPool.this.time;
+      this.maxPollTime = SimpleConsumerPool.this.maxPollTime;
+      this.topicPartition = topicPartition;
+      initialized = false;
+    }
+
+    // support for lazy initialization gives opportunity
+    // to add some kind of cache in future.
+    private Consumer<byte[], byte[]> initializeConsumer(long offset) {
+      if (!initialized) {
```
ewencp: Do you need the separate `initialized` field? Why not just check if `consumer` is `null`?

src/main/java/io/confluent/kafkarest/SimpleConsumerPool.java
```diff
@@ -34,37 +35,192 @@
...
+
+        if (it == null || !it.hasNext()) {
+
+          if (!initialized) {
+            initializeConsumer(offset);
+          }
+
+          ConsumerRecords<byte[], byte[]> records = consumer
+              .poll(Math.max(0, pollTime));
+          if (records.isEmpty()) {
```
ewencp: This isn't necessary as the behavior is the same even if you get the iterator and execute the loop below (which will just exit immediately).

src/main/java/io/confluent/kafkarest/SimpleConsumerPool.java
```diff
@@ -34,37 +35,192 @@
...
-        return new SimpleFetcher(simpleConsumers.get(consumerId), this);
+
+      // If there is a KafkaConsumer available
+      if (!topicPartitionOffsetConsumers.isEmpty()) {
+        Map<Long, List<Consumer<byte[], byte[]>>> consumersMap =
+          topicPartitionOffsetConsumers.get(topicPartition);
+
+        Consumer<byte[], byte[]> consumer;
+
+        if (consumersMap != null) {
```
ewencp: This gets rather complicated. It might be worth trying to break it down into a couple of helper methods. A bit of the logic also seems duplicated between the two (i.e. the parts labeled "get the longest waiting consumer".

src/main/java/io/confluent/kafkarest/SimpleConsumerPool.java
```diff
@@ -81,22 +237,43 @@ synchronized public SimpleFetcher get(final String host, final int port) {
...
-
   }
 
-  synchronized public void release(SimpleFetcher simpleFetcher) {
-    log.debug("Releasing into the pool SimpleConsumer with id " + simpleFetcher.clientId());
-    availableConsumers.add(simpleFetcher.clientId());
+  synchronized public void release(Consumer<byte[], byte[]> consumer, TopicPartition topicPartition, long offset) {
+    Map<Long, List<Consumer<byte[], byte[]>>> offsetsMap =
+      topicPartitionOffsetConsumers.get(topicPartition);
+    if (offsetsMap != null) {
```
ewencp: This seems more complicated than it needs to be. Can't we just add a `new LinkedList<>()` to `topicPartitionOffsetConsumers` if it isn't there, then do the same for `offsetsMap` and `consumers`, then finally do a single `consumers.add(consumer)`? Here it seems we're repeating a number of steps by breaking down every case into its own block.

src/main/java/io/confluent/kafkarest/entities/ConsumerRecord.java
```diff
@@ -27,19 +27,34 @@
...
   protected long offset;
 
+  public ConsumerRecord(K key, V value, String topic, int partition, long offset) {
+    this.key = key;
+    this.value = value;
+    this.topic = topic;
+    this.partition = partition;
+    this.offset = offset;
+  }
+
```
ewencp: Won't the below constructor now be invalid since the `topic` should be `@NotNull` but `this.topic` is not set? I think the existing constructors would need to be replaced. I think this shouldn't matter anyway, a quick check suggests they aren't used anymore anyway.

src/main/java/io/confluent/kafkarest/entities/ConsumerRecord.java
```diff
@@ -27,19 +27,34 @@
...
     this.value = value;
     this.partition = partition;
     this.offset = offset;
   }
 
+  public ConsumerRecord(String topic, int partition, long offset) {
+    this(null, null, topic, partition, offset);
+  }
+
   public ConsumerRecord(int partition, long offset) {
```
ewencp: This constructor too.

src/test/java/io/confluent/kafkarest/mock/MockKafkaConsumer.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+
+public class MockKafkaConsumer implements Consumer<byte[], byte[]> {
```
ewencp: There's also a `MockConsumer` included with Kafka (that ships in the main clients jar). Would that be sufficient  so we could avoid maintaining a separate mock implementation here?

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
cpuchakayaka: I tend to agree with @ewencp regarding the "simple consumer" nomenclature. With the new consumer, since there is only one consumer api, is it more appropriate to have one rest resource for all the new consumer features, rather than splitting into two separate resources one for manual partition assignment and group-based consumers.

ewencp: @cpuchakayaka Hmm, that actually wasn't what I was suggesting. The producer and consumer APIs for the Java clients aren't intended to be RESTful and I don't think the mapping has to be direct. I actually think having the simple consumer style reads on the `TopicsResource` or `PartitionsResource` is quite intuitive for users and probably the right thing to do, whereas consumer groups really are a special type of resource that operate differently. I was just suggesting that user-facing settings that include the "simple" name will eventually be confusing as the old consumer APIs will eventually be deprecated and removed and newer users won't understand what "simple" means in this context.

GRpro:
Please let me know if I've missed anything to include from your comments.
Thanks.

cpuchakayaka:
Guys, can we get this rolling, we in need of 0.9 Kafka Consumer in rest proxy :-)

edoardocomar:
Any updates on merging this PR ?

bharatviswa504:
@ewencp @hachikuji 
Could you please let us know when this PR will be merged?

AndrewJSchofield:
Tidying up dormant PRs.

PR #238, opened 2016-09-12
REST-237 GetLeader do not need to get all  partition data of topic

sven0726:
- [ ] GetLeader  do not need to get all  partition data of topic

### DIFF
--- a/src/main/java/io/confluent/kafkarest/MetadataObserver.java
+++ b/src/main/java/io/confluent/kafkarest/MetadataObserver.java
@@ -158,19 +158,11 @@ public class MetadataObserver {
   }
 
   public int getLeaderId(final String topicName, final int partitionId) {
-    final List<Partition> partitions = getTopicPartitions(topicName);
-
-    if (partitions.size() == 0) {
-      throw Errors.topicNotFoundException();
+    final Partition partition = getTopicPartition(topicName, partitionId);
+    if (partition == null){
+      throw Errors.partitionNotFoundException();
     }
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/kafkarest/MetadataObserver.java
```diff
@@ -141,6 +141,9 @@ public boolean partitionExists(String topicName, int partition) {
 
   public Partition getTopicPartition(String topic, int partition) {
     List<Partition> partitions = getTopicPartitions(topic, partition);
+    if (partitions == null){
+      throw Errors.topicNotFoundException();
```
OopsOutOfMemory: 这里应该是
Errors.partitionNotFoundException()

CrazyJvm:
already running in production based on this patch.

OopsOutOfMemory:
It would be better to **throw Errors.topicNotFoundException();** in method **getTopicPartitions** than throw it each caller side.

throw the exception at this line:
https://github.com/confluentinc/kafka-rest/pull/238/files#diff-451260248c6513e346c62f60fb450d2cR160

sven0726:
@OopsOutOfMemory   sounds good，3ks

gf53520:
@ConfluentJenkins  By the way, I want to know why you get leader from zk, rather than using SimpleConsumer. Whether the way can be faster than the other way or not?

AndrewJSchofield:
Tidying up dormant PRs.

PR #242, opened 2016-09-21
Add incompatible error

nantani:
I need to know the compatibility is broken or not, because if compatibility is broken, and then I would like to produce to "error" topic.
In the current code, client cannot know why schema registration fails because error code "408" is very ambiguous.

Thanks,


nantani:
I have signed http://clabot.confluent.io/cla

PR #246, opened 2016-09-30
Apply cached schema id fix from master to 2.x

crumley:
Apply cached schema id fix from master to 2.x. Also tweak AvroRestProducerTest.testRepeatedProducer to ensure that it passes.


AndrewJSchofield:
Tidying up dormant PRs.

PR #253, opened 2016-11-01
Readme.md format changes , separates api's in the quick start.

ajayk:
You can see the md at https://github.com/ajayk/kafka-rest/blob/master/README.md


AndrewJSchofield:
I've taken these changes and adapted them to the current README. Thanks for the PR.

PR #255, opened 2016-12-02
clarified the key/value format limitations based on user questions

gwenshap:


gwenshap:
I'm pretty sure my doc change did not break the build :)

ewencp:
LGTM. Test failures are timing related due to the _schemas topic not being created fast enough. Not sure why it seems to have started happening a lot more consistently now.

PR #258, opened 2016-12-15
[WIP] Rest proxy security

hqin:
Here are the changes so far to integrate KafkaConsumer for the v2 API and security work:
* Added kafkarest/resources/v2/ConsumersResource.java to route all v2 API end points. 
   This is done to keep v1 and v2 API working side by side.
* kafkarest/v2/KafkaConsumerManager.java mimics  kafkarest/ConsumerManager.java
* Code works now with binary, json, and avro  using the following version formats:
    application/vnd.kafka.binary.v2+json
    application/vnd.kafka.json.v2+json
    application/vnd.kafka.avro.v2+json
* Working endpoints: create / delete consumers, read messages from a topic.
* Code in prototype quality, with no unit tests for the new endpoints yet.


src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,440 @@
...
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+
+import io.confluent.kafkarest.*; //compiling v1 versions
+
+/**
+ * Manages consumer instances by mapping instance IDs to consumer objects, processing read requests,
+ * and cleaning up when consumers disappear.
+ */
+public class KafkaConsumerManager {
```
ewencp: Most of this seems duplicated from v1. Is the assumption that it will eventually diverge enough that it is not worth sharing classes? Or you just want to work with this version to get things working with the new consumer and then work on refactoring the shared components?

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,440 @@
...
+        } else {
+	    consumer = consumerFactory.createConsumer(props);
+        }
+      } finally {
+      }
+
+      KafkaConsumerState state = null;
+      switch (instanceConfig.getFormat()) {
+        case BINARY:
+          state = new BinaryKafkaConsumerState(this.config, cid, consumer);
```
ewencp: Related to the above comment, I notice that these are all duplicated and share a different abstract base class for v2 compared to v1. Might be worth thinking about whether a separate class will be necessary (or easier), or if it's worth trying to share some of the implementation (since I'd imagine at least a few of the core operations will be shared, e.g. lifecycle stuff like expiration, polling for data, committing offsets, etc.. (I'm not really sure about this, but in some other tools like the command line tools for Kafka we have tried to unify the two APIs behind another layer of abstraction and it seems to have worked ok, see e.g. the console consumer for an example of this.)

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerTopicState.java
```diff
@@ -0,0 +1,84 @@
...
+import io.confluent.kafkarest.v2.KafkaConsumerReadTask;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+
+/**
+ * Tracks a consumer's state for a single topic, including the underlying stream and consumed and
+ * committed offsets. It provides manual synchronization primitives to support KafkaConsumerWorkers
+ * protecting access to the state while they process a read request in their processing loop.
+ */
+public class KafkaConsumerTopicState<KafkaK, KafkaV, ClientK, ClientV> {
```
ewencp: Given the suggested changes to the APIs and avoiding the topic URLs under a consumer in favor of just making subscription an explicit step (or part of instance creation), I think we can probably drop this class since each consumer will only need one instance of each of the pieces of data (i.e. the variables can probably just be lifted into `KafkaConsumerState`).

src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
```diff
@@ -0,0 +1,179 @@
+/**
```
ewencp: Is this basically just duplicating the v1 PartitionsResource? I didn't think we needed anything to change from there -- shouldn't we just put v2 on the existing resource?

src/main/java/io/confluent/kafkarest/v2/AvroKafkaConsumerState.java
```diff
@@ -0,0 +1,74 @@
...
+ * limitations under the License.
+ **/
+
+package io.confluent.kafkarest.v2;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import java.util.Properties;
+
+import io.confluent.kafka.serializers.KafkaAvroDecoder;
```
ewencp: There are some unused imports to be cleaned up here.

src/main/java/io/confluent/kafkarest/v2/AvroKafkaConsumerState.java
```diff
@@ -0,0 +1,74 @@
...
+                           ConsumerInstanceId instanceId,
+                           Consumer consumer) {
+    super(config, instanceId, consumer);
+    Properties props = new Properties();
+    props.setProperty("schema.registry.url",
+                      config.getString(KafkaRestConfig.SCHEMA_REGISTRY_URL_CONFIG));
+  }
+
+  @Override
+  protected Decoder<Object> getKeyDecoder() {
```
ewencp: Decoders aren't used for the new consumer, so we can get rid of these. The equivalent for the new consumer would be Deserializers.

Same for Binary and JSON versions of this class.

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,599 @@
...
+ * Manages consumer instances by mapping instance IDs to consumer objects, processing read requests,
+ * and cleaning up when consumers disappear.
+ */
+public class KafkaConsumerManager {
+
+  private static final Logger log = LoggerFactory.getLogger(KafkaConsumerManager.class);
+
+  private final KafkaRestConfig config;
+  private final Time time;
+  //private final String zookeeperConnect;
```
ewencp: We should just clean this out

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,599 @@
...
+      new PriorityQueue<KafkaConsumerState>();
+  private final ExpirationThread expirationThread;
+
+  public KafkaConsumerManager(KafkaRestConfig config, MetadataObserver mdObserver) {
+    this.config = config;
+    this.time = config.getTime();
+    //this.zookeeperConnect = config.getString(KafkaRestConfig.ZOOKEEPER_CONNECT_CONFIG);
+    this.bootstrapServers = config.getString(KafkaRestConfig.BOOTSTRAP_SERVERS_CONFIG);
+    this.mdObserver = mdObserver;
+    this.iteratorTimeoutMs = config.getInt(KafkaRestConfig.CONSUMER_ITERATOR_TIMEOUT_MS_CONFIG);
```
ewencp: Presumably we can get rid of this field since we're not relying on the old consumer iterator semantics anymore?

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,599 @@
...
+      props.setProperty("group.id", group);
+      // This ID we pass here has to be unique, only pass a value along if the deprecated ID field
+      // was passed in. This generally shouldn't be used, but is maintained for compatibility.
+      if (instanceConfig.getId() != null) {
+        props.setProperty("consumer.id", instanceConfig.getId());
+      }
+      // To support the old consumer interface with broken peek()/missing poll(timeout)
+      // functionality, we always use a timeout. This can't perfectly guarantee a total request
+      // timeout, but can get as close as this timeout's value
+      props.setProperty("consumer.timeout.ms", ((Integer) iteratorTimeoutMs).toString());
```
ewencp: Should drop this since new consumer doesn't support it.

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,599 @@
...
+    });
+  }
+
+  public ConsumerCommittedResponse committed(String group, String instance, ConsumerCommittedRequest request) {
+    log.debug("Committed offsets for consumer " + instance + " in group " + group);
+    ConsumerCommittedResponse response = new ConsumerCommittedResponse();	
+    KafkaConsumerState state = getConsumerInstance(group, instance);
+    if (state != null) {
+	response =  state.committed(request);
+
```
ewencp: There are a bunch of minor formatting things that could be cleaned up, e.g. these extra blank lines in these conditional blocks. Same across all these methods, e.g. `deleteConsumer` seems to already omit the check, but everything else seems to have it.

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,599 @@
...
+        }
+      }
+    });
+  }
+
+  public ConsumerCommittedResponse committed(String group, String instance, ConsumerCommittedRequest request) {
+    log.debug("Committed offsets for consumer " + instance + " in group " + group);
+    ConsumerCommittedResponse response = new ConsumerCommittedResponse();	
+    KafkaConsumerState state = getConsumerInstance(group, instance);
+    if (state != null) {
```
ewencp: I think you can skip this check since it looks like if it is missing `getConsumerInstance` will throw an exception.

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,599 @@
...
+
+  public ConsumerSubscriptionResponse subscription(String group, String instance) {
+    log.debug("Unscribing consumer " + instance + " in group " + group);
+    ConsumerSubscriptionResponse response = new ConsumerSubscriptionResponse();
+    KafkaConsumerState state = getConsumerInstance(group, instance);
+    if (state != null) {
+	java.util.Set<java.lang.String> topics =  state.subscription();
+	response.topics = new ArrayList<String>(topics);
+    } else {
+	response.topics = new ArrayList<String>();
```
ewencp: Yeah, I think this behavior is kind of confusing -- if you get a list of topics back in some cases, not getting it back in others is weird.I'd just drop the return type entirely since it doesn't provide any useful info to the user.

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
```diff
@@ -0,0 +1,232 @@
...
+      long startedIteration = parent.getConfig().getTime().milliseconds();
+      final int requestTimeoutMs = parent.getConfig().getInt(
+          KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG);
+      try {
+        // Read off as many messages as we can without triggering a timeout exception. The
+        // consumer timeout should be set very small, so the expectation is that even in the
+        // worst case, num_messages * consumer_timeout << request_timeout, so it's safe to only
+        // check the elapsed time once this loop finishes.
+
+        //FIXME :new api - just get all records for now..
```
ewencp: Maybe fix this up in a follow up PR?

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
```diff
@@ -0,0 +1,232 @@
...
+          }
+
+          //iter.next();
+          messages.add(recordAndSize.getRecord());
+          bytesConsumed += roughMsgSize;
+          // Updating the consumed offsets isn't done until we're actually going to return the
+          // data since we may encounter an error during a subsequent read, in which case we'll
+          // have to defer returning the data so we can return an HTTP error instead
+        }
+      } catch (ConsumerTimeoutException cte) {
```
ewencp: This exception shouldn't happen since the new consumer doesn't use an exception to indicate timeouts. Instead we should be checking for an empty list returned from poll().

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
```diff
@@ -0,0 +1,232 @@
...
+    this.maxResponseBytes = Math.min(
+        maxBytes,
+        parent.getConfig().getLong(KafkaRestConfig.CONSUMER_REQUEST_MAX_BYTES_CONFIG));
+    this.pollTimeout = timeout <= 0 ? 5000 : timeout;
+    this.callback = callback;
+    this.finished = new CountDownLatch(1);
+
+    started = parent.getConfig().getTime().milliseconds();
+    try {
+	topicState = parent.getOrCreateTopicState(topic, pollTimeout);
```
cpuchakayaka: Doing a poll inside the KafkaConsumerState.getOrCreateTopicState is not ideal here, it should be done in doPartialRead(), with some kinda strategy that check if timeout or max_bytes limit has it and finishes the execution. In a case of failed task, the messages from failed task are used first, and which inturn could mean whole batch of messages would be lost because of no internal queue.

src/main/java/io/confluent/kafkarest/resources/v2/ConsumersResource.java
```diff
@@ -0,0 +1,310 @@
...
+  @PerformanceMetric("consumer.records.read-binary+v2")
+  @Produces({Versions.KAFKA_V2_JSON_BINARY_WEIGHTED,
+             Versions.KAFKA_V2_JSON_WEIGHTED
+	      })
+  public void readRecordBinary(final @Suspended AsyncResponse asyncResponse,
+                              final @PathParam("group") String group,
+                              final @PathParam("instance") String instance,
+		              @QueryParam("timeout") @DefaultValue("-1") long timeout,
+                              @QueryParam("max_bytes") @DefaultValue("-1") long maxBytes) {
+      readRecords(asyncResponse, group, instance, timeout, maxBytes, BinaryKafkaConsumerState.class);
```
cpuchakayaka: If the request doesn't contain a timeout, default value **-1** is passed and no other default is set, which trickles down to consumer.poll() in KafkaConsumerState.getOrCreateTopicState(), as per the Consumer code in KafkaConsumer, it will throw a **IllegalArgumentException("Timeout must not be negative")** . More sensible default must be set

Similar thing with max_bytes, if the request doesn't set any, then a **-1** is set, in V1 ConsumersResource it does a **maxBytes = (maxBytes <= 0) ? Long.MAX_VALUE : maxBytes;**, similar can be done here

src/main/java/io/confluent/kafkarest/resources/v2/ConsumersResource.java
```diff
@@ -0,0 +1,310 @@
...
+  @GET
+  @Path("/{group}/instances/{instance}/records")
+  @PerformanceMetric("consumer.records.read-json+v2")
+  @Produces({Versions.KAFKA_V2_JSON_JSON_WEIGHTED_LOW}) // Using low weight ensures binary is default
+  public void readRecordJson(final @Suspended AsyncResponse asyncResponse,
+                            final @PathParam("group") String group,
+                            final @PathParam("instance") String instance,
+		            @QueryParam("timeout") @DefaultValue("-1") long timeout,
+                            @QueryParam("max_bytes") @DefaultValue("-1") long maxBytes) {
+      readRecords(asyncResponse, group, instance, timeout, maxBytes, JsonKafkaConsumerState.class);
```
cpuchakayaka: Same thing as about in regards to default value of timeout and max_bytes

src/main/java/io/confluent/kafkarest/resources/v2/ConsumersResource.java
```diff
@@ -0,0 +1,310 @@
...
+  @GET
+  @Path("/{group}/instances/{instance}/records")
+  @PerformanceMetric("consumer.records.read-avro+v2")
+  @Produces({Versions.KAFKA_V2_JSON_AVRO_WEIGHTED_LOW}) // Using low weight ensures binary is default
+  public void readRecordAvro(final @Suspended AsyncResponse asyncResponse,
+                            final @PathParam("group") String group,
+                            final @PathParam("instance") String instance,
+                            @QueryParam("timeout") @DefaultValue("-1") long timeout,
+                            @QueryParam("max_bytes") @DefaultValue("-1") long maxBytes) {
+      readRecords(asyncResponse, group, instance, timeout, maxBytes, AvroKafkaConsumerState.class);
```
cpuchakayaka: Same thing as about in regards to default value of timeout and max_bytes

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,546 @@
...
+    log.debug("Subscribing consumer " + instance + " in group " + group);
+    KafkaConsumerState state = getConsumerInstance(group, instance);
+    if (state != null) {
+	state.subscribe(subscription);
+    }
+
+  }
+
+  public void unsubscribe(String group, String instance) {
+    log.debug("Unscribing consumer " + instance + " in group " + group);
```
cpuchakayaka: Sorry being picky, spelling mistake

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -0,0 +1,546 @@
...
+    KafkaConsumerState state = getConsumerInstance(group, instance);
+    if (state != null) {
+	state.unsubscribe();
+    }
+
+  }
+
+  public ConsumerSubscriptionResponse subscription(String group, String instance) {
+    log.debug("Unscribing consumer " + instance + " in group " + group);
```
cpuchakayaka: Debug statement is confusing

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
```diff
@@ -0,0 +1,232 @@
...
+          bytesConsumed += roughMsgSize;
+          // Updating the consumed offsets isn't done until we're actually going to return the
+          // data since we may encounter an error during a subsequent read, in which case we'll
+          // have to defer returning the data so we can return an HTTP error instead
+        }
+      } catch (ConsumerTimeoutException cte) {
+        backoff = true;
+      }
+
+      long now = parent.getConfig().getTime().milliseconds();
```
cpuchakayaka: None of this code is really getting used, as the poll is done in the constructor. This needs to cleaned and refactored a bit to see if timeout or maxBytes is reached

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -0,0 +1,425 @@
...
+  }
+
+    public KafkaConsumerTopicState<KafkaK, KafkaV, ClientK, ClientV> getOrCreateTopicState(String topic, long timeout) {
+
+    lock.writeLock().lock();
+    try {
+      if (topics == null) {
+        return null;
+      }
+      if (topic != null) {
```
cpuchakayaka: This is not needed as topic is always null, and don't see the need for topic in V2 consumer

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -0,0 +1,425 @@
...
+	  Map<String, Integer> subscriptions = new TreeMap<String, Integer>();
+	  subscriptions.put(topic, 1);
+
+	  consumer.subscribe(Arrays.asList(topic));
+      }
+      ConsumerRecords<ClientK, ClientV> consumerRecords = consumer.poll(timeout);
```
cpuchakayaka: Doing the poll here doesn't make sense, it must be done in doPartialRead() of KafkaConsumerReadTask. Not sure if you have a follow up PR where your planning to refactor this.

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerTopicState.java
cpuchakayaka: Agree with @ewencp

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
cpuchakayaka: If state is not found it should return a **Consumer Instance Not Found** error

docs/api.rst
```diff
@@ -1178,3 +1181,1395 @@ The brokers resource provides access to the current state of Kafka brokers in th
...
+      Accept: application/vnd.kafka.v2+json, application/vnd.kafka+json, application/json
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 204 No Content
+   
+
+.. http:post:: /consumers/(string:group_name)/instances/(string:instance)/assignments
```
ewencp: Same comment above about `PUT` possibly making sense here as well.

docs/api.rst
```diff
@@ -1178,3 +1181,1395 @@ The brokers resource provides access to the current state of Kafka brokers in th
...
+	       
+   Manually assign a list of partitions to this consumer.
```
ewencp: Also might want to mention that this does replacement.

docs/api.rst
```diff
@@ -1178,3 +1181,1395 @@ The brokers resource provides access to the current state of Kafka brokers in th
...
+
+   :param string group_name: The name of the consumer group
+   :param string instance: The ID of the consumer instance
+
+   :<jsonarr partitions: A list of partitions to assign to this consumer
```
ewencp: I think I mentioned this elsewhere in the review of the actual code, but it seems weird to just get back the same list you sent. We could just return a 200 and that's good enough.

ewencp: Oh, and in fact it seems that the example below shows that there is a 204 with No Content. Does that reflect what the code does?

docs/api.rst
```diff
@@ -1178,3 +1181,1395 @@ The brokers resource provides access to the current state of Kafka brokers in th
...
+   :param string group_name: The name of the consumer group
+   :param string instance: The ID of the consumer instance
+
+   :>jsonarr partitions: A list of partitions manually to assign to this consumer
+   :>jsonarr string partitions[i].topic: Name of the topic
+   :>jsonarr int partitions[i].partition: Partition ID
+
+   :statuscode 404:
+      * Error code 40403 -- Consumer instance not found
+   :statuscode 500:
```
ewencp: Status code 500 isn't really worth repeating for every single endpoint. There's a general errors section near the top that explains general errors. And 500s generally mean you're not getting useful error info anyway, so if there isn't any info specific to this endpoint that can be given here, then it probably just adds noise.

docs/api.rst
```diff
@@ -1178,3 +1181,1395 @@ The brokers resource provides access to the current state of Kafka brokers in th
...
+          {
+            "topic": "test",
+            "partition": 1
+          }
+	  
+        ]
+      }
+
+
+.. http:post:: /consumers/(string:group_name)/instances/(string:instance)/positions
```
ewencp: Another thing we can put as a future improvement since I don't think it's useful *that* often, but it seems kinda weird to have this be an endpoint you can `POST` to, but can't `GET` to list the current position.

docs/api.rst
```diff
@@ -1178,3 +1181,1395 @@ The brokers resource provides access to the current state of Kafka brokers in th
...
+
+   Overrides the fetch offsets that the consumer will use for the next set of records to fetch.
+
+   :param string group_name: The name of the consumer group
+   :param string instance: The ID of the consumer instance
+
+   :<jsonarr offsets: A list of offsets
+   :<jsonarr string offsets[i].topic: Name of the topic for
```
ewencp: sentence fragment. maybe just "Name of the topic", or if there was supposed to be more then "Name of the topic for the offset to seek to" or something like that?

docs/api.rst
```diff
@@ -1178,3 +1181,1395 @@ The brokers resource provides access to the current state of Kafka brokers in th
...
+      }
+      
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 204 No Content
+
+
+.. http:post:: /consumers/(string:group_name)/instances/(string:instance)/positions/beginning
```
ewencp: Hmm, didn't think of this before, but should we consider making the `/positions` endpoint accept slightly more flexible format -- either the offset as a long or `"beginning"` and `"end"`. Seems like it'd be slightly more RESTful (`GET` on `/beginning` and `/end` seem to make sense, but `POST` is actually modifying the parent `/positions` resource) and would also reduce the API surface area.

That said, I'm not sure how easy it'd be to get that mixed-type field working.

docs/api.rst
```diff
@@ -1178,3 +1181,1395 @@ The brokers resource provides access to the current state of Kafka brokers in th
...
+
+   :param string group_name: The name of the consumer group
+   :param string instance: The ID of the consumer instance
+
+   :<jsonarr partitions: A list of partitions 
+   :<jsonarr string partitions[i].topic: Name of the topic
+   :<jsonarr int partitions[i].partition: Partition ID
+ 
+   :statuscode 404:
+      * Error code 40403 -- Consumer instance not found
```
ewencp: There are probably some other error codes as well, although I'm not sure off the top of my head what they'd get converted to. One example would be if you weren't assigned one of the partitions you include here. You'd get an `IllegalStateException` in Java, not sure if that will get converted to anything useful in the REST response (or if we need to surround a bunch of these APIs to convert them to something better than a 500).

docs/monitoring.rst
```diff
@@ -112,3 +114,50 @@ endpoint name ``brokers.list`` with the metric name ``request-rate`` to get
...
+  For API v2:
+  
+  ==================================  ======================================================================
+  ``consumer.commit-offsets+v2``      ``POST /consumers/{group}/instances/{instance}/offsets``
+                                      with ``Content-Type: application/vnd.kafka.v2+json`` header
+  ``consumer.committed-offsets+v2``   ``GET /consumers/{group}/instances/{instance}/offsets``
+                                      with ``Accept: application/vnd.kafka.v2+json`` header
+  ``consumer.create+v2``              ``POST /consumers/{group}`` 
+                                      with ``Content-Type: application/vnd.kafka.v2+json`` header
+  ``consumer.delete``                 ``DELETE /consumers/{group}/instances/{instance}``
```
ewencp: I think this should be `consumer.delete+v2`. Same `+v2` needed on the next few as well.

src/main/java/io/confluent/kafkarest/Errors.java
```diff
@@ -93,7 +93,16 @@ public static RestException consumerAlreadyExistsException() {
                              CONSUMER_ALREADY_EXISTS_ERROR_CODE);
   }
 
+  public final static String ILLEGAL_STATE_MESSAGE = "Illegal state: ";
+  public final static int ILLEGAL_STATE_ERROR_CODE = 40903;
 
+  public static RestException illegalStateException(Throwable t) {
```
ewencp: Ah, I missed this when commenting about `IllegalStateException` above -- 409 makes perfect sense.

src/main/java/io/confluent/kafkarest/resources/v2/ConsumersResource.java
```diff
@@ -0,0 +1,310 @@
...
+  @PerformanceMetric("consumer.subscribe+v2")
+  public void subscribe( @javax.ws.rs.core.Context UriInfo uriInfo, 
+			 final @PathParam("group") String group,
+			 final @PathParam("instance") String instance,
+			 @Valid ConsumerSubscriptionRecord subscription
+			) {
+      try {
+	  ctx.getKafkaConsumerManager().subscribe(group, instance, subscription);
+      } catch(java.lang.IllegalStateException e) {
+	  throw Errors.illegalStateException(e);
```
ewencp: If you don't want to have to wrap every case where an `IllegalStateException` is possible with code like this, you can use an `ExceptionMapper` instead. See `ZkExceptionMapper` for an example -- it's pretty simple to setup.

hqin:
@ewencp The code can be merged into 3.2.x if you approve. Thanks

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -0,0 +1,451 @@
...
+        }
+      }
+    } finally {
+      lock.writeLock().unlock();
+    }
+  }
+
+  public ConsumerRecord<ClientK, ClientV> peek() {
+    if (hasNext()) {
+      ConsumerRecord<ClientK, ClientV> record = consumerRecordList.get(this.index);
```
cpuchakayaka: I don't see any reason why a Java queue can't be used instead of list and building peek() and next() functions on top a list

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -0,0 +1,451 @@
...
+    return null;
+  }
+
+  public boolean hasNext() {
+    return consumerRecordList != null && this.index < consumerRecordList.size();
+  }
+
+  public ConsumerRecord<ClientK, ClientV> next() {
+    if (hasNext()) {
+      ConsumerRecord<ClientK, ClientV> record = consumerRecordList.get(index);
```
cpuchakayaka: I don't see any reason why a Java queue can't be used instead of list and building peek() and next() functions on top a list

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -0,0 +1,451 @@
...
+ * MessageAndMetadata<K,V> values to ConsumerRecords that can be returned to the client (including
+ * translation if the decoded Kafka consumer type and ConsumerRecord types differ).
+ */
+public abstract class KafkaConsumerState<KafkaK, KafkaV, ClientK, ClientV>
+    implements Comparable<KafkaConsumerState> {
+
+  private KafkaRestConfig config;
+  private ConsumerInstanceId instanceId;
+  private Consumer consumer;
+  private ConsumerRecords<ClientK, ClientV> consumerRecords = null;
```
cpuchakayaka: I don't see any reason why a Java queue can't be used instead of a list and building peek() and next() functions on top a list

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -0,0 +1,451 @@
...
+  public ConsumerRecord<ClientK, ClientV> peek() {
+    if (hasNext()) {
+      ConsumerRecord<ClientK, ClientV> record = consumerRecordList.get(this.index);
+      return record;
+    }
+    return null;
+  }
+
+  public boolean hasNext() {
+    return consumerRecordList != null && this.index < consumerRecordList.size();
```
cpuchakayaka: I don't see any reason why a Java queue can't be used instead of list and building peek(), hasNext() and next() functions on top a list

simplesteph:
Hi @hqin
Thanks so much for your work we're building against your branch and testing the proxy today 
Quick question , does this support clients passing their own keytab / principals to the rest proxy or will it in the future ? I'm wondering how to permission the rest proxy with the current model. It seems I'll need a rest proxy instance per application who wants to use it in order to truly secure access to topics

gwenshap:
@simplesteph You are right that currently, the entire REST Proxy authenticates as one user. We are planning on improving this in the future. Stay tuned :)

PR #263, opened 2016-12-28
Fix link to download page in readme

JordyMoos:


JordyMoos:
Where can i find this CLA?

ewencp:
@JordyMoos Sorry, you can find it here: http://clabot.confluent.io/cla

Is this actually a problem? Happy to update it, but am curious if the redirect from the original URL isn't working for you since we'd like to fix that as well if it is a problem.

JordyMoos:
@ewencp The issue seemed to be fixed at th site of confluent.io/
The issue was not the http to http*s* but it was download to download*s* which brought me to a 404 page

Maybe confluent.io/ added a redirect rule already and this PR can be closed

PR #265, opened 2017-01-18
Follow up for KIP-103 changes that fixes ProducerPool's extraction of endpoints to use the security protocol in the URL instead of the listener name.

ewencp:


ewencp:
@ijuma @theduderog Same deal as confluentinc/schema-registry#474. In this case there isn't an explicit option for setting bootstrap servers because you can choose to override any producer configs, so not sure where we'd document the limitation that you need to explicitly set the bootstrap servers if you have a cluster where only some endpoints will be accessible to REST proxy. In SimpleConsumer the previous build fix was ok since we're filtering to PLAINTEXT and only passing host/port to the SimpleConsumer.

src/main/java/io/confluent/kafkarest/ProducerPool.java
```diff
@@ -143,7 +144,8 @@ private static String getBootstrapBrokers(ZkUtils zkUtils) {
         if (bootstrapBrokers.length() > 0) {
           bootstrapBrokers += ",";
         }
-        bootstrapBrokers += ep.connectionString();
+        String hostport = ep.host() == null ? ":" + ep.port() : Utils.formatAddress(ep.host(), ep.port());
+        bootstrapBrokers += ep.securityProtocol() + "://" + hostport;
```
theduderog: this is legit for the bootstrap servers config param?

theduderog: NM.  Found it:  https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L52-L54

Question:  any particular reason to add the protocol prefix?

theduderog: Ooops.  Looking more closely, it accepts backslashes but not forward slashes.  Go figure.

ijuma: It works with and without as discussed in https://github.com/confluentinc/schema-registry/pull/474/files/2c3cf54751e0991140529859096c1f3fc6938719#diff-949e07df8a5fc0851347cd6e6a971a84 .

PR #273, opened 2017-02-25
Rest proxy tool fix

hqin:
@ewencp @sriramsub: Please review the change. This change is to make NativeVsRestConsumerPerformance.test to pass, as we added a field "topic" to the records retrieved by the consumers.


hqin:
Create a new one with a single commit. This is no longer needed.

PR #289, opened 2017-04-15
added statuscode 506 and error code 40601

gschmutz:
additional error added, which occurs, when the embedded data format does not match the deserializer specified in the POST /consumers/ call.


gschmutz:
not a general error, missread documentation

PR #296, opened 2017-05-03
Interceptor configuration options

hqin:


docs/config.rst
```diff
@@ -525,3 +525,27 @@ In addition to these configurations:
   * Importance: low
 
 
+Interceptor Configuration Options
+==============================
```
ewencp: sphinx will complain if the underline is not as long as the section title

docs/config.rst
```diff
@@ -525,3 +525,27 @@ In addition to these configurations:
...
+``consumer.interceptor.classes``
+  Consumer interceptor classes.
+
+  * Type: string
+  * Default: ""
+  * Importance: low
+    
+For example to enable Confluent Control Center monitoring interceptors:
+
+``consumer.interceptor.classes=io.confluent.monitoring.clients.interceptor.MonitoringConsumerInterceptor``
```
ewencp: Could we add these as commented out examples in `config/kafka-rest.properties` as well so it's easy for people to add them there? Maybe with a small comment above them about what they do?

hqin: Done

PR #297, opened 2017-05-04
use the common pom + style checking

norwood:


norwood:
retest this please

mhowlett:
LGTM, except note my experience with maven is limited and I could have missed some implication of changes made to the pom.xml file.

PR #301, opened 2017-05-10
Custom Context - CLIENTS-293

mageshn:
Ability to add resources like filter through a custom extension and inject custom context


src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -62,70 +69,39 @@ public void setupResources(Configurable<?> config, KafkaRestConfig appConfig) {
...
-    context =
-        new Context(
-            appConfig,
-            mdObserver,
-            producerPool,
-            consumerManager,
-            simpleConsumerManager,
-            kafkaConsumerManager
-        );
+    DefaultContextProvider.initializeDefaultContext(zkUtils, appConfig, mdObserver, producerPool,
```
mhowlett: is it necessary to make the default context here even when the `rest-proxy-context` property is set?

src/main/java/io/confluent/kafkarest/KafkaRestMain.java
```diff
@@ -50,8 +52,22 @@ public boolean verify(
...
+      /*TODO discuss with team and look at options to do something like Apache CLI here. Handle
+        exception for incorrect Class args with appropriate message*/
       KafkaRestConfig config = new KafkaRestConfig((args.length > 0 ? args[0] : null));
-      KafkaRestApplication app = new KafkaRestApplication(config);
+
+      RestResourceExtension restResourceExtension = null;
+      String extensionClassName = config.getString(KafkaRestConfig
+                                                       .KAFKA_REST_RESOURCE_EXTENSION_CONFIG);
+      if (StringUtil.isNotBlank(extensionClassName)) {
+        Class<RestResourceExtension>
```
mhowlett: Presumably this will fail if the configured class doesn't implement `RestResourceExtension`. Should we fail with a more descriptive error?

src/main/java/io/confluent/kafkarest/resources/ConsumersResource.java
```diff
@@ -55,23 +56,20 @@
...
   @Valid
   @Path("/{group}")
   @PerformanceMetric("consumer.create")
   public CreateConsumerInstanceResponse createGroup(
-      @javax.ws.rs.core.Context UriInfo uriInfo,
-      final @PathParam("group") String group,
-      @Valid ConsumerInstanceConfig config
-  ) {
+      @javax.ws.rs.core.Context UriInfo uriInfo, final @PathParam("group") String group,
+      @Valid ConsumerInstanceConfig config, @javax.ws.rs.core.Context Context ctx) {
```
mhowlett: My first thought here is that it would seem more natural to have only things specific to the request come through as parameters. i.e. have the request context come through, not the rest proxy context. To make this work, there'd need to be something like a RestContextManager. Usage here, something like:

`RestContextManager.GetContext(requestContext).getConsumerManager()`

This is more long winded than what you've got but on the upside it's descriptive of what's going on. 

In `KafkaRestApplication.setupInjectedResources()`, a `RestContextManager` could be created that is either a default or if the security plugin is found, the custom one. The `RestResourceExtension` class would need a new method `getRestContextManagerInstance()`, or something.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -297,318 +300,117 @@
...
-            METRICS_JMX_PREFIX_CONFIG,
-            Type.STRING,
-            METRICS_JMX_PREFIX_DEFAULT_OVERRIDE,
-            Importance.LOW,
-            METRICS_JMX_PREFIX_DOC
-        )
+    config = baseKafkaRestConfigDef();
+  }
+
+  protected static ConfigDef baseKafkaRestConfigDef() {
```
mhowlett: maybe put an entry in checkstyle/suppressions.xml for MethodLength instead of reformatting to something that is harder to read?

mhowlett:
there are many changes in this PR that are formatting (white space) only and most of these changes are moving away from existing conventions, making the codebase less consistent. These would be best reverted to make the PR as small as possible.

src/main/java/io/confluent/kafkarest/KafkaRestMain.java
mhowlett: also, restResourceExtension depends only on config, so it can be constructed inside the KafkaRestApplication constructor, which would be cleaner.

src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -16,30 +16,33 @@
...
 /**
  * Utilities for configuring and running an embedded Kafka server.
  */
 public class KafkaRestApplication extends Application<KafkaRestConfig> {
 
-  ZkUtils zkUtils;
-  Context context;
+  RestResourceExtension restResourceExtension;
 
   public KafkaRestApplication() throws RestConfigException {
```
mhowlett: Is there any point to having more than one constructor?

src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -73,59 +82,33 @@ protected void setupInjectedResources(
...
-    context =
-        new Context(
-            appConfig,
-            mdObserver,
-            producerPool,
-            consumerManager,
-            simpleConsumerManager,
-            kafkaConsumerManager
-        );
+    KafkaRestContextProvider.initializeDefaultContext(zkUtils, appConfig, mdObserver, producerPool,
```
mhowlett: Though not enforced, I believe we have a style preference for formatting long argument lists like the `ZKUtils.apply` call that was removed here (i.e. different indentation to what you have).

mhowlett: A better name for this would be just `initialize` I think. That it initializes a default context is kind of an implementation detail.

src/main/java/io/confluent/kafkarest/extension/ContextInvocationHandler.java
```diff
@@ -0,0 +1,43 @@
...
+import io.confluent.kafkarest.KafkaRestContext;
+
+
+public class ContextInvocationHandler implements InvocationHandler {
+
+  public ContextInvocationHandler() {
+  }
+
+  private KafkaRestContext getContext() {
+    if (KafkaRestContextProvider.getCurrentContext() != null) {
```
mhowlett: this logic relates only to information inside `KafkaRestContextProvider`, so would be better encapsulated in that class.

src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -49,10 +54,23 @@ public KafkaRestApplication(Properties props) throws RestConfigException {
...
+    String extensionClassName = config.getString(KafkaRestConfig
+                                                     .KAFKA_REST_RESOURCE_EXTENSION_CONFIG);
+    if (StringUtil.isNotBlank(extensionClassName)) {
+      Class<RestResourceExtension>
+          restResourceExtensionClass =
+          (Class<RestResourceExtension>) Class.forName(extensionClassName);
+
+      restResourceExtension = restResourceExtensionClass.newInstance();
+
+    }
```
mhowlett: i was thinking something like: 

```
    if (StringUtil.isNotBlank(extensionClassName)) {
 +      try {
 +        Class<RestResourceExtension>
 +            restResourceExtensionClass =
 +            (Class<RestResourceExtension>) Class.forName(extensionClassName);
 +
 +        this.restResourceExtension = restResourceExtensionClass.newInstance();
 +      }
 +      catch (ClassNotFoundException e)
 +      {
 +        throw new RestConfigException(
 +            "Unable to load resource extension class '"
 +            + extensionClassName
 +            + "'. Check your classpath and that the configured class implements the "
 +            + "RestResourceExtension interface.");
 +      }
 +    }
```

src/main/java/io/confluent/kafkarest/KafkaRestMain.java
```diff
@@ -50,7 +52,10 @@ public boolean verify(
    */
   public static void main(String[] args) throws IOException {
     try {
+      /*TODO discuss with team and look at options to do something like Apache CLI here. Handle
```
mhowlett: I think you can get rid of this comment, after making changes suggested in KafkaRestApplication.

src/main/java/io/confluent/kafkarest/KafkaRestContextImpl.java
```diff
@@ -21,7 +21,7 @@
 /**
  * Shared, global state for the REST proxy server, including configuration and connection pools.
  */
-public class Context {
+public class KafkaRestContextImpl implements KafkaRestContext {
```
mhowlett: perhaps `DefaultKafkaRestContext` or `BasicKafkaRestContext` or `SimpleKafkaRestContext`? i think maybe the last one is my favorite of the lot.

cpuchakayaka: SimpleKafkaRestContext could be a confusing as there are Simple* classes that define something else. (ex. SimpleConsumerManager, SimpleConsumerPool etc). I would go for  DefaultKafkaRestContext here

src/main/java/io/confluent/kafkarest/extension/KafkaRestContextProvider.java
```diff
@@ -0,0 +1,128 @@
...
+
+  public static KafkaRestConfig getDefaultAppConfig() {
+    return defaultAppConfig;
+  }
+
+  public static KafkaRestContext getCurrentContext() {
+    if (restContextInheritableThreadLocal.get() != null) {
+      return restContextInheritableThreadLocal.get();
+    } else {
+      return KafkaRestContextProvider.getDefaultContext();
```
mhowlett: why specify `KafkaRestContextProvider`?

src/main/java/io/confluent/kafkarest/extension/KafkaRestContextProvider.java
```diff
@@ -0,0 +1,128 @@
...
+      defaultZkUtils = zkUtils;
+      context =
+          new KafkaRestContextImpl(appConfig, mdObserver, producerPool, consumerManager,
+              simpleConsumerManager, kafkaConsumerManager
+          );
+      defaultAppConfig = appConfig;
+    }
+  }
+
+  public static KafkaRestContext getDefaultContext() {
```
mhowlett: there is something sub-optimal with the factoring here i think - I don't think the concept of defaultContext is best exposed. maybe it would be better if there was some different type that encapsulated all these objects in addition to DefaultContext. It'd be pretty similar, just but different purpose. or separate methods, though that is more verbose. maybe what's here is not a bad compromise.

mhowlett:
LTGM. Would prefer it if `KafkaRestContextProvider.getDefaultContext` didn't need to be exposed, but don't see an obvious refactoring.

rangwea:
@mageshn How I add a Servlet Filter by the RestResourceExtension Interface?

mageshn:
@rangwea you can register any kind of Jersey resources like Jersey Filter, provider, etc.
You can look at [https://jersey.github.io/documentation/latest/filters-and-interceptors.html#d0e9332](url)

PR #302, opened 2017-05-11
CLIENTS-304: Add test utility that logs entities if they fail to parse.

ewencp:
This will aid in debugging when unexpected responses are received.


mageshn:
LGTM

PR #303, opened 2017-05-12
CLIENTS-304: Fix AvroConsumerRecord constructors and handling of topic for v1 consumer API.

ewencp:
AvroConsumerRecord had two constructors with @JsonProperty parameters which caused it to sometimes fail to properly
deserialize. Additionally, adding the topic field to ConsumerRecord for the v2 API resulted in null entries in the v1
API. This is confusing; instead, we should be filling that in since it is a backwards compatible addition and keeps the
output uniform.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2440,12 +2440,14 @@ error.
           "value": "Y29uZmx1ZW50",
           "partition": 1,
           "offset": 100,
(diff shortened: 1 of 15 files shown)


mageshn:
LGTM

norwood:
😍

PR #306, opened 2017-05-12
Resolves #298, Introduce an AuthenticationMapping…

### LINKED ISSUE #298: Map Basic Auth to Kafka client Auth
For security reason, we are using Kafka with SASL_PLAIN and we did configure ACLs on topics to allow users to perform (or not) read/write operations.

Currently our users can publish messages to theirs topics using a Kafka client but we want to give them the ability to publish their messages over HTTPS + Basic Auth.

My idea was to create a "dynamic" pool.
We could start with a one to one mapping (ie. HTTP credentials == Kafka client credentials) or we could implement an interface to have a fine grained mapping.

The "dynamic" pool will need to create a new producer/consumer for every credentials.

Let's take an example:

Topic | ACL
------------ | -------------
topic_a | principal: user_a, operations: write<br>principal: admin, operations: write,read
topic_b | principal: user_b, operations: write<br>principal: admin, operations: write,read

`admin` can publish message on topic_a and topic_b
```
POST http://localhost:8082/topics/topic_a
admin:password
200
```
(create a new producer with `admin:password` credentials)
```
POST http://localhost:8082/topics/topic_b
admin:password
200
```
(reuse the producer with `admin:password` credentials)

`user_b` cannot publish message on `topic_a`
```
POST http://localhost:8082/topics/topic_a
user_b:password
403
```
(create a new producer with `user_b:password` credentials)

`user_b` can publish message on `topic_b`
```
POST http://localhost:8082/topics/topic_b
user_b:password
200
```
(reuse the producer with `user_b:password` credentials)

This feature requires some changes and before working on it, I want to make sure that this is something you will consider :wink: 

Let me know what you think!

Mogztter:
...to create an authenticated Kafka producer.

Here's a basic implementation:

BasicAuthenticationMapping.java
```java
public class DefaultAuthenticationMapping implements AuthenticationMapping {

  @Override
  public JaasModule getJaasModule(String userPrincipalName) {
    if (userPrincipalName == null) {
      return null;
    }
    Map<String, String> entries = new HashMap<>();
    switch (userPrincipalName) {
      case "admin":
        entries.put("username", "admin");
        entries.put("password", "admin1234");
        return new JaasModule("org.apache.kafka.common.security.plain.PlainLoginModule", true, entries);
      case "foo":
        entries.put("username", "bar");
        entries.put("password", "abcd");
        return new JaasModule("org.apache.kafka.common.security.plain.PlainLoginModule", true, entries);
      default:
        return null;
    }
  }
}
```

kafka-rest.properties
```
authentication.method=BASIC
authentication.realm=KafkaRestServer
authentication.roles=admin
authentication.mapping=com.github.confluentinc.kafkarest.DefaultAuthenticationMapping
```

### DIFF
--- a/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -97,6 +97,13 @@ public class KafkaRestConfig extends RestConfig {
       "Number of threads to run produce requests on.";
   public static final String PRODUCER_THREADS_DEFAULT = "5";
 
+  public static final String AUTHENTICATION_MAPPING_CONFIG = "authentication.mapping";
+  private static final String
+      AUTHENTICATION_MAPPING_DOC =
+      "Mapping class for client authentication that implements the "
+         + "<code>AuthenticationMapping</code> interface. "
+         + "Allow to map Kafka REST authentication to the Kafka client authentication.";
+
   public static final String CONSUMER_ITERATOR_TIMEOUT_MS_CONFIG = "consumer.iterator.timeout.ms";
   private static final String CONSUMER_ITERATOR_TIMEOUT_MS_DOC =
       "Timeout for blocking consumer iterator operations. "
@@ -370,6 +377,13 @@ public class KafkaRestConfig extends RestConfig {
             SCHEMA_REGISTRY_URL_DEFAULT,
             Importance.HIGH,
             SCHEMA_REGISTRY_URL_DOC
+        )
+        .define(
+            AUTHENTICATION_MAPPING_CONFIG,
+            Type.STRING,
+            "",
+            Importance.MEDIUM,
+            AUTHENTICATION_MAPPING_DOC
         )
         .define(
             PRODUCER_THREADS_CONFIG,
--- a/src/main/java/io/confluent/kafkarest/ProducerPool.java
+++ b/src/main/java/io/confluent/kafkarest/ProducerPool.java
@@ -16,30 +16,34 @@
 
 package io.confluent.kafkarest;
 
-import org.apache.kafka.clients.producer.KafkaProducer;
-import org.apache.kafka.clients.producer.ProducerConfig;
-import org.apache.kafka.common.serialization.ByteArraySerializer;
-import org.apache.kafka.common.serialization.Serializer;
+import static io.confluent.kafkarest.ProducerBuilder.buildAvroConfig;
+import static io.confluent.kafkarest.ProducerBuilder.buildAvroProducer;
+import static io.confluent.kafkarest.ProducerBuilder.buildBinaryProducer;
+import static io.confluent.kafkarest.ProducerBuilder.buildJsonProducer;
+import static io.confluent.kafkarest.ProducerBuilder.buildStandardConfig;
+
+import io.confluent.kafkarest.entities.EmbeddedFormat;
+import io.confluent.kafkarest.entities.PartitionOffset;
+import io.confluent.kafkarest.entities.ProduceRecord;
+import io.confluent.kafkarest.entities.ProduceResponse;
+import io.confluent.kafkarest.entities.SchemaHolder;
+import kafka.cluster.Broker;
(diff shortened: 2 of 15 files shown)


Mogztter:
[clabot:check]

src/main/java/io/confluent/kafkarest/ProducerPool.java
```diff
@@ -216,4 +270,45 @@ public void onCompletion(
         List<RecordMetadataOrException> results
     );
   }
+
+  public static ProducerPool.ProduceRequestCallback defaultProduceRequestCallback(final AsyncResponse asyncResponse, final Logger log) {
```
Mogztter: If you want I can revert this change since it's not mandatory but reduce code duplication (this method is used 3 times)

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -96,6 +96,12 @@
       "Number of threads to run produce requests on.";
   public static final String PRODUCER_THREADS_DEFAULT = "5";
 
+  public static final String AUTHENTICATION_MAPPING_CONFIG = "authentication.mapping";
```
cpuchakayaka: should have .class suffix

src/main/java/io/confluent/kafkarest/ProducerPool.java
```diff
@@ -186,16 +199,57 @@ private static String getBootstrapBrokers(ZkUtils zkUtils) {
...
+    RestProducer restProducer = getRestProducer(recordFormat, getJaasModule(securityContext));
     restProducer.produce(task, topic, partition, records);
   }
 
+  private RestProducer getRestProducer(EmbeddedFormat recordFormat, JaasModule jaasModule) {
+    RestProducer restProducer;
+    if (jaasModule != null) {
+      if (!authenticatedProducers.containsKey(jaasModule)) {
+        restProducer = buildAuthenticatedProducer(recordFormat, jaasModule);
+        authenticatedProducers.put(jaasModule, restProducer);
```
cpuchakayaka: each user principal will have its own producer, if we have many principals, like in our case we have 20/25 principals, we will not get a good throughput from the producer if they all producing same topic. If the max.buffer.size is high enough we could also run into memory issues with many producers.

Mogztter: True, do you have an idea to improve it ? 

I could create the producer "on demand" to reduce memory usage but producing messages will be slower since we will need to create a producer for each request.

> each user principal will have its own producer, if we have many principals, like in our case we have 20/25 principals, we will not get a good throughput from the producer if they all are producing to same topic.

As a workaround you could use the same principal to produce messages on a topic. For instance, user1, user2 and user3 are allowed to produce messages on topic A. If user1, user2 or user3 is authenticated then you could return a "generic" principal to produce messages on topic A.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
Mogztter: While I agree, I don't think there's a clear rule.
`key.serializer` is not suffix by class, nor `partition.grouper`, `value.serde` or `timestamp.extractor`.

Though it may be more explicit to add a `.class` suffix.

Mogztter:
Rebased on master and checkstyle issues are fixed.

Mogztter:
Confluent platform 3.3 is out!
Well done everyone 😉 

One cool feature in this release is the **REST Proxy ACL Plug-In**.
> We have added a security plugin module to REST Proxy that allows credentials of authenticated incoming requests to be propagated to Kafka and authorization enforced via Kafka ACLs.

http://docs.confluent.io/current/confluent-security-plugins/docs/kafka_rest_security.html

But I cant' find the code on GitHub for this plug-in.

@ewencp Is this plug-in closed source and part of the "Confluent Enterprise" ? Does that mean that my feature (which is basically the same) will not be merged in this "Confluent Open Source" project ?

PR #317, opened 2017-06-27
CLIENTS-491: Restore log4j dependency

ewencp:
The dependency was presumably lost during refactoring to share the common pom


mageshn:
LGTM

PR #322, opened 2017-07-20
Tiny code simplification

The-Alchemist:
Duplicate "else if" block because line 135 already checks `instanceof GenericContainer`.

Caught by IntelliJ.


ewencp:
ok to test

PR #324, opened 2017-07-20
Strings cannot be compared using !=

The-Alchemist:
Need to use `String::equals()`

Found using IntelliJ

### DIFF
--- a/src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetMetadata.java
+++ b/src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetMetadata.java
@@ -97,7 +97,7 @@ public class TopicPartitionOffsetMetadata {
     if (offset != that.offset) {
       return false;
     }
-    if (metadata != that.metadata) {
+    if (!metadata.equals(that.metadata)) {
       return false;
     }
     if (partition != that.partition) {
--- a/src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetMetadata.java
+++ b/src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetMetadata.java
@@ -97,7 +97,7 @@ public class TopicPartitionOffsetMetadata {
     if (offset != that.offset) {
       return false;
     }
-    if (!metadata.equals(that.metadata)) {
+    if (!Objects.equals(metadata, that.metadata)) {
       return false;
     }
     if (partition != that.partition) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetMetadata.java
```diff
@@ -97,7 +97,7 @@ public boolean equals(Object o) {
     if (offset != that.offset) {
       return false;
     }
-    if (metadata != that.metadata) {
```
mhowlett: Well spotted! But I think `Object.equals` will just do a reference comparison. 
Can you update metadata value comparison to be the same as the topic value comparison below? thanks! 

https://stackoverflow.com/questions/13001971/using-the-equals-method-with-string-and-object-in-java

ewencp:
add to whitelist

src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetMetadata.java
The-Alchemist: I feel you might be mistaken about `Objects.equals()`.  Here's how it's implemented on my JDK:

```java
    /**
     * Returns {@code true} if the arguments are equal to each other
     * and {@code false} otherwise.
     * Consequently, if both arguments are {@code null}, {@code true}
     * is returned and if exactly one argument is {@code null}, {@code
     * false} is returned.  Otherwise, equality is determined by using
     * the {@link Object#equals equals} method of the first
     * argument.
     *
     * @param a an object
     * @param b an object to be compared with {@code a} for equality
     * @return {@code true} if the arguments are equal to each other
     * and {@code false} otherwise
     * @see Object#equals(Object)
     */
    public static boolean equals(Object a, Object b) {
        return (a == b) || (a != null && a.equals(b));
    }

```

More info at https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)

In other words, it does a reference check as a shortcut (also conveniently works for `null`s), and then does a `.equals()` as one would expect.

I feel that expresses the intention much cleaner and more succinctly than what is done for `topic`, and it's a Java 7 API (as opposed to Java 8) so it preserves Java 7 compatibility.

ewencp: Yeah, this should be fine. tbh I'm not sure how this ended up incorrect, generally these are auto-generated by IntelliJ

PR #325, opened 2017-07-21
Update quickstart to use Confluent CLI.

kkonstantine:


docs/intro.rst
```diff
@@ -13,17 +13,27 @@ framework that doesn't yet support Kafka, and scripting administrative actions.
...
-   $ ./bin/schema-registry-start ./etc/schema-registry/schema-registry.properties &
+Each service reads its configuration from its property files under ``etc``.
+
+.. note::
+
+   To manually start each service in its own terminal, run instead:
+
+   .. sourcecode:: bash
+
+      $ bin/zookeeper-server-start ./etc/kafka/zookeeper.properties
```
ewencp: When listed like this without the `&`, you need to call out that each will need its own terminal.

kkonstantine: That's true. I've added the sentence: `To manually start each service in its own terminal, run instead:`

As it is right now, by putting the service on the background, the experience is not optimal. That is because each service prints messages on stdout and stderr. And that continues at various times even way after their startup. 

Thus I think running them in the background doesn't have the desired effect. By removing `&` even if users didn't notice our guideline to run every service in its own terminal, they'll be lead to this since the current terminal will be taken by the first service to run.

PR #332, opened 2017-08-09
CLIENTS-512 - Use Admin Client instead of ZkUtils for all metadata API

mageshn:
- Uses Admin Client instead of ZkUtils for all metadata API

- It is required to configure at least one of ZK connect url or boot strap brokers

- V1 Consumer API would work only if ZK connect URL is configured

- No proactive checks to topic exists or partition exists as that would require additional ACL's to be defined for the producer user. Making config easier compared to super friendly response message & code

### DIFF
--- a/config/kafka-rest.properties
+++ b/config/kafka-rest.properties
@@ -17,6 +17,7 @@
 #id=kafka-rest-test-server
 #schema.registry.url=http://localhost:8081
 #zookeeper.connect=localhost:2181
+bootstrap.servers=PLAINTEXT://localhost:9092
 #
 # Configure interceptor classes for sending consumer and producer metrics to Confluent Control Center
 # Make sure that monitoring-interceptors-<version>.jar is on the Java class path
--- a/src/main/java/io/confluent/kafkarest/AdminClientWrapper.java
+++ b/src/main/java/io/confluent/kafkarest/AdminClientWrapper.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2017 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.kafkarest;
+
+import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.admin.Config;
+import org.apache.kafka.clients.admin.ConfigEntry;
+import org.apache.kafka.clients.admin.DescribeClusterResult;
+import org.apache.kafka.clients.admin.TopicDescription;
+import org.apache.kafka.common.Node;
+import org.apache.kafka.common.TopicPartitionInfo;
+import org.apache.kafka.common.config.ConfigResource;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Properties;
+import java.util.TreeSet;
+import java.util.Vector;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import io.confluent.kafkarest.entities.Partition;
+import io.confluent.kafkarest.entities.PartitionReplica;
+import io.confluent.kafkarest.entities.Topic;
+import io.confluent.rest.exceptions.RestServerErrorException;
+import jersey.repackaged.com.google.common.collect.ImmutableList;
+
+
+public class AdminClientWrapper {
+
+  private AdminClient adminClient;
+  private int initTimeOut;
+
+  public AdminClientWrapper(KafkaRestConfig kafkaRestConfig) {
+    Properties properties = new Properties(kafkaRestConfig.getAdminProperties());
+    properties.put(KafkaRestConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaRestConfig.bootstrapBrokers());
+    adminClient = AdminClient.create(properties);
+    this.initTimeOut = kafkaRestConfig.getInt(KafkaRestConfig.KAFKACLIENT_INIT_TIMEOUT_CONFIG);
+  }
+
+  public List<Integer> getBrokerIds() {
+    List<Integer> brokerIds = new Vector<>();
+    DescribeClusterResult clusterResults = adminClient.describeCluster();
+    try {
+      Collection<Node>
+          nodeCollection =
+          clusterResults.nodes().get(initTimeOut, TimeUnit.MILLISECONDS);
+      for (Node node : nodeCollection) {
+        brokerIds.add(node.id());
+      }
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new RestServerErrorException(Errors.KAFKA_ERROR_MESSAGE, Errors
+          .KAFKA_ERROR_ERROR_CODE, e);
+    }
+    return brokerIds;
+  }
+
+  public Collection<String> getTopicNames() {
+    Collection<String> allTopics = null;
+    try {
+      allTopics =
+          new TreeSet<>(adminClient.listTopics().names().get(initTimeOut, TimeUnit.MILLISECONDS));
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new RestServerErrorException(Errors.KAFKA_ERROR_MESSAGE, Errors
+          .KAFKA_ERROR_ERROR_CODE, e);
+    }
+    return allTopics;
+  }
+
+  public boolean topicExists(String topic) {
+    try {
+      Collection<String> allTopics =
+          adminClient.listTopics().names().get(initTimeOut, TimeUnit.MILLISECONDS);
+      return allTopics.contains(topic);
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new RestServerErrorException(Errors.KAFKA_ERROR_MESSAGE, Errors
+          .KAFKA_ERROR_ERROR_CODE, e);
+    }
+  }
+
+  public Topic getTopic(String topicName) {
+    Topic topic = null;
+    try {
+      if (topicExists(topicName)) {
+        TopicDescription topicDescription = adminClient.describeTopics(
+            ImmutableList.<String>of(topicName)).values().get(topicName)
+            .get(initTimeOut, TimeUnit.MILLISECONDS);
+
+        topic = buildTopic(topicName, topicDescription);
+      }
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new RestServerErrorException(Errors.KAFKA_ERROR_MESSAGE, Errors
+          .KAFKA_ERROR_ERROR_CODE, e);
+    }
+    return topic;
+  }
+
+  public List<Partition> getTopicPartitions(String topicName) {
+    List<Partition> partitions = null;
+    try {
+      TopicDescription topicDescription = adminClient.describeTopics(
+          ImmutableList.<String>of(topicName)).values().get(topicName)
+          .get(initTimeOut, TimeUnit.MILLISECONDS);
+
+      partitions = buildPartitonsData(topicDescription.partitions(), null);
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new RestServerErrorException(Errors.KAFKA_ERROR_MESSAGE, Errors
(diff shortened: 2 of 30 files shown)


config/kafka-rest.properties
```diff
@@ -17,6 +17,7 @@
 #id=kafka-rest-test-server
 #schema.registry.url=http://localhost:8081
 #zookeeper.connect=localhost:2181
+bootstrap.servers=PLAINTEXT://localhost:9092
```
mhowlett: other properties in this config are commented out, should this be too?

src/main/java/io/confluent/kafkarest/AdminClientWrapper.java
```diff
@@ -0,0 +1,197 @@
...
+    properties.put(KafkaRestConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaRestConfig.bootstrapBrokers());
+    adminClient = AdminClient.create(properties);
+    this.initTimeOut = kafkaRestConfig.getInt(KafkaRestConfig.KAFKACLIENT_INIT_TIMEOUT_CONFIG);
+  }
+
+  public List<Integer> getBrokerIds() {
+    List<Integer> brokerIds = new Vector<>();
+    DescribeClusterResult clusterResults = adminClient.describeCluster();
+    try {
+      Collection<Node>
```
mhowlett: general comment: i often find your formatting inconsistent and unusual. Here, I'd put the `nodeCollection =` variable name on line 60 (consistent with what you do on line 87.

src/main/java/io/confluent/kafkarest/AdminClientWrapper.java
```diff
@@ -0,0 +1,197 @@
...
+          nodeCollection =
+          clusterResults.nodes().get(initTimeOut, TimeUnit.MILLISECONDS);
+      for (Node node : nodeCollection) {
+        brokerIds.add(node.id());
+      }
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new RestServerErrorException(Errors.KAFKA_ERROR_MESSAGE, Errors
```
mhowlett: here, why split on `.KAFKA_ERROR_ERROR_CODE`?. I'd split after ( then ,.

src/main/java/io/confluent/kafkarest/AdminClientWrapper.java
```diff
@@ -0,0 +1,197 @@
...
+          .KAFKA_ERROR_ERROR_CODE, e);
+    }
+    return brokerIds;
+  }
+
+  public Collection<String> getTopicNames() {
+    Collection<String> allTopics = null;
+    try {
+      allTopics =
```
mhowlett: here, it would be more usual (and consistent with other code) to put new `TreeSet<>(` on this line and break on (

src/main/java/io/confluent/kafkarest/AdminClientWrapper.java
```diff
@@ -0,0 +1,197 @@
...
+  private int initTimeOut;
+
+  public AdminClientWrapper(KafkaRestConfig kafkaRestConfig) {
+    Properties properties = new Properties(kafkaRestConfig.getAdminProperties());
+    properties.put(KafkaRestConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaRestConfig.bootstrapBrokers());
+    adminClient = AdminClient.create(properties);
+    this.initTimeOut = kafkaRestConfig.getInt(KafkaRestConfig.KAFKACLIENT_INIT_TIMEOUT_CONFIG);
+  }
+
+  public List<Integer> getBrokerIds() {
```
mhowlett: `getTopicNames` returns a `Collection`. That would probably be more appropriate here too, and consistent, but I see this is just transferred over from a previous implementation and not a big deal at all so no need to bother changing.

src/main/java/io/confluent/kafkarest/AdminClientWrapper.java
```diff
@@ -0,0 +1,197 @@
...
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new RestServerErrorException(Errors.KAFKA_ERROR_MESSAGE, Errors
+          .KAFKA_ERROR_ERROR_CODE, e);
+    }
+    return allTopics;
+  }
+
+  public boolean topicExists(String topic) {
+    try {
+      Collection<String> allTopics =
```
mhowlett: could call `getTopicNames` and avoid writing a bit of code here.

src/main/java/io/confluent/kafkarest/AdminClientWrapper.java
```diff
@@ -0,0 +1,197 @@
...
+      throw new RestServerErrorException(Errors.KAFKA_ERROR_MESSAGE, Errors
+          .KAFKA_ERROR_ERROR_CODE, e);
+    }
+    return topic;
+  }
+
+  public List<Partition> getTopicPartitions(String topicName) {
+    List<Partition> partitions = null;
+    try {
+      TopicDescription topicDescription = adminClient.describeTopics(
```
mhowlett: describeTopics and associated exception handling can probably be factored out for terser code.

config/kafka-rest.properties
mageshn: This used to be a default value but we want users to explicitly specify. Specifying this in the config  will make the quick start easy.

src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
```diff
@@ -20,15 +20,17 @@
 
 /**
  * Shared, global state for the REST proxy server, including configuration and connection pools.
+ * The objects for producer, admin and new Consumer would be lazy initialized.
```
mhowlett: maybe 'ProducerPool, AdminClientWrapper and KafkaConsumerManager instances are initialized lazily'.

src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
```diff
@@ -37,47 +39,113 @@ public DefaultKafkaRestContext(
       ConsumerManager consumerManager,
       SimpleConsumerManager simpleConsumerManager
   ) {
-    this(config, metadataObserver, producerPool, consumerManager, simpleConsumerManager, null);
+    this(config, metadataObserver, producerPool, consumerManager, simpleConsumerManager, null,
```
mhowlett: The first two constructors are only used by tests - it would be better if they were removed IMO.

src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
mhowlett: or 'ProducerPool, AdminClientWrapper and KafkaConsumerManager instances are initialized lazily if required' (as they are provided as constructor arguments)

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -67,8 +85,9 @@
       + "use the same "
       + "chroot path in its connection string. For example to give a chroot path of /chroot/path "
       + "you would give "
-      + "the connection string as hostname1:port1,hostname2:port2,hostname3:port3/chroot/path.";
-  public static final String ZOOKEEPER_CONNECT_DEFAULT = "localhost:2181";
+      + "the connection string as hostname1:port1,hostname2:port2,hostname3:port3/chroot/path. "
+      + "This is required to be configured only to use v1 Consumer API's.";
```
mhowlett: Maybe 'NOTE: Only required when using v1 Consumer API's.' right at the beginning of the connection string. I think getting this message across is actually more important than what the property does at this point.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -83,10 +102,10 @@
       + "initial connection to discover the full cluster membership (which may change "
       + "dynamically), "
       + "this list need not contain the full set of servers (you may want more than one, though, "
-      + "in case a server is down).";
+      + "in case a server is down). While it is required to configure either of this or "
```
mhowlett: Is this extra text necessary? bootstrap.servers is always required, right? I think this is covered off by the comment in the zookeeper connect doc string.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -83,10 +102,10 @@
...
+      + "``zookeeper.connect``; it is recommended to always use ``bootstrap.servers``.";
 
-  public static final String BOOTSTRAP_SERVERS_DEFAULT = "PLAINTEXT://localhost:9092";
-  public static final String KAFKACLIENT_CONNECTION_URL_DEFAULT = "localhost:2181";
+  public static final String BOOTSTRAP_SERVERS_DEFAULT = "";
```
mhowlett: there is a blank line separating BOOTSTRAP_SERVERS_DEFAULT from the other BOOTSTRA_SERVERS constants, but no new line separating it from SCHEMA_REGISTRY_URL constants. reverse this ...

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
mhowlett: oh, bootstrap servers can be found via zookeeper as well. hmm. but since we're trying to remove the zookeeper dependency, i don't understand why we want to support this? It appears you've added this support in this PR too, so it seems it's not for backwards compatibility (haven't looked into that thoroughly, could be mistaken).

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -712,6 +697,56 @@ public Properties getConsumerProperties() {
...
 
+  public Properties getAdminProperties() {
+    Properties consumerProps = new Properties();
+    //copy cover the properties with prefixes "client." and  "admin."
+    addPropertiesWithPrefix("client.", consumerProps);
+    addPropertiesWithPrefix("admin.", consumerProps);
+    return consumerProps;
+  }
+
+  public String bootstrapBrokers() {
```
mhowlett: see previous note questioning whether this is necessary.

src/main/java/io/confluent/kafkarest/ProducerPool.java
```diff
@@ -52,16 +52,15 @@
...
+  public ProducerPool(KafkaRestConfig appConfig) {
+    this(appConfig, null);
   }
 
   public ProducerPool(
       KafkaRestConfig appConfig,
-      ZkUtils zkUtils,
       Properties producerConfigOverrides
   ) {
-    this(appConfig, getBootstrapBrokers(zkUtils), producerConfigOverrides);
```
mhowlett: you can remove the `getBootstrapBrokers` method from this class as it's now in the config class.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
mhowlett: ok, so now I see this is for backwards compatibility and was previously in `ProducerPool` (which was not a good place for it). It's a little bit questionable putting this in `KafkaRestConfig` too, but better than `ProducerPool`.

src/main/java/io/confluent/kafkarest/extension/KafkaRestContextProvider.java
```diff
@@ -49,22 +52,26 @@ public static void initialize(
...
       }
-      if (mdObserver == null) {
-        mdObserver = new MetadataObserver(appConfig, zkUtils);
-      }
-      if (producerPool == null) {
-        producerPool = new ProducerPool(appConfig, zkUtils);
+
+      if (zkUtils == null) {
+        mdObserver = new UnsupportedMetaDataObserver(zkUtils);
+      } else if (mdObserver == null && zkUtils != null) {
```
mhowlett: check for zkUtils != null not necessary here, always true.

config/kafka-rest.properties
mhowlett: very good. I completely agree this definitely shouldn't be a in-code default.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
mhowlett: I would still omit the text you've added here. If you include something, I'd reword it a bit and probably use the word 'depreciated' rather than 'recommended'.

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
mhowlett: not suggesting this be moved, i think it's a good enough factoring all things considered.

mhowlett:
LGTM after comments addressed as you see fit.

src/main/java/io/confluent/kafkarest/AdminClientWrapper.java
mageshn: will try to make it consistent

src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
mageshn: I still didn't want to use depreciated because we would need to depreciated V1 API's first

PR #336, opened 2017-08-15
CLIENTS-555 - Fix Admin api for secure tests

mageshn:


mageshn:
This is a trivial fix and hence going ahead and merging

PR #345, opened 2017-09-15
allowing pass through of extra consumer configs

### LINKED ISSUE #344: Not all properties get passed to the v2 consumer (which breaks Confluent interceptors)
The confluent interceptors can be controlled by using the `confluent.monitoring.interceptor.*` configs. For example, a commonly used one is: `confluent.monitoring.interceptor.security.protocol=SASL_SSL` (for secure mode).

As it stands today, the REST Proxy won't allow these configs to be passed to the consumer, because of this missing line: https://github.com/confluentinc/kafka-rest/blob/3.3.0-post/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java#L699 
But it is included for the producer so the interceptor would be correctly configured.

I'll submit a PR to include the fix. Can we please get this merged and released in 3.3.1?

simplesteph:
@ewencp can you please review and is it possible to include this in 3.3.1 ?
this fixes #344

### DIFF
--- a/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -690,6 +690,9 @@ public class KafkaRestConfig extends RestConfig {
 
   public Properties getConsumerProperties() {
     Properties consumerProps = new Properties();
+    // add consumer configuration parameters to pass through configs
+    // that may be used by interceptors, or other classes
+    addExistingV1Properties(consumerProps);
     //copy cover the properties with prefixes "client." and  "consumer."
     addPropertiesWithPrefix("client.", consumerProps);
     addPropertiesWithPrefix("consumer.", consumerProps);


mageshn:
ready to test

mageshn:
@simplesteph Thanks for the PR. addExistingV1Properties exists today to support v1 API's and when we deprecate V1 this might possibly go away as well. Long run, its best to express all Rest Proxy client properties with prefix 'client', 'producer' or consumer'.  It should pass through any config that starts with 'client.'. So for example, if you want to pass through a monitoring interceptor config, you could just add the config as 

client.confluent.monitoring.interceptor.security.protocol=SASL_SSL

simplesteph:
Hi 

It  needs to add extra logic. I think it's more of a confluent logic decision than a bug fix for 3.3.1. You're definitely looking for a long-term fix, but I wouldn't be the one implementing it. 

In the meantime, I'd like to get this merged in. The change you're mentioning would be a breaking change for many, which is a 4.0 concern. Let me know your thoughts

mageshn:
Just to add a little more, without any code changes, we should be able to add something like below
client.confluent.monitoring.interceptor.security.protocol=SASL_SSL

and have confluent.monitoring.interceptor.security.protocol=SASL_SSL being available to both producer and consumer

simplesteph:
That's correct it should work. Let me test that on Monday and I'll get back to you. Thanks !

simplesteph:
@mageshn all working thanks!
`client.confluent.monitoring.interceptor.security.protocol=SASL_SSL`

PR #363, opened 2017-11-16
Add unique name to Kafka REST operations

joel-hamill:


joel-hamill:
@ewencp where should we start such a discussion? Marketing team? A PM?

PR #365, opened 2017-12-01
Make kafka rest multi module

mageshn:


aayars:
Can you retarget this against 4.0.x?

mageshn:
@aayars why 4.0.x?

aayars:
That's where 4.0 dev is happening. Version should also be 4.0.1-SNAPSHOT at this point.

PR #366, opened 2017-12-01
Make kafka rest multi module

mageshn:


mageshn:
ready to test

aayars:
Just making sure you want to target master here. For something like this which is changing the structure of a project, I think we'd want to target the oldest viable branch (perhaps even 3.2.x or 3.3.x) and merge up.

By targeted `master` only, we'll only be able to leverage the multi-module-ness of this project in versions 4.1.0 and newer.

mageshn:
I don't see the necessity of a multi module project in older versions. 
@ewencp  what are your thoughts on this?

pom.xml
```diff
@@ -52,169 +58,42 @@
...
+                    <artifactId>exec-maven-plugin</artifactId>
+                    <version>${exec-maven-plugin.version}</version>
+                </plugin>
+            </plugins>
+        </pluginManagement>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <inherited>true</inherited>
```
ewencp: do we need to specify this plugin at all? don't we inherit all of this from the parent pom?

mageshn:
@ewencp  trying to keep it in sync with other OSS projects like Schema Registry and KSQL. Having said that, we can also include another assembly and get it done. 
@aayars  your final thoughts and we can take this towards closure :)

aayars:
The multi-module approach was my suggestion, for consistency and to avoid mucking with lower-level plugins. If we use "just an assembly", I believe we would also need the deploy plugin to deploy-file on that custom assembly, and the project just gets more and more different than our baseline platform project. FWIW, this submodule approach was used in KSQL and Schema Registry.

ewencp:
I don't think your latter point about deploy is true, this is also "just an assembly" since the project itself is being marked as pom, not zip. In any case, I'm fine w/ multimodule and its actually kind of surprising kafka-rest never had need to go multimodule. But until there's something in this repo explaining why you would have this separate module, the bare minimum is documenting why it even exists.

xvrl:
@aayars There shouldn't be any need to specify specific assemblies as part of the deploy phase. Any assembly that is created by the assembly:single goal will automatically be attached to the project and deployed as part of the deploy phase, using the assemblyId as the classifier for that artifact.
If you look at projects that deploy test jar and source packages, you can see they are already deployed as part of the same project without the need to create a separate module. Test classes and source jars are also simply attached to the project at build time and deployed as part of the same deploy phase.

aayars:
@ewencp Would adding a README to the new submodule address your documentation concern?

mageshn:
@aayars May be creating an assembly with appropriate classifier is cleaner for all the projects?

ewencp:
@aayars yeah, just want some documentation, in any form. could also just be xml comments in the pom file.

PR #370, opened 2017-12-08
Debian systemd

edenhill:
One of four PRs (across master, rpm, archive, debian branches) to add systemd init scripts.

All four PRs need to be merged at the same time since packaging files are moved from deb,archive,rpm branches to master.


debian/confluent-kafka-rest.postinst
```diff
@@ -0,0 +1,46 @@
...
+
+set -e
+
+case "$1" in
+    configure)
+	_user=cp-kafka-rest
+	_group=confluent
+
+	(getent group $_group 2>&1 >/dev/null) || addgroup --quiet --system $_group
+        adduser --system --disabled-password --disabled-login --home /var/empty \
```
ewencp: very minor nit, but i think the spacing is off on this line

PR #371, opened 2017-12-08
Systemd

edenhill:
One of four PRs (across master, rpm, archive, debian branches) to add systemd init scripts.

All four PRs need to be merged at the same time since packaging files are moved from deb,archive,rpm branches to master.


edenhill:
Urgh, I just saw your comment after merging a bunch of these PRs.
Maybe it is better to backport to create_archive.sh move, but remove the init script installation.

This is a problem with the rpm/deb/archive branche approach as they are not versioned.
Probably better to do something like kafka-packaging does and have one single "packaging-<BRANCH>" branch.

ewencp:
yes, that has been the plan for quite awhile. nobody has found the time to collapse the other branches like i did with kafka.

edenhill:
Okay, cool. I'll restructure the branching to following the packaging-NNN format and revert the changes to rpm,debian,archive to unbreak previous release builds.

ewencp:
@edenhill i was just thinking about this and if we wanted to, we could just change the packaging to be included in the normal source tree and update the packaging scripts. we can't do that for kafka, but we can follow that pattern for all our other repos. just collapse all the branches now and keep things simpler, get the versioning for free during our normal branching, not worry about tagging packaging  branches during release as well, etc.

PR #382, opened 2018-01-10
DOCS-204: add REST proxy content from https://docs.confluent.io/curre…

ybyzek:
…nt/kafka/platform-security.html


ybyzek:
@ewencp your suggestion is great and I generally advocate for diagrams.  I filed a separate Jira to track

PR #383, opened 2018-01-11
Increase timeout for a few tests to make them less flaky on Jenkins

ewencp:


ewencp:
hah, 10s felt long, but i'd rather be conservative given how many flaky tests we have to deal with. better to just get rid of the issue even if it "costs" a bit more.

edenhill:
Right, but this is the worst case timeout, not the typical test runtime, the only downside is that slow-down regression is not caught as easily, but a test-timeout is a bit of a crude tool for that anyway.

PR #384, opened 2018-01-14
Add file sink, similar to schema-registry configs

synhershko:
To make all Confluent Platform parts play the same tune and use the same logging conventions, the file sink is missing and needs to be added to the default rest proxy logging configs. (See https://github.com/confluentinc/schema-registry/blob/master/config/log4j.properties#L1 for the schema registry configs)


synhershko:
Tho this seems to also be included in https://github.com/confluentinc/kafka-rest/pull/371

synhershko:
[clabot:check]

Issue #394, opened 2018-02-05
Schema registry lookup failed returns HTTP 408 Request Timeout


koreiko:
When posting data to Rest Proxy with an invalid `value_schema_id` (one that simply doesn't exist in the schema registry) HTTP 408 Request Timeout response code is received. This sounds like the wrong response code to return.

```
curl -X POST -H "Content-Type: application/vnd.kafka.avro.v2+json" -H "Accept: application/vnd.kafka.v2+json" --data '{"records": [{"value": { ... }}], "value_schema_id": 12341}' "http://localhost:8082/topics/esb.logrecord" -vvv
* About to connect() to localhost port 8082 (#0)
*   Trying ::1...
* Connected to localhost (::1) port 8082 (#0)
> POST /topics/esb.logrecord HTTP/1.1
> User-Agent: curl/7.29.0
> Host: localhost:8082
> Content-Type: application/vnd.kafka.avro.v2+json
> Accept: application/vnd.kafka.v2+json
> Content-Length: 649
>
* upload completely sent off: 649 out of 649 bytes
< HTTP/1.1 408 Request Timeout
< Date: Mon, 05 Feb 2018 09:16:22 GMT
< Content-Type: application/vnd.kafka.v2+json
< Content-Length: 69
< Server: Jetty(9.2.22.v20170606)
<
* Connection #0 to host localhost left intact
{"error_code":40801,"message":"Schema registration or lookup failed"}
```

koreiko:
Looking more into this, there is no logging in WARN / ERROR level in the logs.

Also, and much more importantly, **it seems like the message does go through into Kafka despite this error code and the schema not being found.**

addisonhuddy:
Hi @koreiko, thanks for the feedback.  Today, the brokers will not reject a message if it the message does not have a valid schema.  It will simply send a message back to the application warning that a valid schema was not found but still write the message.  I'm interested in learning what would you expect the behavior to be?

blootsvoets:
Scenario: The schema ID does not exist.
Cause: This could happen on corruption of the schema registry, or the client caching stale schema IDs.
Expected response: a message with an unknown schema ID means that the message was semantically invalid, and a 422 HTTP response code seems appropriate. This is the same code as when a schema is present but the content does not comply to the schema.
Expected behaviour: The client should get the opportunity to fix the request. The server likely cannot process it, perhaps because the client still had some stale schema IDs.

[...]

PR #399, opened 2018-02-12
Don't return HTTP 408 on schema errors

synhershko:
Partially fixes https://github.com/confluentinc/kafka-rest/issues/394 (at least avoids endless retries by properly configured clients)


AndrewJSchofield:
Tidying up dormant PRs. These errors are not handled in this way nowadays.

PR #408, opened 2018-03-02
Clean up CLASSPATH assignment

wrp:
Remove unnecessary '../' from CLASSPATH entries, and ensure that a `:` is only inserted if the CLASSPATH is not already empty.


wrp:
[clabot:check]

PR #412, opened 2018-03-08
DOCS-339 - Fix config hyperlink

joel-hamill:
Fix wrong link in kafka rest doc

https://confluentinc.atlassian.net/browse/DOCS-339

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -271,7 +271,7 @@ running:
 
 If you installed Debian or RPM packages, you can simply run ``kafka-rest-start``
 as it will be on your ``PATH``. The ``kafka-rest.properties`` file contains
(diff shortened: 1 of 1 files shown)


docs/intro.rst
```diff
@@ -271,7 +271,7 @@ running:
 
 If you installed Debian or RPM packages, you can simply run ``kafka-rest-start``
```
edenhill: unrelated to this change: we should probably mention the systemd init scripts here.

joel-hamill: @edenhill do you think this needs mentioning in each individual component where this is available, e.g. Kafka, Schema-Registry, C3, and Kafka Connect? Related, see https://github.com/confluentinc/docs/pull/764/files.

PR #414, opened 2018-03-12
Remove choosePort and rely on dynamic, OS-allocated ports during tests

ewencp:


ewencp:
Works in conjunction with https://github.com/confluentinc/rest-utils/pull/90 and avoids `Address already in use` errors during tests due to `choosePort` not being a reliable way to select a random available port.

PR #419, opened 2018-03-29
Fix configuration parameter for schema registry

ewencp:


ewencp:
ping @confluentinc/clients , this and #420 should be trivial to review

C0urante:
Doesn't this occur at line 1582 as well? Would comment on that in the diff but can't figure out how to scroll down that far...

PR #420, opened 2018-03-29
Fix configuration parameter for schema registry

ewencp:


ewencp:
This is basically the same as #419 which got the necessary LGTM

PR #421, opened 2018-04-01
Add configurable x-header reflection capability

joncourt:
added a global filter to reflect any X headers (headers starting with X- or x-) in the response initially to simplify correlating requests with responses. Default is disabled.


joncourt:
closed in favour of a new pull request against 5.0.0-post.

PR #438, opened 2018-06-10
Don't use deprecated RecordMetadata constructor

ijuma:
This constructor has been removed in Apache Kafka 2.0.


ijuma:
Since this is a trivial change and it fixes the build, I'm going to go ahead and merge it. If there are any comments, I will address them in a follow-up.

PR #439, opened 2018-06-19
Temporarily pin Kafka and Confluent versions

ijuma:
The Scala consumers have been removed from Apache Kafka and
this change gets the build working again. I will submit
another pull request that makes things work with Kafka 2.0.


dhruvilshah3:
Thanks, LGTM!

PR #447, opened 2018-07-11
Adding RAW type support (support for StringSerializer records)

Tribwerk:
This is an enhanced version of the Kafka REST Proxy with an additional RAW type which enables for consumption and production of Kafka records with key and/or value of type string.

The new RAW type can be used from within the same REST api that the original REST Proxy offers.

### Creation of RAW Consumer

curl -X POST \
  http://rest-proxy.dev.aareality.aareal.org/consumers/myconsumergroup \
  -H 'Content-Type: application/vnd.kafka.v2+json' \
  -d '{"name": "myconsumerinstance", "format": "raw", "auto.offset.reset": "smallest"}'

### Reading records from RAW Consumer

curl -X GET \
  http://rest-proxy.dev.aareality.aareal.org/consumers/myconsumergroup/instances/myconsumerinstance/records \
  -H 'Accept: application/vnd.kafka.raw.v2+json'

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -39,4 +39,24 @@
             checks="(MethodLength)"
             files="(KafkaRestConfig).java"
     />
+    <suppress
+            checks="(ClassDataAbstractionCoupling)"
+            files="(ProducerPool).java"
+    />
+    <suppress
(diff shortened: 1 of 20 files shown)


Tribwerk:
[clabot:check]

PR #449, opened 2018-07-18
Fix API v1 /consumers documentation for HTTP Content-Type headers

hanmd82:
Current documentation is inaccurate - 'Accept' HTTP header throws the following error:
```
{"error_code":415,"message":"HTTP 415 Unsupported Media Type"}
```

This documentation fix is important for users of the v1 REST API.


hanmd82:
[clabot:check]

PR #454, opened 2018-08-08
Add x-header-reflection and host.port config setting

joncourt:
added a global filter to reflect any X headers (headers starting with X- or x-) in the response initially to simplify correlating requests with responses. Default is disabled.


joncourt:
closing in favour of #585

PR #466, opened 2018-09-11
Change Consumer threading model

stanislavkozlovski:
Previously, we would have a fixed number of Worker threads which ran Consumer tasks. This proved problematic due to consumer rebalances requiring consumers to poll. It could easily result in a deadlock when the number of consumers in a group were bigger than the allocated threads.

This commit:
* Changes Executor service to be with a cached thread pool (dynamically expanding)
* Changes consumer read tasks to serially run on separate threads on said executor service. Meaning, we have one thread per running task.
* Removes `consumer.threads` config
* Adds `consumer.max.threads` config - the maximum amount of threads the executor service can spin up. Worth thinking whether the default value of 100 may be little in some cases
* Change `max.poll.records` to 10. Reasoning is that we want to ensure we call `poll()` on most requests


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -60,12 +60,24 @@
       + "are used.";
   public static final String ID_DEFAULT = "";
 
+  public static final String MAX_POLL_RECORDS_CONFIG = "max.poll.records";
```
hachikuji: I wonder if we need to expose this or if the default is reasonable for most cases. I suspect we'd ultimately get rid of it if we move to a lower-level consumer fetch API.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -60,12 +60,24 @@
...
   private static final String HOST_NAME_DOC =
       "The host name used to generate absolute URLs in responses. If empty, the default canonical"
       + " hostname is used";
   public static final String HOST_NAME_DEFAULT = "";
 
+  public static final String CONSUMER_MAX_THREADS_CONFIG = "consumer.threads";
+  private static final String CONSUMER_MAX_THREADS_DOC =
+      "The maximum number of threads to run consumer requests on."
+      + " The value of -1 denotes unbounded thread creation";
+  public static final String CONSUMER_MAX_THREADS_DEFAULT = "1";
```
hachikuji: Is this right? We document 100 above.

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -82,14 +81,12 @@ public ConsumerManager(KafkaRestConfig config, MetadataObserver mdObserver) {
...
-      ConsumerWorker worker = new ConsumerWorker(config);
-      workers.add(worker);
-      worker.start();
-    }
-    nextWorker = new AtomicInteger(0);
-    this.executor = Executors.newFixedThreadPool(1);
+    // Cached thread pool
+    int maxThreadCount = config.getInt(CONSUMER_MAX_THREADS_CONFIG) < 0 ? Integer.MAX_VALUE
+            : config.getInt(CONSUMER_MAX_THREADS_CONFIG);
+    this.executor = new ThreadPoolExecutor(0, maxThreadCount,
```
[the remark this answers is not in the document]
stanislavkozlovski: Yeah. I'll open an issue to keep track of that once we merge this PR

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerWorker.java
```diff
@@ -19,38 +19,23 @@
...
-
 import io.confluent.kafkarest.ConsumerWorkerReadCallback;
 import io.confluent.kafkarest.KafkaRestConfig;
 
 /**
- * Worker thread for consumers that multiplexes multiple consumer operations onto a single thread.
+ * Synchronous worker thread for a consumer
  */
-public class KafkaConsumerWorker extends Thread {
+public class KafkaConsumerWorker {
```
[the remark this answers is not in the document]
stanislavkozlovski: Agreed

docs/config.rst
```diff
@@ -60,8 +60,9 @@ General
 
 ``consumer.threads``
   The maximum number of threads to run consumer requests on. Note that this must be greater than the maximum number of consumers in a single consumer group.
+  The sentinel value of -1 denotes an unbounded number of threads.
```
hachikuji: How about this?

> The sentinel value of -1 allows the number of threads to grow as needed to fulfill consumer requests. Inactive threads will ultimately be stopped and cleaned up.

stanislavkozlovski: Way better, yeah

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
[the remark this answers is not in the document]
hachikuji: Sounds good.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
stanislavkozlovski: Sounds fair considering it's a cached pool

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
hachikuji: maybe 30?

stanislavkozlovski: Sounds good, yep

docs/config.rst
```diff
@@ -58,18 +58,18 @@ General
...
 
-``consumer.request.timeout.ms``
-  The maximum total time to wait for messages for a request if the maximum number of messages has not yet been reached.
-
+``consumer.threads``
+  The maximum number of threads to run consumer requests on. Note that this must be greater than the maximum number of consumers in a single consumer group.
+  The sentinel value of -1 allows the number of threads to grow as needed to fulfill active consumer requests. Inactive threads will ultimately be stopped and cleaned up.
   * Type: int
-  * Default: 1000
+  * Default: 1
```
hachikuji: Should be 16

stanislavkozlovski:
Thanks for the review @hachikuji. I'll make sure to run all associated system tests tomorrow before I merge

stanislavkozlovski:
Tests passed

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
ijuma: Isn't 16 a bit low? Given that this is a cached thread pool, should we be considering a larger value by default? Since we decrease the thread pool size if there are idle threads, it seems like the risk is low. As @stanislavkozlovski said, the main downside is there were a lot of 1 group consumers, but not sure how common that is.

ijuma: Also, maybe the core number should be higher than 0 so that we don't end up with no threads at all during idle periods.

PR #470, opened 2018-09-13
CPKAFKA-1681: Add long-polling functionality

stanislavkozlovski:
This PR enables long-polling in V1 and V2 APIs of the REST proxy.

Previously, consumer.request.timeout.ms would control when the proxy sent back a response to the user. The default value was 1000ms, but setting it explicitly to 1000 (or 2000) would cause it to be passed to the underlying consumer connection. This caused a NullPointerException, most likely because configuration exceptions were ignored in the V2 API. This PR removes the consumer.request.timeout.ms functionality of controlling when the proxy returns a response.

Adds new config settings:
* proxy.fetch.max.wait.ms - time before a consumer response is sent back when neither min or max bytes are fulfilled
* proxy.fetch.min.bytes - amount of bytes of records the proxy will collect before immediately returning a response. In the V2 API, the proxy will exhaust all already-loaded records when it hits that threshold.
* deprecate consumer.request.max.bytes and introduce proxy.fetch.max.bytes - same config but renamed. If proxy.fetch.max.bytes is explicitly defined it will override the set consumer.request.max.bytes
* proxy.fetch.max.wait.ms and proxy.fetch.min.bytes are configurable per consumer as well as globally
These new settings control how the user->proxy connection, not the underlying consumer->broker.

Now also raises exceptions and returns proper 4xx response when V2 Consumer settings are invalid.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
@@ -19,6 +19,7 @@ package io.confluent.kafkarest.v2;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.config.ConfigException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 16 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -199,34 +200,18 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
           );
       }
 
-      Consumer consumer = null;
+      Consumer consumer;
       try {
         if (consumerFactory == null) {
           consumer = new KafkaConsumer(props);
```
hachikuji: Can we add the logic to strip `consume.request.timeout.ms` from these properties? It would be better to hard-code a reasonable default than to allow the current misuse.

hachikuji:
@stanislavkozlovski Sorry, accidentally submitted a partial review. Haven't quite made it through all the code yet.

stanislavkozlovski:
@hachikuji I've addressed the small comments. Let's move this PR discussion to https://github.com/confluentinc/kafka-rest/pull/481
This was branched off trunk which shouldn't have been the case

PR #472, opened 2018-09-13
Consumer threading - Backport to 4.1.x

stanislavkozlovski:


kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerState.java
```diff
@@ -21,8 +21,7 @@
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.Vector;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
```
hachikuji: For the old consumer, the read/write lock is fine. The underlying `ConsumerIterator` class was thread-safe.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -112,11 +106,11 @@ public ConsumerInstanceId getId() {
 
 
   public void startRead() {
-    lock.readLock().lock();
+    lock.lock();
```
hachikuji: My main concern is that we hold onto this lock even when the task is not scheduled. This is probably why we tried to use the read lock before even though it didn't work. In the worst case, this can cause deadlock if a task which depends on this lock is scheduled ahead of the one that is holding it. I'm wondering if it would be better to hold the lock only during the `partialRead` operation. The main drawback of doing so as far as I can tell is that certain "write" operations may interfere with a read which is actively polling for data. For example, if there is a concurrent `seek`, then the fetch may include data both from before and after the `seek`. 

One way to solve this would be to have a queue of pending operations for each consumer instance and only submit them to the executor one at a time. I think this can be left for future work though. It's more important to address the problems with the current threading.

stanislavkozlovski: I see what you mean.
I am not very knowledgeable on the thread scheduling internals. In the deadlock case, wouldn't a task that is stuck on a lock eventually be context-switched out of? A deadlock happens when both threads want something the other one is holding whereas in this case, only one thread wants something the other is holding.  The original thread which has the lock would be free to proceed?

In regards to a concurrent `seek`, that behavior seems to be described in the javadoc of the method in Kafka itself:
```java
    /**
     * Overrides the fetch offsets that the consumer will use on the next {@link #poll(long) poll(timeout)}. If this API
     * is invoked for the same partition more than once, the latest offset will be used on the next poll(). **Note that
     * you may lose data if this API is arbitrarily used in the middle of consumption, to reset the fetch offsets**
...
    public void seek(TopicPartition partition, long offset) {
```
so maybe it is expected behavior?

My concern is what a concurrent fetch would do. I imagine there would be a possibility where we return non-consecutive offsets from both requests, e.g
```
request A
messages with offset 10-20. 40-60

request B
messages with offset 0-10. 20-40
```
because that is what they both got respectively from the `poll()` calls.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -175,6 +199,8 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
       if (instanceConfig.getAutoOffsetReset() != null) {
         props.setProperty("auto.offset.reset", instanceConfig.getAutoOffsetReset());
       }
+      // override request.timeout.ms to the default
```
hachikuji: Probably worth adding some detail to the comment about this.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -127,7 +121,7 @@ public void finishRead() {
       String async,
       ConsumerOffsetCommitRequest offsetCommitRequest
   ) {
-    lock.writeLock().lock();
+    lock.acquireUninterruptibly();
```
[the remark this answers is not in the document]
stanislavkozlovski: I understand now. I think I'll go with simply not holding the lock while reading

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -503,6 +567,95 @@ private synchronized void updateExpiration(KafkaConsumerState state) {
...
+            }
+          } else {
+            sleepTimeMs = 25;
+          }
+          Thread.sleep(sleepTimeMs);
+        }
+      } catch (InterruptedException e) {
+        // Interrupted by other thread, do nothing to allow this thread to exit
+      }
+      shutdownLatch.countDown();
```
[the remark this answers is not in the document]
stanislavkozlovski: nice catch

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -82,14 +81,12 @@ public ConsumerManager(KafkaRestConfig config, MetadataObserver mdObserver) {
...
-      ConsumerWorker worker = new ConsumerWorker(config);
-      workers.add(worker);
-      worker.start();
-    }
-    nextWorker = new AtomicInteger(0);
-    this.executor = Executors.newFixedThreadPool(1);
+    // Cached thread pool
+    int maxThreadCount = config.getInt(CONSUMER_MAX_THREADS_CONFIG) < 0 ? Integer.MAX_VALUE
+            : config.getInt(CONSUMER_MAX_THREADS_CONFIG);
+    this.executor = new ThreadPoolExecutor(0, maxThreadCount,
```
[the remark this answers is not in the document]
stanislavkozlovski: What would the thread count for all the ConsumerWorkers be then?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -503,6 +600,74 @@ private synchronized void updateExpiration(KafkaConsumerState state) {
...
+        long timeoutMs = 25;
+        while (isRunning.get()) {
+          RunnableReadTask readTask = delayedReadTasks.poll(timeoutMs, TimeUnit.MILLISECONDS);
+          if (readTask != null) {
+            delayedTasks.put(readTask, false);
+            executor.submit(readTask);
+            continue;
+          }
+
+          readTask = delayedReadTasks.peek();
```
hachikuji: I'm not sure I understand why we need this logic. The call to `DelayedQueue.poll(timeout)` will return as soon as the next delay has expired. I think it would be reasonable to poll with a fixed timeout like 500ms just so we're fairly responsive to shutdown. We could make the timeout even longer if we have a nice way to wakeup the delay queue (such as the poison pill I mentioned offline)

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -503,6 +600,74 @@ private synchronized void updateExpiration(KafkaConsumerState state) {
...
+      }
+    }
+
+    public void shutdown() {
+      try {
+        isRunning.set(false);
+        this.interrupt();
+        shutdownLatch.await();
+      } catch (InterruptedException e) {
+        throw new Error("Interrupted when shutting down read task scheduler thread.");
```
hachikuji: Not sure on the use of `Error` here. Could we use a `RuntimeException`?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -78,34 +87,48 @@
...
+            new RejectedExecutionHandler() {
+            @Override
+            public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
+              RunnableReadTask readTask = (RunnableReadTask) r;
+              if (delayedTasks.get(readTask) == null
+                      || delayedTasks.get(readTask)) {
+                return;
+              }
+              delayedTasks.put(readTask, true);
+              delayedReadTasks.add(readTask);
```
hachikuji: Did we not want to add a backoff to the wait expiration?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -78,34 +87,48 @@
...
-  // A few other operations, like commit offsets and closing a consumer can't be interleaved, but
-  // they're also comparatively rare. These are executed serially in a dedicated thread.
+  // All kind of operations, like reading records, committing offsets and closing a consumer
+  // are executed separately in dedicated threads via a cached thread pool.
   private final ExecutorService executor;
   private KafkaConsumerFactory consumerFactory;
   private final PriorityQueue<KafkaConsumerState> consumersByExpiration =
       new PriorityQueue<KafkaConsumerState>();
+  private final DelayQueue<RunnableReadTask> delayedReadTasks = new DelayQueue<>();
+  private final ConcurrentHashMap<RunnableReadTask, Boolean> delayedTasks =
```
hachikuji: I'm not sure I understand why we need this. The state machine as I understand is the following:

1. Task creation
2. Submission to executor [rejection -> delay queue]
3. Either:
a. Execution [incomplete read -> delay queue]
b Delay queue poll and resubmission
4. Task completion

Is there anywhere that we might submit the task twice to the queue?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +303,107 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+          ClientKeyT, ClientValueT>(
+          state,
+          timeout,
+          maxBytes,
+          new ConsumerWorkerReadCallback<ClientKeyT, ClientValueT>() {
+            @Override
+            public void onCompletion(
+                    List<? extends ConsumerRecord<ClientKeyT, ClientValueT>> records,
+                    Exception e) {
+              KafkaConsumerManager.this.updateExpiration(state);
```
hachikuji: It's a little curious that we only update the consumer expiration when the read task completes. I'd have though we'd do it every time the task executes at least.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +303,107 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+        Exception responseException = e;
+        if (!(e instanceof RestException)) {
+          responseException = Errors.kafkaErrorException(e);
+        }
+        taskState.callback.onCompletion(null, responseException);
+      }
+    }
+
+    @Override
+    public long getDelay(TimeUnit unit) {
```
hachikuji: Should we be ignoring the time unit? This might be used internally in `DelayQueue`, so we had better not I think.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +303,107 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+      this.requestExpiration = this.started
+              + consumerConfig.getInt(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG);
+      this.waitExpiration = 0;
+    }
+
+    @Override
+    public void run() {
+      try {
+        log.trace("Executing consumer read task ({})", taskState.task);
+        long backoffTime = config.getTime().milliseconds()
```
hachikuji: Hmm.. we compute this prior to the call to `doPartialRead` which can block. The value will be stale when the read returns. I think it would be better to move this into the task not done case following the partial read.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -503,6 +600,74 @@ private synchronized void updateExpiration(KafkaConsumerState state) {
     Consumer createConsumer(Properties props);
   }
 
+  private class ReadTaskState {
```
hachikuji: nit: could be static? Should probably check all these inner classes to see if they can be static. The advantage is they avoid the reference to the parent class.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +303,107 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+
+    @Override
+    public long getDelay(TimeUnit unit) {
+      return waitExpiration - config.getTime().milliseconds();
+    }
+
+    @Override
+    public int compareTo(Delayed o) {
+      if (o == null) {
+        return 1;
```
hachikuji: Can you explain this? The Comparable interface specifies throwing an NPE.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +303,107 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+      }
+      if (taskState.task.isDone()) {
```
hachikuji: This seems like overkill. If we're concerned about accidental submission, maybe we could check for the task completion prior to running? If we happen to hit this, it would likely crash the scheduler thread. I would probably just remove it and rely on our logic to properly check for completion.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
hachikuji: I agree it is a concern, but I'm not sure the serialized read behavior gives you what you want either. If there are concurrent requests, there's a good chance one of them is a retry, so this behavior would ensure that the user loses data because the response to the first request will never be received. 

To be honest, I don't see how we can have good semantics if the read request doesn't take an expected fetch offset. I think it's more important at the moment to ensure that reads can always make progress even if the read semantics are not ideal.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: derp. right. forgot to remove this. Makes total sense to just have a larger timeout in the poll

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: wow that's a very good catch. I did not look into this.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: It uses `TimeUnit.NANOSECONDS`. From the docs:
```
.Expiration occurs when an element's getDelay(TimeUnit.NANOSECONDS) method returns a value less than or equal to zero.
```
Since we keep track in milliseconds and we keep it consistent (always return millis), I had thought it's not important.

If we would not ignore the parameter, it would make sense to track expirations in nanoseconds  or make our variables `double`, since we currently can't go more fine-grained than 1ms. I think there's no use - wdyt?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: ouch. good catch

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: My bad, I missed that from the docs

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: Yeah it was just sprinkled on top for sanity's sake

stanislavkozlovski:
@hachikuji I've moved the PR from a branch in my own fork. Can we continue the conversation there?
https://github.com/confluentinc/kafka-rest/pull/476

PR #476, opened 2018-09-20
Consumer Threading Improvements

stanislavkozlovski:
Changes:

* Use a `ThreadPoolExecutor` for running v2 consumer read tasks. It will spin up/down threads according to load, up to the max thread config.
* Do not propagate the `consumer.timeout.ms` to the AK consumer. Set the default 30 seconds explicitly
* Set max.poll.records to 30 for the v2 consumer. 
* Use single lock instead of read/write lock for the v2 consumer


kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -83,7 +83,7 @@ public ConsumerManager(KafkaRestConfig config, MetadataObserver mdObserver) {
     this.mdObserver = mdObserver;
     this.iteratorTimeoutMs = config.getInt(KafkaRestConfig.CONSUMER_ITERATOR_TIMEOUT_MS_CONFIG);
     this.workers = new Vector<ConsumerWorker>();
-    for (int i = 0; i < config.getInt(KafkaRestConfig.CONSUMER_THREADS_CONFIG); i++) {
+    for (int i = 0; i < config.getInt(KafkaRestConfig.CONSUMER_MAX_THREADS_CONFIG); i++) {
```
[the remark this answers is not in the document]
hachikuji: Yeah, that's a good point. We also have to figure out what the behavior should be if -1 is used for max threads. How risky do you think the change from before to use the thread pool executor instead of the dedicated workers is?  We can keep all the existing locking unchanged at least.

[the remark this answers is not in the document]
stanislavkozlovski: Shouldn't be any riskier than the v2 implementation. Do you see any potential complications?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
```diff
@@ -146,6 +125,8 @@ public boolean doPartialRead() {
             exceededMaxResponseBytes
         );
         finish();
+      } else {
+        parent.finishRead();
```
hachikuji: Don't we need this in a `finally`? Otherwise we are still holding the lock if there is an error or we finish successfully. I wonder in fact if we can push the locking into `hasNext()`?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -503,6 +601,57 @@ private synchronized void updateExpiration(KafkaConsumerState state) {
...
+                         ReadCallback callback) {
+
+      this.task = task;
+      this.consumerState = state;
+      this.callback = callback;
+    }
+  }
+
+  private class ReadTaskSchedulerThread extends Thread {
+    AtomicBoolean isRunning = new AtomicBoolean(true);
```
hachikuji: nit: these fields could be final

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +312,99 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+        if (!(e instanceof RestException)) {
+          responseException = Errors.kafkaErrorException(e);
+        }
+        taskState.callback.onCompletion(null, responseException);
+      }
+    }
+
+    @Override
+    public long getDelay(TimeUnit unit) {
+      return waitExpirationMs - config.getTime().milliseconds();
```
hachikuji: To make use of the unit, we could do something like this:
```java
long delayMs = waitExpirationMs - config.getTime().milliseconds();
return TimeUnit.MILLISECONDS.convert(delayMs, unit);
```

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -548,7 +697,7 @@ public void shutdown() {
         this.interrupt();
         shutdownLatch.await();
       } catch (InterruptedException e) {
-        throw new Error("Interrupted when shutting down consumer worker thread.");
+        throw new Error("Interrupted when shutting down expiration thread.");
```
hachikuji: nit: may as well be consistent and use `RuntimeException`? The type `Error` is usually reserved for abnormal situations such as out of memory.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -78,34 +94,50 @@
...
+    // Cached thread pool
+    int maxThreadCount = config.getInt(CONSUMER_MAX_THREADS_CONFIG) < 0 ? Integer.MAX_VALUE
+            : config.getInt(CONSUMER_MAX_THREADS_CONFIG);
+
+    this.executor = new ThreadPoolExecutor(0, maxThreadCount,
+            60L, TimeUnit.SECONDS,
+            new SynchronousQueue<Runnable>(),
+            new RejectedExecutionHandler() {
+            @Override
+            public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
```
hachikuji: Maybe useful to log a debug message if we hit this.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +312,99 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+              + consumerConfig.getInt(KafkaRestConfig.CONSUMER_ITERATOR_BACKOFF_MS_CONFIG);
+          waitExpirationMs = Math.min(backoffTime, requestExpiration);
+
+          // add to delayedReadTasks so the scheduler thread can re-schedule another partial read
+          delayedReadTasks.add(this);
+        } else {
+          log.trace("Finished executing consumer read task ({})", taskState.task);
+        }
+      } catch (Exception e) {
+        log.error("Failed to read records consumer "
```
hachikuji: nit: Failed to read record **from** consumer

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
stanislavkozlovski: No, because calls to `finish()` call finishRead. And `finish()` is called in all other code paths

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: I did not know it was that easy. For sure

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: Absolutely. Good catch

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: Very right

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
hachikuji: I see. Still, is there any reason not to move the locking into `hasNext`? That keeps the locking encapsulated. The `startRead` and `finishRead` approach only made sense when we had the read-write lock which we held for the duration of the read.

stanislavkozlovski: Sorry yeah, you're absolutely right

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -78,34 +94,52 @@
...
+            new SynchronousQueue<Runnable>(),
+            new RejectedExecutionHandler() {
+            @Override
+            public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
+              log.debug("The runnable {} was rejected execution."
+                  + "The thread pool must be satured or shutiing down", r);
+              if (r instanceof RunnableReadTask) {
+                RunnableReadTask readTask = (RunnableReadTask) r;
+                int delayMs = ThreadLocalRandom.current().nextInt(25, 75 + 1);
+                readTask.waitExpirationMs = config.getTime().milliseconds() + delayMs;
```
hachikuji: I am wondering if it makes sense to add a `RunnableReadTask.delay(long delayMs)`, which is used both here and when we are backing off to set `waitExpirationMs` and add to the delay queue. Then we could add the logic to check for expiration of the task in that method so that we have it down both paths.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -417,12 +401,18 @@ void getOrCreateConsumerRecords() {
   }
 
   public boolean hasNext() {
-    if (consumerRecordList != null && this.index < consumerRecordList.size()) {
+    lock.lock();
+    boolean hasRecords = consumerRecordList != null && this.index < consumerRecordList.size();
+    if (hasRecords) {
+      lock.unlock();
```
hachikuji: We unlock in all cases. Why not use a `finally`?

kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
```diff
@@ -15,6 +15,7 @@
  **/
 package io.confluent.kafkarest.unit;
 
+import io.confluent.kafkarest.*;
```
hachikuji: nit: we usually prefer explicit imports

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: That certainly is a better abstraction.
What task expiration are you referring to?

kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
stanislavkozlovski: yeah sorry, I haven't been able to set up IDEA to follow this

stanislavkozlovski:
@hachikuji regarding the tests, I think I could write up a quick python script to continuously add consumers to a group and ensure the proxy works

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +313,103 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+        KafkaConsumerManager.this.updateExpiration(taskState.consumerState);
+        if (!taskState.task.isDone()) {
+          long backoffTime = config.getTime().milliseconds()
+              + consumerConfig.getInt(KafkaRestConfig.CONSUMER_ITERATOR_BACKOFF_MS_CONFIG);
+          delayUntil(Math.min(backoffTime, requestExpiration));
+        } else {
+          log.trace("Finished executing consumer read task ({})", taskState.task);
+        }
+      } catch (Exception e) {
+        log.error("Failed to read records from consumer "
```
hachikuji: nit: convert to use a `{}` substitution. This works even if the last argument is an exception. Also, the other log messages use `taskState.task`. Should we do the same for consistency?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +313,103 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+    public RunnableReadTask(ReadTaskState taskState) {
+      this.taskState = taskState;
+      this.started = config.getTime().milliseconds();
+      this.consumerConfig = taskState.consumerState.getConfig();
+      this.requestExpiration = this.started
+              + consumerConfig.getInt(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG);
+      this.waitExpirationMs = 0;
+    }
+
+    public void delayUntil(long delayMs) {
```
hachikuji: Maybe we could make this `delayFor` and take the delay duration. We can add on `config.getTime().milliseconds()` in the function.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -78,34 +94,51 @@
...
+    int maxThreadCount = config.getInt(CONSUMER_MAX_THREADS_CONFIG) < 0 ? Integer.MAX_VALUE
+            : config.getInt(CONSUMER_MAX_THREADS_CONFIG);
+
+    this.executor = new ThreadPoolExecutor(0, maxThreadCount,
+            60L, TimeUnit.SECONDS,
+            new SynchronousQueue<Runnable>(),
+            new RejectedExecutionHandler() {
+            @Override
+            public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
+              log.debug("The runnable {} was rejected execution."
```
hachikuji: nit: missing space before start of next sentence. Also, if we move this message into the `if` check below, we can probably print a better message which refers to the task state. Either that or we can add good `toString` methods to `RunnableReadTask` and any other `Runnable` implementation we submit to this executor.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +313,103 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+      this.taskState = taskState;
+      this.started = config.getTime().milliseconds();
+      this.consumerConfig = taskState.consumerState.getConfig();
+      this.requestExpiration = this.started
+              + consumerConfig.getInt(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG);
+      this.waitExpirationMs = 0;
+    }
+
+    public void delayUntil(long delayMs) {
+      waitExpirationMs = delayMs;
```
hachikuji: Does it make sense to use `Math.min(delayMs, requestExpiration)` here? I am also wondering if we should check whether expiration has actually been reached and `finish()` the read task directly if it has. Otherwise, a rejected read task might be delayed past its expiration.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
```diff
@@ -93,14 +82,11 @@ public boolean doPartialRead() {
     try {
       // Initial setup requires locking, which must be done on this thread.
       if (messages == null) {
-        parent.startRead();
```
hachikuji: Seems `doPartialRead` should have a void return instead of boolean. The javadoc comment doesn't actually match the implemented behavior (even in the old code) and we just use `isDone` to check for backoff anyway.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +313,103 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
-            if (e != null) {
-              // Ensure caught exceptions are converted to RestExceptions so the user gets a
-              // nice error message. Currently we don't define any more specific errors because
-              // the old consumer interface doesn't classify the errors well like the new
-              // consumer does. When the new consumer is available we may be able to update this
-              // to provide better feedback to the user.
-              Exception responseException = e;
-              if (!(e instanceof RestException)) {
-                responseException = Errors.kafkaErrorException(e);
+    final KafkaConsumerReadTask task = new KafkaConsumerReadTask<KafkaKeyT, KafkaValueT,
```
hachikuji: Should the `ReadCallback` parameter to this function be typed as `ReadCallback<ClientKeyT, ClientValueT>`?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +313,103 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+                    Exception e) {
+              KafkaConsumerManager.this.updateExpiration(state);
+              if (e != null) {
+                // Ensure caught exceptions are converted to RestExceptions so the user gets
+                // a nice error message. Currently we don't define any more specific
+                // errors because the old consumer interface doesn't classify the errors well
+                // like the new consumer does. When the new consumer is available we may be
+                // able to update this to provide better feedback to the user.
+                Exception responseException = e;
+                if (!(e instanceof RestException)) {
```
hachikuji: It's a little annoying that we have to repeat this check since we already did it prior to invoking `onCompletion`. As far as I can tell, we should be able to use the more specific `RestException` type in the callback signature and avoid some of these redundant checks:

```java
  public interface ConsumerWorkerReadCallback <K, V> {
    void onCompletion(List<? extends ConsumerRecord<K, V>> records, RestException e);
  }
```

On a related note, the comment above mentions that we should use `RestException` to ensure the user gets a nice message. That makes sense to me, so perhaps we should do the same thing in `ConsumerReadTask.finish(Exception)`.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -274,33 +313,103 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
-              // to provide better feedback to the user.
-              Exception responseException = e;
-              if (!(e instanceof RestException)) {
-                responseException = Errors.kafkaErrorException(e);
+    final KafkaConsumerReadTask task = new KafkaConsumerReadTask<KafkaKeyT, KafkaValueT,
+          ClientKeyT, ClientValueT>(
+          state,
+          timeout,
+          maxBytes,
+          new ConsumerWorkerReadCallback<ClientKeyT, ClientValueT>() {
```
hachikuji: `ConsumerWorkerReadCallback` seems to have the same signature as `ReadCallback`. Do we need both?

Also, the only reason I can see that we are wrapping the user callback is the call to `updateExpiration`. But now that we are calling `updateExpiration` after every partial read, perhaps we can just get rid of this wrapper?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: I also like that API better but we use it like this in one of the two places:
```java
          delayUntil(Math.min(backoffTime, requestExpiration));
```
and requestExpiration is populated on class instantiation
```java
this.requestExpiration = this.started
              + consumerConfig.getInt(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG);
```

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: Seeing as both callbacks in `ConsumerManager` and `KafkaConsumerManager` expose the same interface, I think it makes sense to have both use one. So I'll rename the `WorkerReadCallback` and delete the local `ReadCallbacks`s

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
```diff
@@ -125,27 +124,27 @@ public boolean doPartialRead() {
...
   }
 
-  private void finish() {
+  void finish() {
     finish(null);
   }
 
   private void finish(Exception e) {
-    log.trace("Finishing KafkaConsumerReadTask id={} exception={}", this, e);
+    log.trace("Finishing KafkaConsumerReadTask id={}", this);
```
hachikuji: We should still include the exception, we just don't need a substitution for it.
```java
log.trace("Finishing KafkaConsumerReadTask id={}", this, e);
```

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -251,7 +245,7 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
       Class<? extends ConsumerState<KafkaKeyT, KafkaValueT, ClientKeyT, ClientValueT>>
           consumerStateType,
       final long maxBytes,
-      final ReadCallback callback
+      final ConsumerReadCallback callback
```
hachikuji: Should this be `ConsumerReadCallback<ClientKeyT, ClientValueT>`?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -58,11 +59,8 @@
   private KafkaRestConfig config;
   private ConsumerInstanceId instanceId;
   private Consumer<KafkaKeyT, KafkaValueT> consumer;
-  private ConsumerRecords<KafkaKeyT, KafkaValueT> consumerRecords = null;
-
-  private List<ConsumerRecord<KafkaKeyT, KafkaValueT>> consumerRecordList = null;
-  private int index = 0;
 
+  private Queue<ConsumerRecord<KafkaKeyT, KafkaValueT>> consumerRecords = null;
```
hachikuji: If we initialize this to an empty collection, then I think we don't need the null check in `hasNextCached`. Is that right?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
stanislavkozlovski: ah, gotcha

Issue #475, opened 2018-09-17
Kafka Rest API incorrect message returned by GET request for specified offset when data is compressed


scottnice:
Test on a compressed topic vs a non-compressed topic. Should fail on compressed topic.

GET /topics/myTopic/partitions/0/messages?offset=10&count=1

Tested using this call on the rest api using docker image version 4.1.0 and 5.0.0 and re-created on both.
When requesting an offset between 0-8 it returns offset 0, when requesting from 9-149 it returns 9, when requesting from 150-265 it returns 150 etc...

Performing a /topics/myTopic/partitions/0/messages?offset=0&count=1000 shows that we have no offset gaps from 0 to 1000

Compressed format:

ByteBufferMessageSet(
LegacyRecordBatch(offset=8, Record(magic=1, attributes=1, compression=GZIP, crc=3359863988, CreateTime=1537196426232, key=0 bytes, value=703 bytes)), 
LegacyRecordBatch(offset=149, Record(magic=1, attributes=1, compression=GZIP, crc=712877603, CreateTime=1537196426317, key=0 bytes, value=6199 bytes)), 

Regular format:
ByteBufferMessageSet(
LegacyRecordBatch(offset=1, Record(magic=1, attributes=0, compression=NONE, crc=1309344491, CreateTime=-1, key=3 bytes, value=4 bytes)), 
LegacyRecordBatch(offset=2, Record(magic=1, attributes=0, compression=NONE, crc=1309344491, CreateTime=-1, key=3 bytes, value=4 bytes)), 
LegacyRecordBatch(offset=3, Record(magic=1, attributes=0, compression=NONE, crc=1309344491, CreateTime=-1, key=3 bytes, value=4 bytes)),

scottnice:
After stepping through the code I found that the call to fetchRecords in SimpleConsumerManager will sometimes (depending on the topic) return a ByteBufferMessageSet where the first message has a lower offset than the one requested via the API.

scottnice:
See this PR #477

scottnice:
Verifying when I get a stable release

PR #477, opened 2018-09-20
Fix SimpleConsumerManager properly return requested offsets for compressed topic

scottnice:
Fixed bug where the API would return the incorrect offset when specific offset was requested and the topic was compressed. The SimpleConsumerManager assumed the first offset in the returned MessageSet was the requested offset, this is not necessarily the case with compressed formats.

See bug #475


hachikuji:
retest this please

scottnice:
This test is failing with or without my changes. 

java.lang.AssertionError: Callback failed to fire
	at io.confluent.kafkarest.v2.KafkaConsumerManagerTest.testConsumerNormalOps(KafkaConsumerManagerTest.java:161)

hachikuji:
@scottnice Yeah, seems unrelated. I will go ahead and merge and then figure out how/whether to backport.

PR #480, opened 2018-09-25
Deprecate kafka-rest specific resource extension mech

big-andy-coates:
in preference to the common one in rest-utils add here https://github.com/confluentinc/rest-utils/pull/119

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -207,8 +207,8 @@ General
   * Default: 1000
   * Importance: low
 
-``kafka.rest.resource.extension.class``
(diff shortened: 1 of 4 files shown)


stanislavkozlovski:
I think we will also want to mark it as deprecated in `docs/config.rst`

big-andy-coates:
Good shout - I've just updated the docs to reflect the preferred way of working, i.e. to use the new settings and types.

stanislavkozlovski:
Is the new functionality for using `resource.extension.classes` handled anywhere? I think I don't see the config parsing it - or is it from rest-utils? @big-andy-coates

PR #481, opened 2018-09-25
CPKAFKA-1681:  Add long polling functionality

stanislavkozlovski:
This commit enables long-polling functionality in the V1 and V2 consumer APIs.

Previously, consumer.request.timeout.ms would control when the proxy sent back a response to the user. The default value was 1000ms, but setting it explicitly to 1000 (or 2000) would cause it to be passed to the underlying consumer connection. This caused a NullPointerException because configuration exceptions were ignored in the V2 API.
This PR stops the behavior of passing the consumer.request.timeout.ms to the underlying consumer.
The proxy now also raises exceptions and returns proper 4xx responses when the given V2 Consumer settings are invalid.

Adds a new config setting:
* fetch.min.bytes - the amount of bytes of records the proxy will collect before immediately returning a response. It is configurable per consumer.
This new setting controls how the user->proxy connection works, not the underlying consumer->broker.

### Future Work
In a future PR aimed at `master`, where we can be more lenient with backwards-compatibility, we should consider renaming some configs and deprecating old ones
* fetch.max.wait.ms - time before a consumer response is sent back when neither min or max bytes are fulfilled
* deprecate consumer.request.max.bytes and introduce proxy.fetch.max.bytes - same config but renamed. If proxy.fetch.max.bytes is explicitly defined it will override the set consumer.request.max.bytes
* proxy.fetch.max.wait.ms and proxy.fetch.min.bytes are configurable per consumer as well as globally
* consumer.request.timeout.ms is totally unused now

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -46,6 +46,9 @@ Java Kafka clients.
   * Importance: high
 
 ``consumer.request.max.bytes``
+  DEPRECATED: see ``fetch.max.bytes``
+
+``fetch.max.bytes``
   Maximum number of bytes in unencoded message keys and values returned by a single request. This can be used by administrators to limit the memory used by a single consumer and to control the memory usage required to decode responses on clients that cannot perform a streaming decode. Note that the actual payload will be larger due to overhead from base64 encoding the response data and from JSON encoding the entire response.
 
   * Type: long
@@ -60,6 +63,7 @@ Java Kafka clients.
   * Importance: medium
 
 ``consumer.request.timeout.ms``
+  DEPRECATED: see ``fetch.max.wait.ms``
   The maximum total time to wait for messages for a request if the maximum number of messages has not yet been reached.
 
   * Type: int
@@ -109,6 +113,21 @@ Java Kafka clients.
   * Type: int
   * Default: 50
   * Importance: low
+
+
+``fetch.max.wait.ms``
+  Maximum amount of time the proxy will wait before returning a consumer response. Note that a response will be returned earlier if ``fetch.min.bytes`` is fulfilled.
+
+  * Type: int
+  * Default: 1000
+  * Importance: medium
+
+``fetch.min.bytes``
+  Minimum number of bytes in message keys and values returned by a single request before the timeout of `fetch.max.wait.ms` passes.
+
+  * Type: int
+  * Default: 1000
+  * Importance: medium
 
 ``consumer.iterator.timeout.ms``
   Timeout for blocking consumer iterator operations. This should be set to a small enough value that it is possible to effectively peek() on the iterator.
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -46,9 +46,6 @@ Java Kafka clients.
   * Importance: high
 
 ``consumer.request.max.bytes``
-  DEPRECATED: see ``fetch.max.bytes``
-
-``fetch.max.bytes``
   Maximum number of bytes in unencoded message keys and values returned by a single request. This can be used by administrators to limit the memory used by a single consumer and to control the memory usage required to decode responses on clients that cannot perform a streaming decode. Note that the actual payload will be larger due to overhead from base64 encoding the response data and from JSON encoding the entire response.
 
   * Type: long
@@ -63,7 +60,6 @@ Java Kafka clients.
   * Importance: medium
 
 ``consumer.request.timeout.ms``
-  DEPRECATED: see ``fetch.max.wait.ms``
   The maximum total time to wait for messages for a request if the maximum number of messages has not yet been reached.
 
   * Type: int
@@ -114,19 +110,12 @@ Java Kafka clients.
   * Default: 50
   * Importance: low
 
-
-``fetch.max.wait.ms``
-  Maximum amount of time the proxy will wait before returning a consumer response. Note that a response will be returned earlier if ``fetch.min.bytes`` is fulfilled.
-
-  * Type: int
-  * Default: 1000
-  * Importance: medium
-
 ``fetch.min.bytes``
-  Minimum number of bytes in message keys and values returned by a single request before the timeout of `fetch.max.wait.ms` passes.
+  Minimum number of bytes in message keys and values returned by a single request before the timeout of `consumer.request.timeout.ms` passes.
+  The special sentinel value of -1 disables this functionality.
 
   * Type: int
-  * Default: 1000
+  * Default: -1
   * Importance: medium
 
 ``consumer.iterator.timeout.ms``
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -178,7 +178,7 @@ changing these based on your specific use case.
      before the timeout of `consumer.request.timeout.ms` passes.
 
      * Type: int
-     * Default: the value of `consumer.request.max.bytes`
+     * Default: -1
      * Importance: medium
 
    ``consumer.request.timeout.ms``
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -156,9 +156,6 @@ Other settings are important to the health and performance of the proxy. You sho
 changing these based on your specific use case.
 
    ``consumer.request.max.bytes``
-     DEPRECATED: See ``fetch.max.bytes``
-
-   ``fetch.max.bytes``
      Maximum number of bytes in message keys and values returned by a single request.
      Smaller values reduce the maximum memory used by a single consumer and may be helpful to
      clients that cannot perform a streaming decode of responses, limiting the maximum memory
@@ -176,24 +173,15 @@ changing these based on your specific use case.
      * Default: 67108864
      * Importance: medium
 
-   ``fetch.max.wait.ms``
-     The maximum amount of time the proxy will wait before returning a consumer response if
-     there isn't sufficient daa to satisfy the requirement given by ``fetch.min.bytes``.
-
-     * Type: int
-     * Default: 1000
-     * Importance: medium
-
    ``fetch.min.bytes``
      The minimum number of bytes in message keys and values returned by a single request
-     before the timeout of `fetch.max.wait.ms` passes.
+     before the timeout of `consumer.request.timeout.ms` passes.
 
      * Type: int
-     * Default: 1000
+     * Default: the value of `consumer.request.max.bytes`
      * Importance: medium
 
    ``consumer.request.timeout.ms``
-     DEPRECATED: see ``fetch.max.wait.ms``
      The maximum total time to wait for messages for a request if the maximum request size has
      not yet been reached. The consumer uses a timeout to enable batching. A larger value will
      allow the consumer to wait longer, possibly including more messages in the response.
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -156,6 +156,9 @@ Other settings are important to the health and performance of the proxy. You sho
 changing these based on your specific use case.
 
    ``consumer.request.max.bytes``
+     DEPRECATED: See ``fetch.max.bytes``
+
+   ``fetch.max.bytes``
      Maximum number of bytes in message keys and values returned by a single request.
      Smaller values reduce the maximum memory used by a single consumer and may be helpful to
      clients that cannot perform a streaming decode of responses, limiting the maximum memory
@@ -173,7 +176,24 @@ changing these based on your specific use case.
      * Default: 67108864
      * Importance: medium
 
+   ``fetch.max.wait.ms``
+     The maximum amount of time the proxy will wait before returning a consumer response if
+     there isn't sufficient daa to satisfy the requirement given by ``fetch.min.bytes``.
+
+     * Type: int
+     * Default: 1000
+     * Importance: medium
+
+   ``fetch.min.bytes``
+     The minimum number of bytes in message keys and values returned by a single request
+     before the timeout of `fetch.max.wait.ms` passes.
+
+     * Type: int
+     * Default: 1000
+     * Importance: medium
+
    ``consumer.request.timeout.ms``
+     DEPRECATED: see ``fetch.max.wait.ms``
      The maximum total time to wait for messages for a request if the maximum request size has
      not yet been reached. The consumer uses a timeout to enable batching. A larger value will
      allow the consumer to wait longer, possibly including more messages in the response.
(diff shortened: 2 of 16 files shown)


stanislavkozlovski:
@hachikuji do you think we should outright remove `consumer.request.timeout.ms` from the code OR have it work similar like `consumer.request.max.bytes` does now - namely, if `fetch.max.wait.ms` is defined it'll be used, otherwise we will use `consumer.request.timeout.ms` ?

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -133,28 +133,180 @@ public void testConsumerOverrides() {
...
+        expectCreate();
+        EasyMock.replay(mdObserver, consumerFactory);
+        String cid = consumerManager.createConsumer(
+            groupName, new ConsumerInstanceConfig(EmbeddedFormat.BINARY));
+        consumerManager.subscribe(groupName, cid, new ConsumerSubscriptionRecord(Collections.singletonList(topicName), null));
+
+        readFromDefault(cid);
+        assertEquals(config.getInt(KafkaRestConfig.PROXY_FETCH_MAX_WAIT_MS_CONFIG),
+            Integer.parseInt(KafkaRestConfig.PROXY_FETCH_MAX_WAIT_MS_DEFAULT));
+        Thread.sleep(config.getInt(KafkaRestConfig.PROXY_FETCH_MAX_WAIT_MS_CONFIG));
```
stanislavkozlovski: In the `KafkaConsumerManagerTest` we need to call `Thread.sleep()` as `KafkaConsumerReadTask` does not implement the `Future` interface like the v1 `ConsumerReadTask` does.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -50,10 +49,7 @@
 import io.confluent.rest.RestConfigException;
 import org.junit.runner.RunWith;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
stanislavkozlovski: Do we care if we `import *` in tests? This wasn't intentional but I now see that other test classes have that as well

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerInstanceConfig.java
```diff
@@ -33,22 +37,27 @@
...
 
   public ConsumerInstanceConfig(
       @JsonProperty("id") String id,
       @JsonProperty("name") String name,
       @JsonProperty("format") String format,
       @JsonProperty("auto.offset.reset") String autoOffsetReset,
-      @JsonProperty("auto.commit.enable") String autoCommitEnable
+      @JsonProperty("auto.commit.enable") String autoCommitEnable,
+      @JsonProperty(KafkaRestConfig.PROXY_FETCH_MIN_BYTES_CONFIG)
+              String responseMinBytes,
```
hachikuji: It's a little curious that we have to use string types. I thought jackson was pretty clever with type detection.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerInstanceConfig.java
```diff
@@ -88,6 +116,68 @@ public String getName() {
...
+    }
+
+    try {
+      this.requestWaitMs = Integer.parseInt(requestWaitMs);
+      KafkaRestConfig.PROXY_FETCH_MAX_WAIT_MS_VALIDATOR.ensureValid(
+              KafkaRestConfig.PROXY_FETCH_MAX_WAIT_MS_CONFIG,
+              this.requestWaitMs
+      );
+    } catch (io.confluent.common.config.ConfigException e) {
+      throw new RestConstraintViolationException(
```
hachikuji: Is it helpful to have an error code like we have for consumer config exceptions?

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
hachikuji: Just my personal preference, but I like tests to follow the same conventions as regular code. Not sure why we don't enforce checkstyle on test cases...

docs/config.rst
```diff
@@ -110,6 +114,21 @@ Java Kafka clients.
...
+
+  * Type: int
+  * Default: 1000
+  * Importance: medium
+
+``fetch.min.bytes``
+  Minimum number of bytes in message keys and values returned by a single request before the timeout of `fetch.max.wait.ms` passes.
+
+  * Type: int
+  * Default: 1000
```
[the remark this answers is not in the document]
stanislavkozlovski: That's a good point. I frankly don't remember why I chose 1000.

I agree that 1 is the ideal default value.

My concern now is that we might be changing the behavior of the proxy. Since we plan on releasing this in a minor version `4.1.x`, this might not be a good thing to do. 
Users up until now most likely tuned `consumer.request.max.bytes` to receive a certain payload from the proxy every time. If we were to change this behavior to return at the first sight of data, users' applications might start issuing requests more frequently than hoped for which could bring up issues like higher CPU use.

Even though it is a bit of an indirect change, I'm wondering if it is worth it to disable the behavior by default until 6.0

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerInstanceConfig.java
```diff
@@ -69,10 +78,29 @@ public ConsumerInstanceConfig(
...
+   * Attaches proxy-specific configurations to the given Properties object
+   */
+  public static Properties attachProxySpecificProperties(Properties props,
+                                                         ConsumerInstanceConfig config) {
+    if (config.getResponseMinBytes() != null) {
+      props.setProperty(KafkaRestConfig.PROXY_FETCH_MIN_BYTES_CONFIG,
+              config.getResponseMinBytes().toString());
+    }
+    if (config.getRequestWaitMs() != null) {
+      props.setProperty(KafkaRestConfig.PROXY_FETCH_MAX_WAIT_MS_CONFIG,
```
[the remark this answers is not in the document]
stanislavkozlovski: Ugh, nice catch.

Agreed that we don't want the proxy to use memory when it has the option not to. I think we need to propagate `fetch.max.wait.ms` as well since the configs are tied together in behavior.

e.g suppose we have `fetch.min.bytes=100KB` and `fetch.max.wait=60s` set. The default `fetch.max.wait.ms` is 500 for the AK consumer. The behavior isn't changed - the proxy will receive records every half a second from the consumer and that will use up memory until either `min.bytes` is reached or 60 seconds pass.
Does that make sense?

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
```diff
@@ -65,22 +73,29 @@ public ConsumerReadTask(
...
+        maxBytes,
+        conf.consumerResponseMaxBytes()
     );
     this.callback = callback;
     this.finished = new CountDownLatch(1);
 
-    started = parent.getConfig().getTime().milliseconds();
+    this.requestTimeoutMs = conf.getInt(KafkaRestConfig.PROXY_FETCH_MAX_WAIT_MS_CONFIG);
+    int responseMinBytes = conf.getInt(KafkaRestConfig.PROXY_FETCH_MIN_BYTES_CONFIG);
+    this.responseMinBytes = responseMinBytes < 0 ? Integer.MAX_VALUE : responseMinBytes;
```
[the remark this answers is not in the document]
stanislavkozlovski: Negative min.bytes means the feature is disabled. This is to preserve backward compatibility with the `max.bytes` setting. Like I said earlier, I don't remember why I set 1000 as the default for min.bytes - it must have been -1 and got lost in a merge conflict.

With min.bytes -1, the proxy continues working as before with the `max.bytes`  or `timeout.ms` telling it when to return

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerInstanceConfig.java
stanislavkozlovski: You're most likely right. I'm not familiar with the library

stanislavkozlovski:
Thanks for the review @hachikuji. I addressed the smaller comments. The bigger question here is how we want to approach backward compatibility since this is going out in a minor version

docs/deployment.rst
```diff
@@ -173,6 +173,14 @@ changing these based on your specific use case.
      * Default: 67108864
      * Importance: medium
 
+   ``fetch.min.bytes``
+     The minimum number of bytes in message keys and values returned by a single request
+     before the timeout of `consumer.request.timeout.ms` passes.
+
+     * Type: int
+     * Default: the value of `consumer.request.max.bytes`
```
hachikuji: Should be -1? We should file a jira to avoid this duplication.

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
```diff
@@ -115,13 +128,18 @@ public void doPartialRead() {
           ConsumerRecordAndSize<ClientKeyT, ClientValueT> recordAndSize =
                   parent.createConsumerRecord(msg);
           roughMsgSize = recordAndSize.getSize();
-          if (bytesConsumed + roughMsgSize >= maxResponseBytes) {
+          this.willExceedMaxResponseBytes = bytesConsumed + roughMsgSize >= maxResponseBytes;
```
hachikuji: This value is recomputed on every call to `doPartialRead` and it is only used from within `doPartialRead`. Couldn't it be a local variable instead of a field? Same for `exceededMinResponseBytes`.

stanislavkozlovski: https://github.com/confluentinc/kafka-rest/blob/7ce62592ba58bf4cf71e176146ea68f7555d63be/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerReadTask.java#L97

We assign them when inheriting a previous task which might have failed. I think the problem is that we do not respect those in that case.

docs/deployment.rst
stanislavkozlovski: True, nice catch.

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerReadTask.java
hachikuji: I guess my point is that we can always recompute this using `bytesConsumed`, so why do we need the additional fields?

stanislavkozlovski:
Hey @mente, thanks for the question!
I've updated the PR's description with the following:

### How to use
To enable long-polling, simply configure `consumer.request.timeout.ms` to the desired time in milliseconds. If you want to return earlier than that if a given byte threshold is reached, configure `fetch.min.bytes`. 

Does that help?

mente:
Thanks! It should help, I would like to verify later.

IMHO in general your description goes against the one in [rest proxy api](https://docs.confluent.io/current/kafka-rest/docs/api.html#post--consumers-(string-group_name)):

> consumer.request.timeout.ms (string) – Sets the consumer.request.timeout.ms setting for this consumer specifically. This setting controls the maximum total time to wait for messages for a request if the maximum request size has not yet been reached. It does not affect the underlying consumer->broker connection

The part

> if the maximum request size has not yet been reached

led me think that `max_bytes` from [/consumers/.../records](https://docs.confluent.io/current/kafka-rest/docs/api.html#get--consumers-(string-group_name)-instances-(string-instance)-records) endpoint controls the maximum request size and I was surprised to find out that timeout is still equal to`consumer.request.timeout.ms` despite having records in the result.

mente:
TL;DR please update documentation :)

mente:
I've just rechecked, you're absolutely right. I was using too big `max_bytes` and it wasn't hit 🤦‍♂️

mente:
Follow up question - if `consumer.request.timeout.ms` provides long polling in combination with `max_bytes` and `fetch.min.bytes` what's stands behind the logic of `timeout` in `/consumers/.../records` endpoint?

stanislavkozlovski:
@mente 
`timeout` is `consumer.request.timeout.ms` for that specific request - it allows you to override it on a per-request basis. But it's not a total override, it would only take it if [it is less than the configured consumer/proxy-level ](https://github.com/confluentinc/kafka-rest/blob/5b5fd70fc03471092aad920674f995b153b03de8/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java#L76)`consumer.request.timeout.ms` (note that you can configure that [timeout per consumer](https://docs.confluent.io/current/kafka-rest/docs/api.html#post--consumers-(string-group_name)) otherwise we'd take whats in the proxy config file)

Our documentation there is stale though, the default is not 5000ms but rather -1 (disabled). Nice catch. Do you want to contribute a PR to fix that (since you found it)? If not, just say and I'll open one :)

mente:
Now I'm even more lost than I was before :). Please review if I'm right:

long polling is when rest proxy doesn't return response until certain criterias are met - being it max expected response size is reached (via `max_bytes`) or minimal response size is reached (via `fetch.min.bytes` or timeout is reached (via timeout which can be defined in multiple places).

Timeout is controlled in 2 places (above overrides below one):
* `timeout` query parameter in [`/consumers/.../records`](https://docs.confluent.io/current/kafka-rest/docs/api.html#get--consumers-(string-group_name)-instances-(string-instance)-records) endpoint. Taken only if below one `consumer.timeout.ms` is greater than
* `consumer.timeout.ms` in body request for [`/consumer/*name*`](https://docs.confluent.io/current/kafka-rest/docs/api.html#post--consumers-(string-group_name)) endpoint

Prior to this PR `timeout` served different purpose and long-polling functionality wasn't present

mente:
> Do you want to contribute a PR to fix that (since you found it)?

I can, as long as I understand how long polling functionality is working and how parameters were used before :)

mente:
Okay, so this PR is bug fixing of long-polling which apparently was not working + more parameters added to fine-tune long polling from client-side. Thanks!

PR #485, opened 2018-09-27
Revert Consumer Threading 5.0

stanislavkozlovski:
Since we went with a slightly different model for consumer threading in 4.0 (https://github.com/confluentinc/kafka-rest/pull/476), we are going to be merging that upstream to 5.0.x and trunk. The merge would be cleaner if we first reverted the previous changes

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -58,18 +58,18 @@ General
   * Default: 67108864
   * Importance: medium
 
-``consumer.threads``
-  The maximum number of threads to run consumer requests on. Note that this must be greater than the maximum number of consumers in a single consumer group.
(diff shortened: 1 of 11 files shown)


hachikuji:
Thanks @stanislavkozlovski. I'm a little confused why the spotbugs change is also pulled in.

stanislavkozlovski:
@hachikuji I think the difference is that I committed it on my fork's `master` branch while Ewen committed it himself to this repo's `master` branch: https://github.com/confluentinc/kafka-rest/commit/078a213e3380a43bc233cad8316c2c831209c430

Is it problematic if we merge this?

PR #487, opened 2018-09-28
Merge 4.1.x into 5.0.x

stanislavkozlovski:
`pint merge` produced conflicts and I can't push directly to 5.0.x since it's a protected branch

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -18,7 +18,7 @@
 
     <suppress
             checks="(ClassDataAbstractionCoupling)"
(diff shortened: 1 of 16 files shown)


stanislavkozlovski:
On second thought, this shouldn't have conflicts. The reason it has conflicts is because I forgot to merge (https://github.com/confluentinc/kafka-rest/pull/485) into 5.0.x

We originally implemented consumer threading one way and merged that into 5.0.x and master. Later we realized we need those changes in 4.1.x.
We iterated through a lot of implementations and now the 4.1.x code and 5.0.x+ is very different. 
We'll now revert everything we did 5.0.x+ and re-merge 4.1.x

PR #490, opened 2018-10-01
[WIP] Consumer x-forwarded-port and x-forwarded-protocol headers

lprashant-94:
Use X-Forwarded-Protocol and X-Forwarded-Port header while building base_uri in Consumer Create call.


lprashant-94:
[clabot:check]

PR #500, opened 2018-10-29
the ThreadPoolExecutor RejectedExecutionHandler was error

huangyiminghappy:
```  java 
this.executor = new KafkaConsumerThreadPoolExecutor(0, maxThreadCount,
            60L, TimeUnit.SECONDS,
            new SynchronousQueue<Runnable>(),
            new RejectedExecutionHandler() {
            @Override
            public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
              log.debug("The runnable {} was rejected execution. "
                  + "The thread pool must be satured or shutiing down", r);
              if (r instanceof ReadFutureTask) {
                ReadFutureTask readFutureTask = (ReadFutureTask) r;
                RunnableReadTask readTask = (RunnableReadTask)readFutureTask.getReadTask();
                readTask.delayFor(ThreadLocalRandom.current().nextInt(25, 76));
              } else {
                // run commitOffset and consumer close tasks from the caller thread
                if (!executor.isShutdown()) {
                  r.run();
                }
              }
            }
          }
    );
```   
in RejectedExecutionHandler the rejectedExecution method,the runnable is the future task type,so Can't convert directly like  

```  java
 RunnableReadTask readTask = (r)readFutureTask.getReadTask();
``` 
because the runnable is the future task type.
wo can override the 
```  java
  public KafkaConsumerThreadPoolExecutor(int corePoolSize,
                                               int maximumPoolSize,
                                               long keepAliveTime,
                                               TimeUnit unit,
                                               BlockingQueue<Runnable> workQueue,
                                               RejectedExecutionHandler handler) {
       super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue, handler);

     }

       @Override
       protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
        return new ReadFutureTask(runnable, value);
       }
    }
``` 
and get the task like this 
```   java
if (r instanceof ReadFutureTask) {
                ReadFutureTask readFutureTask = (ReadFutureTask) r;
                RunnableReadTask readTask = (RunnableReadTask)readFutureTask.getReadTask();
                readTask.delayFor(ThreadLocalRandom.current().nextInt(25, 76));
              }
```

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -407,7 +407,7 @@ public class ConsumerManager {
       if (callable instanceof RunnableReadTask) {
         return new ReadFutureTask(callable);
       }
-      return super.newTaskFor((callable);
+      return super.newTaskFor((callable));
     }
 
   }
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -37,11 +37,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.Future;
 import java.util.concurrent.Callable;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.RunnableFuture;
-import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.BlockingQueue;
 
 import javax.ws.rs.core.Response;
@@ -100,7 +98,7 @@ public class ConsumerManager {
           @Override
           public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
             if (r instanceof ReadFutureTask) {
-             RunnableReadTask readTask = ((ReadFutureTask)r).getReadTask();
+              RunnableReadTask readTask = ((ReadFutureTask)r).getReadTask();
               int delayMs = ThreadLocalRandom.current().nextInt(25, 75 + 1);
               readTask.waitExpirationMs = config.getTime().milliseconds() + delayMs;
               delayedReadTasks.add(readTask);
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -360,26 +360,15 @@ public class ConsumerManager {
 
   class ReadFutureTask<V> extends FutureTask<V> {
 
-    private Runnable myTask;
-
-    private Callable callable;
+    private final Runnable readTask;
 
     public ReadFutureTask(Runnable runnable, V result) {
       super(runnable, result);
-      this.myTask = runnable;
-    }
-
-    public ReadFutureTask(Callable callable) {
-      super(callable);
-      this.callable = callable;
+      this.readTask = runnable;
     }
 
     public RunnableReadTask getReadTask() {
-      if (myTask != null) {
-        return (RunnableReadTask)myTask;
-      } else {
-        return (RunnableReadTask)callable;
-      }
+        return (RunnableReadTask)readTask;
     }
   }
 
@@ -400,14 +389,6 @@ public class ConsumerManager {
       }
       return super.newTaskFor(runnable, value);
     }
-
-    protected <T> RunnableFuture<T> newTaskFor(Callable<T> callable) {
-      if (callable instanceof RunnableReadTask) {
-        return new ReadFutureTask(callable);
-      }
-      return super.newTaskFor((callable));
-    }
-
   }
 
   class RunnableReadTask implements Runnable, Delayed {
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -37,7 +37,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.Callable;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.RunnableFuture;
 import java.util.concurrent.BlockingQueue;
@@ -368,7 +367,7 @@ public class ConsumerManager {
     }
 
     public RunnableReadTask getReadTask() {
-        return (RunnableReadTask)readTask;
+      return (RunnableReadTask)readTask;
     }
   }
 
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -37,6 +37,12 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.Future;
+import java.util.concurrent.Callable;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.RunnableFuture;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.BlockingQueue;
 
 import javax.ws.rs.core.Response;
 
@@ -87,14 +93,14 @@ public class ConsumerManager {
     int maxThreadCount = config.getInt(KafkaRestConfig.CONSUMER_MAX_THREADS_CONFIG) < 0
         ? Integer.MAX_VALUE : config.getInt(KafkaRestConfig.CONSUMER_MAX_THREADS_CONFIG);
 
-    this.executor = new ThreadPoolExecutor(0, maxThreadCount,
+    this.executor = new KafkaConsumerThreadPoolExecutor(0, maxThreadCount,
         60L, TimeUnit.SECONDS,
         new SynchronousQueue<Runnable>(),
         new RejectedExecutionHandler() {
           @Override
           public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
-            if (r instanceof RunnableReadTask) {
-              RunnableReadTask readTask = (RunnableReadTask) r;
+            if (r instanceof ReadFutureTask) {
+             RunnableReadTask readTask = ((ReadFutureTask)r).getReadTask();
               int delayMs = ThreadLocalRandom.current().nextInt(25, 75 + 1);
               readTask.waitExpirationMs = config.getTime().milliseconds() + delayMs;
               delayedReadTasks.add(readTask);
@@ -352,6 +358,58 @@ public class ConsumerManager {
   public interface ConsumerFactory {
 
     ConsumerConnector createConsumer(ConsumerConfig config);
+  }
+
+  class ReadFutureTask<V> extends FutureTask<V> {
+
+    private Runnable myTask;
+
+    private Callable callable;
+
+    public ReadFutureTask(Runnable runnable, V result) {
+      super(runnable, result);
+      this.myTask = runnable;
+    }
+
+    public ReadFutureTask(Callable callable) {
+      super(callable);
+      this.callable = callable;
+    }
+
+    public RunnableReadTask getReadTask() {
+      if (myTask != null) {
+        return (RunnableReadTask)myTask;
+      } else {
+        return (RunnableReadTask)callable;
+      }
+    }
+  }
+
+  class KafkaConsumerThreadPoolExecutor extends ThreadPoolExecutor {
+    public KafkaConsumerThreadPoolExecutor(int corePoolSize,
+                                           int maximumPoolSize,
+                                           long keepAliveTime,
+                                           TimeUnit unit,
+                                           BlockingQueue<Runnable> workQueue,
+                                           RejectedExecutionHandler handler) {
+      super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue, handler);
+    }
+
+    @Override
+    protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
+      if (runnable instanceof RunnableReadTask) {
+        return new ReadFutureTask(runnable, value);
+      }
+      return super.newTaskFor(runnable, value);
+    }
+
+    protected <T> RunnableFuture<T> newTaskFor(Callable<T> callable) {
+      if (callable instanceof RunnableReadTask) {
+        return new ReadFutureTask(callable);
+      }
+      return super.newTaskFor((callable);
+    }
+
   }
 
   class RunnableReadTask implements Runnable, Delayed {
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -97,7 +97,7 @@ public class ConsumerManager {
           @Override
           public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
             if (r instanceof ReadFutureTask) {
-              RunnableReadTask readTask = ((ReadFutureTask)r).getReadTask();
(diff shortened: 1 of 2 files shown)


stanislavkozlovski:
Hey @huangyiminghappy. Thanks for the PR! 

Could you help clarify what the benefits of extending both classes are?

To me it looks like it just introduces more code for no benefit, hampering readability somewhat

huangyiminghappy:
> Hey @huangyiminghappy. Thanks for the PR!
> 
> Could you help clarify what the benefits of extending both classes are?
> 
> To me it looks like it just introduces more code for no benefit, hampering readability somewhat
here is the example:
```  java 
public class ExecutorTest {
    static class MyThreadPoolExecutor extends ThreadPoolExecutor {

        public MyThreadPoolExecutor(int corePoolSize,
                                    int maximumPoolSize,
                                    long keepAliveTime,
                                    TimeUnit unit,
                                    BlockingQueue<Runnable> workQueue,
                                    RejectedExecutionHandler handler) {
            super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue, handler);
        }

        @Override
        protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
            return new MyFutureTask(runnable, value);
        }
    }

    public static void main(String[] args) {

        ExecutorService executor = new ThreadPoolExecutor(0, 1,
                60L, TimeUnit.SECONDS,
                new SynchronousQueue<Runnable>(),
                new RejectedExecutionHandler() {
                    @Override
                    public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
                        if (r instanceof Read) {
                            Read read = (Read)r;
                            System.out.println("test cast : " + read.getName());
                        } else {
                            if(r instanceof FutureTask){
                                System.out.println("yes,i am the type of FutureTask");
                            }
                            if (!executor.isShutdown()) {
                                r.run();
                            }
                        }
                    }
                }
        );

        for (int i = 0; i < 10; i++) {
            //Read read = new Read("thread--"+i);

            executor.submit(new Read("thread--" + i));
        }
    }

}

class MyFutureTask<V> extends FutureTask<V> {

    private Runnable myTask;

    public MyFutureTask(Runnable runnable, V result) {
        super(runnable, result);
        this.myTask = runnable;
    }

    public Runnable getMyTask() {
        return myTask;
    }
}

class Read implements Runnable {
    private String name;

    public Read(String name) {
        this.name = name;

    }

    @Override
    public void run() {
        System.out.println("====" + name + " begin run");
        try {
            Thread.sleep(5000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
        System.out.println("====" + name + " end ");

    }

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }
}

```




> Hey @huangyiminghappy. Thanks for the PR!
> 
> Could you help clarify what the benefits of extending both classes are?
> 
> To me it looks like it just introduces more code for no benefit, hampering readability somewhat



> Hey @huangyiminghappy. Thanks for the PR!
> 
> Could you help clarify what the benefits of extending both classes are?
> 
> To me it looks like it just introduces more code for no benefit, hampering readability somewhat

```  java 
import java.util.concurrent.*;

public class ExecutorTest {
    static class MyThreadPoolExecutor extends ThreadPoolExecutor {

        public MyThreadPoolExecutor(int corePoolSize,
                                    int maximumPoolSize,
                                    long keepAliveTime,
                                    TimeUnit unit,
                                    BlockingQueue<Runnable> workQueue,
                                    RejectedExecutionHandler handler) {
            super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue, handler);
        }

        @Override
        protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
            return new MyFutureTask(runnable, value);
        }
    }

    public static void main(String[] args) {

        ExecutorService executor = new MyThreadPoolExecutor(0, 1,
                60L, TimeUnit.SECONDS,
                new SynchronousQueue<Runnable>(),
                new RejectedExecutionHandler() {
                    @Override
                    public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
                        System.out.println("go to the rejectedExecution");
                        if (r instanceof Read) {
                            Read read = (Read) r;
                            System.out.println("test cast : " + read.getName());
                        } else {
                            if (r instanceof FutureTask) {
                                System.out.println("yes,i am the type of FutureTask");
                                MyFutureTask futureTask = (MyFutureTask) r;
                                System.out.println("you can put the runnable to delayQueue" + futureTask.getMyTask());
                            }
                            if (!executor.isShutdown()) {
                                r.run();
                            }
                        }
                    }
                }
        );

        for (int i = 0; i < 10; i++) {
            //Read read = new Read("thread--"+i);

            executor.submit(new Read("thread--" + i));
        }
    }

}

class MyFutureTask<V> extends FutureTask<V> {

    private Runnable myTask;

    public MyFutureTask(Runnable runnable, V result) {
        super(runnable, result);
        this.myTask = runnable;
    }

    public Runnable getMyTask() {
        return myTask;
    }
}

class Read implements Runnable {
    private String name;

    public Read(String name) {
        this.name = name;

    }

    @Override
    public void run() {
        System.out.println("====" + name + " begin run");
        try {
            Thread.sleep(5000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
        System.out.println("====" + name + " end ");

    }

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }
}

``` 
because in RejectedExecutionHandler#rejectedExecution method  ,the runnable is not  the type of RunnableReadTask ,so 
``` java
if (r instanceof RunnableReadTask) {	              if (r instanceof ReadFutureTask) {
                RunnableReadTask readTask = (RunnableReadTask) r;	                ReadFutureTask readFutureTask = (ReadFutureTask) r;
                RunnableReadTask readTask = (RunnableReadTask)readFutureTask.getReadTask();
                readTask.delayFor(ThreadLocalRandom.current().nextInt(25, 76));	                readTask.delayFor(ThreadLocalRandom.current().nextInt(25, 76));
              } 
``` 
is never true.
that because the task is be build as the RunnableFuture type:
image
so we override the executor's newTaskFor method ,throuth the newTaskFor method build our object to get runnable.

stanislavkozlovski:
I see... I just tested this locally and confirm that this is a bug. Great catch @huangyiminghappy!

The implications of this bug are that the read tasks are executed in the Jetty Request Threads when the executor is full. This is not ideal and does not really respect the `consumer.threads` bound.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -35,16 +35,25 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.RunnableFuture;
 import java.util.concurrent.Delayed;
-import java.util.concurrent.RejectedExecutionHandler;
-import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Future;
+
+
+
```
stanislavkozlovski: nit: Can we remove the extra new lines here?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -35,16 +35,25 @@
 import java.util.Properties;
 import java.util.UUID;
 import java.util.Vector;
-import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.FutureTask;
+
+
```
stanislavkozlovski: nit: Extra new lines

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -315,6 +325,37 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+                                               TimeUnit unit,
+                                               BlockingQueue<Runnable> workQueue,
+                                               RejectedExecutionHandler handler) {
+    super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue, handler);
+
+    }
+
+    @Override
+    protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
+      return new ReadFutureTask(runnable, value);
```
stanislavkozlovski: We only want to wrap `RunnableReadTask` in `ReadFutureTask` since that's the class that requires special handling.
We could do it similar to this:
```
if (runnable instanceof RunnableReadTask) {
      return new ReadFutureTask(runnable, value);
}
return super.newTaskFor(runnable, value);
```

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -315,6 +325,37 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+                                               long keepAliveTime,
+                                               TimeUnit unit,
+                                               BlockingQueue<Runnable> workQueue,
+                                               RejectedExecutionHandler handler) {
+    super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue, handler);
+
+    }
+
+    @Override
+    protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
```
stanislavkozlovski: Do we want to also override the `protected <T> RunnableFuture<T> newTaskFor(Callable<T> callable)` ? Which one gets called?
It might be better to override both in my opinion

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -315,6 +325,37 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+
+  class KafkaConsumerThreadPoolExecutor extends ThreadPoolExecutor {
+    public KafkaConsumerThreadPoolExecutor(int corePoolSize,
+                                               int maximumPoolSize,
+                                               long keepAliveTime,
+                                               TimeUnit unit,
+                                               BlockingQueue<Runnable> workQueue,
+                                               RejectedExecutionHandler handler) {
+    super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue, handler);
+
```
stanislavkozlovski: nit: extra new line

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -315,6 +325,37 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
...
+  class ReadFutureTask<V> extends FutureTask<V> {
+
+    private Runnable myTask;
+
+    public ReadFutureTask(Runnable runnable, V result) {
+      super(runnable, result);
+      this.myTask = runnable;
+    }
+
+    public Runnable getReadTask() {
```
stanislavkozlovski: Since this class is meant to wrap `RunnableReadTask`, I think this method should return `RunnableReadTask` directly

huangyiminghappy: ok,i will accept you suggest

huangyiminghappy: > ass is meant to wrap `RunnableReadTask`, I think
hi,do you think to modify like this ?: 
``` java
class MyFutureTask<V> extends FutureTask<V> {

    private Runnable myTask;

    private Callable callable;

    private V result;


    public MyFutureTask(Runnable runnable, V result) {
        super(runnable, result);
        this.myTask = runnable;
        this.result = result;
    }

    public MyFutureTask(Callable callable) {
        super(callable);
     }

    public Read getMyTask() {
        return (Read) myTask;
    }

    public V getResult() {
        return result;
    }
}
``` 
in the MyFutureTask,the class is just for this executor,so wo cast the runnable to the class directly

stanislavkozlovski: I don't understand, `MyFutureTask` is not in the committed changes.
My idea is to have only one cast in lines 132-134.
Current:
```
                ReadFutureTask readFutureTask = (ReadFutureTask) r;
                RunnableReadTask readTask = (RunnableReadTask)readFutureTask.getReadTask();
```
Proposed:
```
                RunnableReadTask readTask = ((ReadFutureTask) r).getReadTask();
```

huangyiminghappy: ok,i unstand it .

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
huangyiminghappy: because we need to support executor.submit(runnable) or executor.submit(callable),so i override the protected <T> RunnableFuture<T> newTaskFor(Callable<T> callable) and in 
ReadFutureTask's getTask method,use the consistent code to return RunnableReadTask

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: @huangyiminghappy I think we need to do this as we don't always call `executor.submit()` on `RunnableReadTask`

huangyiminghappy:
@huangyiminghappy I think we need to do this as we don't always call executor.submit() on RunnableReadTask
=========
@stanislavkozlovski . do you mean that the code 
``` java

public RunnableReadTask getReadTask() {
      if (myTask != null) {
        return (RunnableReadTask)myTask;
      } else {
        return (RunnableReadTask)callable;
      }
    }
  }
``` 
modify like 
``` java
public Object getReadTask() {
      if (myTask != null) {
        return myTask;
      } else {
        return callable;
      }
    }
  }
```

huangyiminghappy:
@stanislavkozlovski  what do you think if i create new java class like  KafkaConsumerThreadPoolExecutor.java and  ReadFutureTask.java .and kafkaConsumerManager/ConsumerManager use the common  KafkaConsumerThreadPoolExecutor/ReadFutureTask

stanislavkozlovski:
Hey @huangyiminghappy 

No, I didn't mean that.
I think we need to do it similar to this:
```
 @Override
    protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
if (runnable instanceof RunnableReadTask) {
      return new ReadFutureTask(runnable, value);
}
return super.newTaskFor(runnable, value);
}
```
That is, return a `ReadFutureTask` **only if** we are given a `RunnableReadTask`

huangyiminghappy:
@stanislavkozlovski  thank you review

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -354,6 +358,58 @@ private ConsumerState getConsumerInstance(String group, String instance) {
     ConsumerConnector createConsumer(ConsumerConfig config);
   }
 
+  class ReadFutureTask<V> extends FutureTask<V> {
+
+    private Runnable myTask;
```
hachikuji: Couldn't we use the more specific type `RunnableReadTask` and avoid the cast below?

Also a couple nits: 

1) perhaps a better name is simply `readTask`?
2) could this field be final?

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -354,6 +358,58 @@ private ConsumerState getConsumerInstance(String group, String instance) {
...
+
+    private Callable callable;
```
hachikuji: I can understand the desire to make this more general, but we only need this to wrap `RunnableReadTask`, which implements `Runnable` and not `Callable`.  I think we can simplify a little bit by removing this field and the `newTaskFor(Callable<T> callable)` override in `KafkaConsumerThreadPoolExecutor` below.

huangyiminghappy: thanks

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
huangyiminghappy: good idea,thanks

hachikuji: @huangyiminghappy I think we needed to rebase. I went ahead and did it since you have been waiting so patiently to get this patch in.

PR #504, opened 2018-11-05
CPKAFKA-1865:  Performance degrades drastically after running for extended periods

stanislavkozlovski:
### Problem
The REST Proxy has a memory leak in maintaining one data structure.

Its `consumersByExpiration` priority queue is supposed to track the consumer that is closest to expiry. There is a separate `ExpirationThread` that monitors this structure and deletes a consume if it has expired.

We update a consumer's expiration and add it back to the queue (so it can re-order itself) on every `poll()` call. The root problem is that we do not remove the old consumer instance from the queue. This results in a `consumersByExpiration` structure with huge sizes, since `poll()` is called very often, especially when consuming high volumes of records. (max.poll.records is set to 30 by default) Even worse, the proxy is currently configured to `poll()` repeatedly with a timeout of 50ms until the default consume ('/records` endpoint) request timeout of 1000ms is reached.

Testing locally 4 clients consuming and producing every second, the `consumersByExpiration` structure would reach 5.5 million elements in under five minutes.

This explains the CPU usage and the consumer creation requests blocking for a long time in the escalation ticket.

This problem has always been present in both the v1 and v2 consumer implementation. The reason it popped up now in the v2 implementation is that the recent changes in https://github.com/confluentinc/kafka-rest/pull/476 made more frequent calls to `updateExpiration`

### Solution
Remove the consumer from `consumersByExpiration` prior to adding it when updating its expiry.
The overall state machine should now ensure that `consumersByExpiration` contains 1 entry per consumer in the proxy.

1. Create consumer -> add to `consumersByExpiration`
2. Delete consumer - remove from `consumersByExpiration`
3. Update expiry - remove and add to `consumersByExpiration`

There also existed a race condition where a concurrent read/offset-commit might re-add the consumer to `consumersByExpiration` because of the expiration update. Adding a check to see if the consumer is present from the `consumers` data structure (the source of truth for active consumer instances) should prevent this from happening.

### DIFF
--- a/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -201,7 +201,6 @@ public class ConsumerManager {
 
       synchronized (this) {
         consumers.put(cid, state);
-        this.notifyAll();
       }
       succeeded = true;
       return name;
@@ -355,6 +354,7 @@ public class ConsumerManager {
     if (state == null) {
       throw Errors.consumerInstanceNotFoundException();
     }
+    state.updateExpiration();
     return state;
   }
 
@@ -379,9 +379,9 @@ public class ConsumerManager {
 
     @Override
     public void run() {
-      synchronized (ConsumerManager.this) {
-        try {
-          while (isRunning.get()) {
+      try {
+        while (isRunning.get()) {
+          synchronized (ConsumerManager.this) {
             long now = time.milliseconds();
             Iterator itr = consumers.values().iterator();
             while (itr.hasNext()) {
@@ -397,12 +397,12 @@ public class ConsumerManager {
                 });
               }
             }
-
-            ConsumerManager.this.wait(1000);
           }
-        } catch (InterruptedException e) {
-          // Interrupted by other thread, do nothing to allow this thread to exit
+
+          Thread.sleep(1000);
         }
+      } catch (InterruptedException e) {
+        // Interrupted by other thread, do nothing to allow this thread to exit
       }
       shutdownLatch.countDown();
     }
--- a/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -20,9 +20,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.ArrayList;
 import java.util.Properties;
 import java.util.UUID;
 import java.util.Vector;
@@ -383,22 +383,21 @@ public class ConsumerManager {
         try {
           while (isRunning.get()) {
             long now = time.milliseconds();
-            List<ConsumerState> statesToRemove = new ArrayList<>();
-            for (ConsumerState state: consumers.values()) {
+            Iterator itr = consumers.values().iterator();
+            while (itr.hasNext()) {
+              final ConsumerState state = (ConsumerState) itr.next();
               if (state.expired(now)) {
-                statesToRemove.add(state);
+                log.debug("Removing the expired consumer {}", state.getId());
+                itr.remove();
+                executor.submit(new Runnable() {
+                  @Override
+                  public void run() {
+                    state.close();
+                  }
+                });
               }
             }
-            for (ConsumerState staleState: statesToRemove) {
-              log.debug("Removing the expired consumer {}", staleState.getId());
-              final ConsumerState state = consumers.remove(staleState.getId());
-              executor.submit(new Runnable() {
-                @Override
-                public void run() {
-                  state.close();
-                }
-              });
-            }
+
             ConsumerManager.this.wait(1000);
           }
         } catch (InterruptedException e) {
--- a/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -259,7 +259,7 @@ public class ConsumerManager {
           public void onCompletion(
               List<? extends ConsumerRecord<ClientKeyT, ClientValueT>> records, Exception e
           ) {
-            updateExpiration(state);
+            state.updateExpiration();
             if (e != null) {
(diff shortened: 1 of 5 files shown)


gardnervickers:
To elaborate a bit more on why this happens, the `run` method of the `RunnableReadTask` updates consumer state expiration with `KafkaConsumerManager.this.updateExpiration(taskState.consumerState);`. `taskState.consumerState` is captured on the `RunnableReadTask` here.
There is an alternate branch in `run` which calls `delayFor`, `delayFor` resubmits the `RunnableReadTask` in its entirety (with the captured `taskState.consumerState`) after a timeout, calling `run` again. This causes `updateExpiration(taskState.consumerState)` to be called again too, potentially with a duplicate `consumerState`. 

Another possible way to solve this would be to move [`updateExpiration`](https://github.com/confluentinc/kafka-rest/blob/4.1.x/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java#L369) call to the branch where the runnable does not retry, the `else` branch, however that has some semantic differences

src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -403,21 +387,23 @@ public void run() {
         try {
           while (isRunning.get()) {
             long now = time.milliseconds();
-            while (!consumersByExpiration.isEmpty() && consumersByExpiration.peek().expired(now)) {
-              final ConsumerState state = consumersByExpiration.remove();
-              consumers.remove(state.getId());
+            List<ConsumerState> statesToRemove = new ArrayList<>();
+            for (ConsumerState state: consumers.values()) {
```
hachikuji: Probably not a big deal, but we could do this in a single pass using `Iterator.remove()`. 

I'm also debating whether it's worthwhile changing the `consumers` collection to be a `ConcurrentHashMap`. Maybe then we'd only need the lock when we have a consumer to expire? Note that this can be left for a follow up. There probably won't be that many consumers, so holding the lock may be ok.

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -493,15 +481,12 @@ private KafkaConsumerState getConsumerInstance(String group, String instance) {
     return getConsumerInstance(group, instance, false);
   }
 
+  /*
+    Update the expiration of a KafkaConsumerState.
+    Synchronized to avoid race conditions with the ExpirationThread
+   */
   private synchronized void updateExpiration(KafkaConsumerState state) {
```
hachikuji: This method is contributing much anymore. Perhaps we can remove it?

src/main/java/io/confluent/kafkarest/ConsumerManager.java
stanislavkozlovski: I think that holding the lock should okay as well.

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: Yeah I thought about that initially. I wrongly remembered that the `updateExpiration` is stateful by using the `expirationTime` instance variable but it was only setting it by using config variables.

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -473,31 +468,20 @@ public void shutdown() {
...
     ConsumerInstanceId id = new ConsumerInstanceId(group, instance);
-    final KafkaConsumerState state = remove ? consumers.remove(id) : consumers.get(id);
+    final KafkaConsumerState state = toRemove ? consumers.remove(id) : consumers.get(id);
     if (state == null) {
       throw Errors.consumerInstanceNotFoundException();
     }
-    // Clear from the timeout queue immediately so it isn't removed during the read operation,
-    // but don't update the timeout until we finish the read since that can significantly affect
-    // the timeout.
-    consumersByExpiration.remove(state);
```
hachikuji: The effect of this previously was to prevent expiration until the request had been handled. Should we have some additional logic to accomplish the same thing? Maybe it's good enough if we update the expiration time both when a request is received and when the response is sent?

src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -202,7 +201,6 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
 
       synchronized (this) {
         consumers.put(cid, state);
-        consumersByExpiration.add(state);
         this.notifyAll();
```
dhruvilshah3: Do we need this with the changed expiration thread logic?

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -519,21 +503,22 @@ public void run() {
...
+                });
+              }
             }
-            long timeout =
-                consumersByExpiration.isEmpty()
-                ? Long.MAX_VALUE
-                : consumersByExpiration.peek().untilExpiration(now);
-            KafkaConsumerManager.this.wait(timeout);
+
+            KafkaConsumerManager.this.wait(1000);
```
dhruvilshah3: Could we use `sleep` instead now that we don't require the expiration thread to be woken up explicitly?

src/main/java/io/confluent/kafkarest/ConsumerManager.java
huangyiminghappy: @dhruvilshah3  could you help me review the https://github.com/confluentinc/kafka-rest/pull/500? thank you

src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: Oh I see... yeah we could update the expiry here as well

hachikuji: I think the two cases are offset commits and reads. Both have some logic to update the expiration time after completion, so that's probably good enough.

PR #507, opened 2018-11-07
CPKAFKA-1871: Fix problem where ReadTaskSchedulerThread would loop too often

stanislavkozlovski:
The problem originates from the `getDelay` implementation here:
https://github.com/confluentinc/kafka-rest/blob/2928ac85ae77d2350c774271c5c6d4ae0d34c317/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java#L389

We want each read task to be delayed for at least 50 milliseconds but the conversion done there takes the 50 milliseconds leftover as nanoseconds and converts them to milliseconds (returning 0). This causes the ReadTaskSchedulerThread to use more CPU than needed by looping read tasks and calling `poll()` too often


stanislavkozlovski:
Yeah. I'll make a note in CPKAFKA-1868

PR #508, opened 2018-11-08
CPKAFKA-1868: Improve consumer tests

stanislavkozlovski:
This PR adds:
* tests verifying that the `DelayQueue<RunnableReadTask> delayedReadTasks` is populated
* tests verifying that `RunnableReadTask` delays are updated
* tests verifying that `backoff.ms` affects task re-schedules
* a bigger, integration-like load tests where we spawn 50 consumers and run for 1 minute

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
@@ -386,7 +386,6 @@ public class ConsumerManagerTest {
       public void onCompletion(List<? extends ConsumerRecord<byte[], byte[]>> records,
                                RestException e) {
         sawCallback = true;
-        actualRecords = records;
         actualException = e;
       }
     });
@@ -394,16 +393,19 @@ public class ConsumerManagerTest {
     // backoff is 250
     Thread.sleep(100);
     // backoff should be in place right now. the read task should be delayed and re-ran until the max.bytes or timeout is hit
-    System.out.println(consumerManager.delayedReadTasks.size());
+    assertEquals(1, consumerManager.delayedReadTasks.size());
     Thread.sleep(100);
     assertEquals(1, consumerManager.delayedReadTasks.size());
     f.get();
+    assertTrue("Callback not called", sawCallback);
+    assertNull("Callback exception should not be populated", actualException);
   }
 
   @Test
   public void testBackoffMsUpdatesReadTaskExpiry() throws Exception {
     Properties props = new Properties();
-    props.setProperty(KafkaRestConfig.CONSUMER_ITERATOR_BACKOFF_MS_CONFIG, "1000");
+    int backoffMs = 1000;
+    props.setProperty(KafkaRestConfig.CONSUMER_ITERATOR_BACKOFF_MS_CONFIG, Integer.toString(backoffMs));
     // This setting supports the testConsumerOverrides test. It is otherwise benign and should
     // not affect other tests.
     props.setProperty("exclude.internal.topics", "false");
@@ -432,8 +434,8 @@ public class ConsumerManagerTest {
       fail("Could not get read task in time. It should not be null");
     }
     long delay = readTask.getDelay(TimeUnit.MILLISECONDS);
-    assertTrue(delay < 1000);
-    assertTrue(delay > 700);
+    assertTrue(delay < backoffMs);
+    assertTrue(delay > (backoffMs * 0.5));
     f.get();
   }
 
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
@@ -437,6 +437,44 @@ public class ConsumerManagerTest {
     f.get();
   }
 
+  @Test
+  public void testConsumerExpirationIsUpdated() throws Exception {
+    expectCreateNoData();
+    EasyMock.expect(mdObserver.topicExists(topicName)).andReturn(true);
+    EasyMock.replay(mdObserver, consumerFactory);
+    String cid = consumerManager.createConsumer(
+            groupName, new ConsumerInstanceConfig(EmbeddedFormat.BINARY));
+    ConsumerState state = consumerManager.getConsumerInstance(groupName, cid);
+    long initialExpiration = state.expiration;
+
+    Future f = readTopicFuture(cid, topicName, Long.MAX_VALUE, new ConsumerReadCallback<byte[], byte[]>() {
+      @Override
+      public void onCompletion(List<? extends ConsumerRecord<byte[], byte[]>> records,
+                               RestException e) {
+        sawCallback = true;
+        actualRecords = records;
+        actualException = e;
+      }
+    });
+    Thread.sleep(100);
+    assertTrue(state.expiration > initialExpiration);
+    initialExpiration = state.expiration;
+
+    f.get();
+    assertTrue(state.expiration > initialExpiration);
+
+    initialExpiration = state.expiration;
+    consumerManager.commitOffsets(groupName, cid, new ConsumerManager.CommitCallback() {
+      @Override
+      public void onCompletion(List<TopicPartitionOffset> offsets, Exception e) {
+        sawCallback = true;
+
+        actualException = e;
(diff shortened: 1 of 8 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
```diff
@@ -364,6 +364,119 @@ public void testMultipleTopicSubscriptionsFail() throws Exception {
...
+    expectCreateNoData();
+    EasyMock.expect(mdObserver.topicExists(topicName)).andReturn(true);
+    EasyMock.replay(mdObserver, consumerFactory);
+    String cid = consumerManager.createConsumer(groupName,
+            new ConsumerInstanceConfig(EmbeddedFormat.BINARY));
+    Future f = readTopicFuture(cid, topicName, Long.MAX_VALUE, new ConsumerReadCallback<byte[], byte[]>() {
+      @Override
+      public void onCompletion(List<? extends ConsumerRecord<byte[], byte[]>> records,
+                               RestException e) {
+        sawCallback = true;
```
hachikuji: Unless I'm missing something, we don't appear to do any validation of these fields.

kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
```diff
@@ -364,6 +364,119 @@ public void testMultipleTopicSubscriptionsFail() throws Exception {
...
+    });
+
+    Thread.sleep(100);
+    ConsumerManager.RunnableReadTask readTask = consumerManager.delayedReadTasks.peek();
+    if (readTask == null) {
+      fail("Could not get read task in time. It should not be null");
+    }
+    long delay = readTask.getDelay(TimeUnit.MILLISECONDS);
+    assertTrue(delay < 1000);
+    assertTrue(delay > 700);
```
hachikuji: Where does this value come from? Are there any concerns about this test becoming flaky?

kafka-rest/src/test/java/io/confluent/kafkarest/v2/LoadTest.java
```diff
@@ -0,0 +1,172 @@
...
+            assertNull(actualException);
+            assertNull(actualRecords);
+            assertFalse(sawCallback);
+            schedulePoll();
+            consumerManager.readRecords(consumer.groupName, consumer.cid(), BinaryKafkaConsumerState.class,
+                    -1, Long.MAX_VALUE, callback);
+        }
+
+        void verifyRead() {
+            Assert.assertTrue("Callback failed to fire", sawCallback);
```
hachikuji: nit: static import `assertTrue`?

kafka-rest/src/test/java/io/confluent/kafkarest/v2/LoadTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private Capture<Properties> capturedConsumerConfig;
+
+    private long requestTimeoutMs = 1000;
+
+    class ConsumerTestRun {
+        private final MockConsumer consumer;
+        private volatile boolean sawCallback;
+        private ConsumerReadCallback callback;
+        private RestException actualException;
```
hachikuji: I guess this and `actualRecords` should be volatile also?

kafka-rest/src/test/java/io/confluent/kafkarest/v2/LoadTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private static final String topicName = "testtopic";
+
+    private Capture<Properties> capturedConsumerConfig;
+
+    private long requestTimeoutMs = 1000;
+
+    class ConsumerTestRun {
+        private final MockConsumer consumer;
+        private volatile boolean sawCallback;
```
hachikuji: nit: this field seems redundant since we expect one of `actualRecords` or `actualException` to be set.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/LoadTest.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Properties;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+
+@RunWith(EasyMockRunner.class)
+public class LoadTest {
```
hachikuji: Nice work on this. This seems pretty good for verifying the backoff behavior. I guess the main gap is that we don't have a good way to simulate the interaction with the rebalance protocol. It might also be nice to be able to mix offset commit calls. Anyway, it's a good start.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/LoadTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+        for (ConsumerTestRun run: consumers) {
+            run.bootstrap();
+        }
+
+        for (ConsumerTestRun run: consumers) {
+            run.read();
+        }
+        for (int i = 0; i < 60; i++) {
+            awaitRead();
```
hachikuji: I'm wondering if there's any way we could add some variability here. To simulate real load, the requests would be arriving at different intervals rather than all at once.

kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
stanislavkozlovski: It's possible but I believe it shouldn't be flaky. I'll make the value more clear - it comes from the backoff.ms

kafka-rest/src/test/java/io/confluent/kafkarest/v2/LoadTest.java
stanislavkozlovski: Yeah. I don't think volatile should have much effect here since we're doing things in a serial manner - maybe for the memory barrier it's useful. I added the original in good taste

kafka-rest/src/test/java/io/confluent/kafkarest/v2/LoadTest.java
stanislavkozlovski: I'm not sure how easy it would be to verify the rebalance protocol. The `MockConsumer` does not seem to respect other consumers that have joined the group - you only schedule the records it should return on `poll()`.
We could modify the callback we give it to mimic the rebalances. I can open a JIRA for that if you think it's a reasonable approach

kafka-rest/src/test/java/io/confluent/kafkarest/v2/LoadTest.java
stanislavkozlovski: I think that it helps keep the test code clearer and more succinct. Otherwise I think it would be more code to print out a message saying whether the callback was fired at all or not (check exception, check records and if both null fail)

kafka-rest/src/test/java/io/confluent/kafkarest/unit/ConsumerManagerTest.java
hachikuji: There are a few more with the same problem. Maybe we could just create a function which does the following:
```java
f.get();
assertTrue("Callback not called", sawCallback);
assertNull("Callback exception should not be populated", actualException);
```
Then we can call it at the end of each function instead of `f.get()`.

hachikuji:
One test failure: https://jenkins.confluent.io/job/confluentinc-pr/job/kafka-rest/job/PR-508/5/testReport/io.confluent.kafkarest/ConsumerManagerTest/testConsumerExceptions/.

stanislavkozlovski:
The exception raised is `io.confluent.rest.exceptions.RestServerErrorException: Kafka error: Iterator is in failed state` from the null record provided in `referenceRecords`. I'm not sure if it's possible to even recover that `ConsumerIterator`.

Unless we can find an easy way to fix this, I propose we delete the test, given it has apparently never worked and the `ConsumerIterator` is deprecated itself

hachikuji:
@stanislavkozlovski Ok, if there's no easy way to salvage anything useful from the test, then let's go ahead and remove it. I agree that we'd need more logic to recover from a failed iterator.

stanislavkozlovski:
@hachikuji removed

PR #512, opened 2018-11-20
Fix race condition when expiration thread iterates over partially-initialized consumer

colinhicks:


stanislavkozlovski:
Wow, that's a nasty race condition you caught here.
https://github.com/confluentinc/kafka-rest/blob/31b6deb785264c30a1d7ebb9207dedcbb9811f9d/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java#L184
Is where we can add `null` into the `consumers` map and if the expiry thread catches that before the consumer is created, the expiry thread will shut down!

colinhicks:
Looking at the logic in previous versions, it looks like we should apply this to 4.0.x and up. This is also consistent with CI failure history on the respective jobs.

stanislavkozlovski:
Absolutely right, my bad for not saying. The changes for iterating on `consumers` in the expiry thread was merged in 4.0.x

PR #513, opened 2018-11-28
CPKAFKA-1929: Adjust packaging for Scala 2.12

colinhicks:
See https://github.com/confluentinc/kafka-packaging/pull/90.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -8,3 +8,7 @@ override_dh_systemd_enable:
 
 override_dh_systemd_start:
 	dh_systemd_start --no-start
+
+override_jh_manifest:
+	# Disable jh_manifest target, which is unnecessary and fails against Scala 2.12.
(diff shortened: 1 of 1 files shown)


debian/rules
```diff
@@ -8,3 +8,7 @@ override_dh_systemd_enable:
 
 override_dh_systemd_start:
 	dh_systemd_start --no-start
+
+override_jh_manifest:
+	# Disable jh_manifest target, which is unnecessary and fails against Scala 2.12.
+	# See https://manpages.debian.org/testing/debhelper/dh.1.en.html#OVERRIDE_TARGETS
```
mohnishbasha: Adding some documentation as to why disabling jh_manifest doesnt affect us might be helpful as well, since we set our own class path and doesnt rely on jh_manifest.

PR #515, opened 2018-12-03
Set consumer setting max.poll.interval.ms to rest proxy setting of co…

mmercedes:
…nsumer.timeout.ms

If `consumer.timeout.ms` has been set to a value greater than the default value of `max.poll.interval.ms` and a consumer has set `auto.commit.enable=false` then it is possible the kafka brokers will consider a consumer as failed and release its partition assignments, while the rest proxy maintains a consumer instance handle. This then leads to an exception on the next call to poll, commitSync, or similar.

This commit sets max.poll.interval.ms equal to consumer.timeout.ms to ensure that kafka will not consider the consumer failed until the rest proxy does as well.

Steps to reproduce the issue : 
- Kafka version 0.10.1.0 or higher (where `max.poll.interval.ms` is introduced)
- Set `consumer.timeout.ms` to a value greater than the `max.poll.interval.ms` default of 5 minutes
- Create a consumer via the rest proxy
- Sleep for a time greater than `max.poll.interval.ms` but less than `consumer.timeout.ms`
- Have consumer attempt to read a record or post an offset

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
@@ -185,6 +185,12 @@ public class ConsumerManager {
       // functionality, we always use a timeout. This can't perfectly guarantee a total request
       // timeout, but can get as close as this timeout's value
       props.setProperty("consumer.timeout.ms", Integer.toString(iteratorTimeoutMs));
+      // If the consumer.timeout.ms value is set higher than the default value
+      // for max.poll.interval.ms then it is possible the consumer will be
+      // considered failed by the brokers while it has yet to hit the timeout
+      // for the rest proxy.
+      props.setProperty("max.poll.interval.ms", ((Integer) iteratorTimeoutMs).toString());
+
       if (instanceConfig.getAutoCommitEnable() != null) {
         props.setProperty("auto.commit.enable", instanceConfig.getAutoCommitEnable());
       } else {
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
@@ -216,6 +216,12 @@ public class KafkaConsumerManager {
       // and should not be propagated to the consumer
       props.setProperty("request.timeout.ms", "30000");
 
+      // If the consumer.timeout.ms value is set higher than the default value
+      // for max.poll.interval.ms then it is possible the consumer will be
+      // considered failed by the brokers while it has yet to hit the timeout
+      // for the rest proxy.
+      props.setProperty("max.poll.interval.ms", props.getProperty("consumer.timeout.ms", ""));
+
       props.setProperty(
           "schema.registry.url",
           config.getString(KafkaRestConfig.SCHEMA_REGISTRY_URL_CONFIG)


mmercedes:
[clabot:check]

mmercedes:
anyone able to take a look at this?

bgreenlee:
Bump

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -216,6 +216,12 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
       // and should not be propagated to the consumer
       props.setProperty("request.timeout.ms", "30000");
 
+      // If the consumer.timeout.ms value is set higher than the default value
+      // for max.poll.interval.ms then it is possible the consumer will be
+      // considered failed by the brokers while it has yet to hit the timeout
+      // for the rest proxy.
+      props.setProperty("max.poll.interval.ms", props.getProperty("consumer.timeout.ms", ""));
```
stanislavkozlovski: The consumer should not be considered failed by the brokers if it violates the `max.poll.interval.ms`. The validation is done in the client side (here: https://github.com/apache/kafka/blob/0d55f0f3ec8f97bc250b325481f6f2fa70f52a5c/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java#L1052)

In the case of the v2 consumer, `max.poll.interval.ms` should be close to `request.timeout.ms`. For what it's worth, I don't think we should set this property in the code for the v2 consumer at all. Users could configure it in their proxy config via adding the `consumer.` prefix to it.

kafka-rest/src/main/java/io/confluent/kafkarest/ConsumerManager.java
```diff
@@ -185,6 +185,12 @@ public String createConsumer(String group, ConsumerInstanceConfig instanceConfig
       // functionality, we always use a timeout. This can't perfectly guarantee a total request
       // timeout, but can get as close as this timeout's value
       props.setProperty("consumer.timeout.ms", Integer.toString(iteratorTimeoutMs));
+      // If the consumer.timeout.ms value is set higher than the default value
+      // for max.poll.interval.ms then it is possible the consumer will be
+      // considered failed by the brokers while it has yet to hit the timeout
+      // for the rest proxy.
+      props.setProperty("max.poll.interval.ms", Integer.toString(iteratorTimeoutMs));
```
stanislavkozlovski: The old consumer did not have the `max.poll.interval.ms` config at all.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
stanislavkozlovski: Additionally, `consumer.timeout.ms` is not something we have in any config nor is it documented anywhere in the proxy. It is part of the old consumer configs (https://kafka.apache.org/20/documentation.html#oldconsumerconfigs)

mmercedes:
@stanislavkozlovski yes I have tested the issue locally and it resolved the issue I mentioned in my original comment. You can run these steps to reproduce the issue:

Steps to reproduce the issue :

Kafka version 0.10.1.0 or higher (where max.poll.interval.ms is introduced)
Set consumer.timeout.ms to a value greater than the max.poll.interval.ms default of 5 minutes
Create a consumer via the rest proxy
Sleep for a time greater than max.poll.interval.ms but less than consumer.timeout.ms
Have consumer attempt to read a record or post an offset

PR #527, opened 2019-01-03
Update tools to use API v2

zigarn:


kafka-rest/src/main/java/io/confluent/kafkarest/tools/ProducerPerformance.java
```diff
@@ -104,7 +104,7 @@ protected void doIteration(PerformanceStats.Callback cb) {
       URL url = new URL(targetUrl);
       connection = (HttpURLConnection) url.openConnection();
       connection.setRequestMethod("POST");
-      connection.setRequestProperty("Content-Type", Versions.KAFKA_MOST_SPECIFIC_DEFAULT);
```
cmccabe: @stanislavkozlovski , @ewencp : do you know why KAFKA_MOST_SPECIFIC_DEFAULT is still KAFKA_V1_JSON?

stanislavkozlovski: I do not know why that is. I see that it is used extensively throughout our tests so changing it might not be trivial and require us to update a lot of tests (while also re-introduce ones for V1)

kafka-rest/src/main/java/io/confluent/kafkarest/tools/ConsumerPerformance.java
```diff
@@ -138,8 +152,10 @@ protected void close() {
       URL url = new URL(target);
       connection = (HttpURLConnection) url.openConnection();
       connection.setRequestMethod(method);
-      if (entity != null) {
-        connection.setRequestProperty("Content-Type", Versions.KAFKA_MOST_SPECIFIC_DEFAULT);
+      if (! method.equals("GET")) {
```
stanislavkozlovski: nit: extra space between `!` and `method.equals()`

zigarn:
I took a look for changing `Versions.KAFKA_MOST_SPECIFIC_DEFAULT` to V2 but it's clearly not trivial as the default use of the API is to use V1 and the client need to set the content-type to V2 to use V2 API.

cmccabe:
LGTM.  Let's consider a follow-up change to rename MOST_SPECIFIC_DEFAULT so that it's not so confusing.

stanislavkozlovski:
@cmccabe I'll sync with @ewencp offline on `KAFKA_MOST_SPECIFIC_DEFAULT`.

Thanks for the PR @zigarn! Merging into master.

kafka-rest/src/main/java/io/confluent/kafkarest/tools/ProducerPerformance.java
ewencp: it's just because that's all that existed when this was written and I don't think updating this got scoped into the v2 effort

PR #529, opened 2019-01-18
Explain `fetch.min.bytes`

mente:
Another follow up from #481 , this time trying to explain minimal amount of bytes to fetch from consumer

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -767,11 +767,10 @@ any consumers before it is terminated.
                          defaults to "binary".
    :<json string auto.offset.reset: Sets the ``auto.offset.reset`` setting for the consumer
    :<json string auto.commit.enable: Sets the ``auto.commit.enable`` setting for the consumer
-   :<json string fetch.min.bytes: Sets the ``fetch.min.bytes``
-                                                    setting for this consumer specifically
+   :<json string fetch.min.bytes: Minimal amount of bytes to return before timeout has been reached. This setting allows consumer to return before timeout is reached. Default is `-1`
    :<json string consumer.request.timeout.ms: Sets the ``consumer.request.timeout.ms`` setting for this consumer specifically.
                                               This setting controls the maximum total time to wait for messages for a request
-                                              if the maximum request size has not yet been reached.
+                                              if the maximum request size has not yet been reached. If you want to return earlier than that if a given byte threshold is reached, configure `fetch.min.bytes`.
                                               It does not affect the underlying consumer->broker connection. Default value is taken from the REST proxy config file  
 
    :>json string instance_id: Unique ID for the consumer instance in this group.


mente:
And follow up question: isn't `fetch.min.bytes` has the same functionality as `max_bytes` in `/consumer/..../records` endpoint?

mente:
@stanislavkozlovski maybe you can take a look here?

stanislavkozlovski:
> And follow up question: doesn't fetch.min.bytes have the same functionality as max_bytes in /consumer/..../records endpoint?
In a sense, yes it does. The moment `max_bytes` is reached - the proxy returns a response. We will always return less than `max_bytes` in a response.
For `fetch.min.bytes`, we will not poll to Kafka the moment it is reached but we will return all the already-received from the previous `Consumer#poll()` calls.

The code for this logic resides here: https://github.com/confluentinc/kafka-rest/blob/5e8aad6d767a15be5b54b52904e8384a02786e2f/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java#L140

docs/api.rst
```diff
@@ -767,11 +767,10 @@ any consumers before it is terminated.
                          defaults to "binary".
    :<json string auto.offset.reset: Sets the ``auto.offset.reset`` setting for the consumer
    :<json string auto.commit.enable: Sets the ``auto.commit.enable`` setting for the consumer
-   :<json string fetch.min.bytes: Sets the ``fetch.min.bytes``
-                                                    setting for this consumer specifically
+   :<json string fetch.min.bytes: Minimal amount of bytes to return before timeout has been reached. This setting allows consumer to return before timeout is reached. Default is `-1`
```
stanislavkozlovski: It is probably useful to define the setting here as well, but it is worth it to mention that this config is already documented in https://docs.confluent.io/current/kafka-rest/docs/config.html

I think that at the very least we should contain 
```
Sets the ``fetch.min.bytes`` proxy setting for this consumer specifically
```

stanislavkozlovski: No, what I meant was that the 
```
Sets the ``fetch.min.bytes`` proxy setting for this consumer specifically
```
is sufficient documentation, because `fetch.min.bytes` is explained here -> https://docs.confluent.io/current/kafka-rest/docs/config.html

I think that your addition of
```
If you want to return earlier than that if a given byte threshold is reached, configure `fetch.min.bytes`.
```
should be in the `docs/config.rst` part of `consumer.request.timeout.ms`.

`docs/config.rst` is the authoritative source of truth for a configuration's behavior. I don't know if it's a good idea to duplicate that explanation in the `api.rst` docs

stanislavkozlovski: @mente I believe you might have missed my point here.
The changes you made currently duplicate our definition of `fetch.min.byte`, which I think is unnecessary and will create extra burden when updating the config. Could you re-read my previous comment and share your thoughts?

mente:
@stanislavkozlovski can not find your comment anymore. I do understand your point about confusion but on the other side documentation should be consumer-oriented and from my side it brings another confusion in another way :)

docs/api.rst
mente: Point is that you need to connect the dots between the server configuration and REST api. Because in the big companies (IMHO where kafka is common solution) you might have different people setting up the proxy (admins, devops, bla-bla) and using the APIs of the server (developers, data engineers, etc.)

AndrewJSchofield:
Tidying up dormant PRs. This config has an explanation in the current codebase, roughly matching this PR.

PR #534, opened 2019-02-15
Add "the License" ref

maxzheng:


maxzheng:
@gwenshap Added the license to the missing files.

For agreement vs license, I believe they are distinct. If you search for "the License" and `the “License”` in the CCL Agreement text, that is distinctly called out as a separate thing from the agreement. The agreement grants the license -- see section 1.1 in https://www.confluent.io/confluent-community-license

In these source code, it is not referring to the agreement that grants the license, but the license itself that the source code is under. If you search for "the License", there are several references to the license, and hence it's good to add "the License" to provide the reference mapping from Confluent Community License to "the License". 

Let me know if we want to follow up with legal.

gwenshap:
Thank you and LGTM.

PR #546, opened 2019-03-15
CPKAFKA-2590: Bump ZK connection timeout to avoid test flakes.

gardnervickers:
Also addresses CPKAFKA-2591.

This PR just bumps the timeout for connecting to Zookeeper in ClusterTestHarness. There seemed to be multiple tests flaking due to:
```
kafka.zookeeper.ZooKeeperClientTimeoutException: Timed out waiting for connection while in state: CONNECTING
```

I ran the tests 20 times locally and could not reproduce the above error.


stanislavkozlovski:
Could you reproduce it before bumping the timeout?

PR #556, opened 2019-04-13
CPKAFKA-2757 : Set error codes for kafka authorization and authentication errors

omkreddy:


rajinisivaram:
@omkreddy Have we considered creating a shared exception mapper that REST proxy as well as other components (C3, MDS etc.) will use?

omkreddy:
@rajinisivaram C3 exception mappers currently handles only [ExecutionException](https://github.com/confluentinc/blueway/blob/master/control-center/src/main/java/io/confluent/controlcenter/rest/KafkaExceptionMapper.java#L37) (AdminClient errors) and returns[ custom response entity](https://github.com/confluentinc/blueway/blob/master/control-center/src/main/java/io/confluent/controlcenter/rest/KafkaExceptionMapper.java#L144), without any custom error code. Kafka-rest does not have custom exception mapper and returns standard rest utils [ErrorMessage](https://github.com/confluentinc/rest-utils/blob/master/core/src/main/java/io/confluent/rest/entities/ErrorMessage.java#L24) with custom [error code](https://github.com/confluentinc/kafka-rest/blob/master/kafka-rest-common/src/main/java/io/confluent/kafkarest/Errors.java#L28). 

I am not sure we can have common exception mapper which can be used by all rest utils projects.
Is this common exception mapper is only for Kafka related exceptions? MDS may not have kafka related errors.

 One option is, we can have common Exception -> (Response code , error_code) mapping, and have individual projects write their own mappers/custom handling.

rajinisivaram:
@omkreddy Thanks, I thought @norwood mentioned on Slack that C3 could be updated to use the same exception mapper as the REST proxy if we had a shared exception mapper. Apart from Kafka exceptions, the shared mapper would also handle authentication and authorization exceptions consistently across all components. Agree that if we can't use a shared mapper then we could map to error codes as you suggested and have different components implement their own custom handlers.

kafka-rest/src/test/java/io/confluent/kafkarest/unit/PartitionsResourceBinaryProduceTest.java
```diff
@@ -255,4 +266,25 @@ public void testProduceInvalidRequest() {
...
+  @Test
+  public void testProduceToPartitionAuthorizationErrors() {
+    for (TestUtils.RequestMediaType mediatype : TestUtils.V1_ACCEPT_MEDIATYPES) {
+      for (String requestMediatype : TestUtils.V1_REQUEST_ENTITY_TYPES_BINARY) {
+        Response
+            rawResponse =
+            produceToPartition(topicName, 0, mediatype.header, requestMediatype,
+                EmbeddedFormat.BINARY,
+                produceRecordsOnlyValues, produceKafkaAuthorizationExceptionResults);
+        assertOKResponse(rawResponse, mediatype.expected);
```
stanislavkozlovski: Shouldn't this result in a 401?

kafka-rest/src/main/java/io/confluent/kafkarest/Utils.java
```diff
@@ -43,4 +53,44 @@ public static int errorCodeFromProducerException(Throwable e) {
...
+      return Errors.authenticationException(exception.getMessage());
+    } else if (exception instanceof AuthorizationException) {
+      return Errors.authorizationException(exception.getMessage());
+    } else {
+      throw Errors.kafkaErrorException(exception);
+    }
+  }
+
+  public static RestException convertConsumerException(Exception exception) {
+    if (exception != null && !(exception instanceof RestException)) {
```
stanislavkozlovski: Does this null check make sense? We expect it to be non-null anyway since we have a cast below on the return statement

kafka-rest/src/main/java/io/confluent/kafkarest/Utils.java
```diff
@@ -43,4 +53,44 @@ public static int errorCodeFromProducerException(Throwable e) {
     }
   }
 
+  public static RestException convertProducerException(KafkaException exception) {
+    if (StringUtil.startsWithIgnoreCase(exception.getMessage(), "Invalid partition")) {
+      throw Errors.partitionNotFoundException();
```
stanislavkozlovski: Should we be consistent and just return the error here? Same as in the `throw Errors.kafkaErrorException(exception);` line

kafka-rest/src/main/java/io/confluent/kafkarest/Utils.java
```diff
@@ -43,4 +53,44 @@ public static int errorCodeFromProducerException(Throwable e) {
...
+    } else if (exception instanceof AuthenticationException) {
+      return Errors.authenticationException(exception.getMessage());
```
stanislavkozlovski: Maybe it wouldn't hurt to reduce the code duplication in between these three methods and have a single method to check instances and return `Errors.authenticationException`, `Errors.authorizationException` or `Errors.kafkaErrorException`.

kafka-rest/src/test/java/io/confluent/kafkarest/unit/AbstractConsumerResourceTest.java
```diff
@@ -34,7 +35,6 @@
 import io.confluent.kafkarest.KafkaRestApplication;
 import io.confluent.kafkarest.KafkaRestConfig;
 import io.confluent.kafkarest.KafkaRestContext;
-import io.confluent.kafkarest.Errors;
```
stanislavkozlovski: Could we add a simple test to ensure these changes work with the consumer path?

kafka-rest/src/test/java/io/confluent/kafkarest/unit/PartitionsResourceBinaryProduceTest.java
omkreddy: Updated the code to set Response status code  for  authz failures

kafka-rest/src/test/java/io/confluent/kafkarest/unit/AbstractConsumerResourceTest.java
omkreddy: Added a integration test for consumer.

omkreddy:
@stanislavkozlovski  Thanks for the review. I have addressed the review comments.
There was ask for implementing common KafkaExceptionMapper for handling kafka related exceptions, which  can be used in other projects.  I have raised another PR https://github.com/confluentinc/kafka-rest/pull/558 for mapper. I have included  this PR commits to https://github.com/confluentinc/kafka-rest/pull/558.  I will close this PR and will continue discussion in  https://github.com/confluentinc/kafka-rest/pull/558.

omkreddy:
Closing this PR in favor of  #558.

PR #559, opened 2019-04-23
Add support to override SR properties and make few tests extensible

omkreddy:
- Add support to override SR properties in ClusterTestHarness
- Pass Query params to AbstractProducerTest methods
- Make AvroProducerTest test methods accessible in subclasses


omkreddy:
@stanislavkozlovski Thanks for the review. merging to master

PR #563, opened 2019-05-16
CPKAFKA-2810: Update file descriptor limit to 1 million

vikasconfluent:
Kafka rest proxy opens a lot of files while running. And systemd ignores the
open file limit in /etc/security/limits.conf.

So this change updates kafka-rest.service systemd service file that we
ship to increase open file descriptor limit by adding this line in
[Service] section:

LimitNOFILE=1000000

This sets both hard and soft limit for "the maximum number of open file
descriptors" the program will use. NOTE: If we want to set different
value for hard and soft limits, the syntax is "soft:hard".

### DIFF
--- a/debian/confluent-kafka-rest.service
+++ b/debian/confluent-kafka-rest.service
@@ -9,7 +9,7 @@ User=cp-kafka-rest
 Group=confluent
 Environment="LOG_DIR=/var/log/confluent/kafka-rest"
 ExecStart=/usr/bin/kafka-rest-start /etc/kafka-rest/kafka-rest.properties
-LimitNOFILE=1000000
+LimitNOFILE=100000
 TimeoutStopSec=180
 Restart=no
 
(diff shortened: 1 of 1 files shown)


vikasconfluent:
This is same as commit made on kafka-packaging for kafka service: https://github.com/confluentinc/kafka-packaging/commit/92e7602e39bb40092d2671c5957cdfd742fbfd2b

edenhill:
> Kafka rest proxy opens a lot of files while running

By that you mean file descriptors as in tcp sockets, right? Or does it open actual files too?

vikasconfluent:
> > Kafka rest proxy opens a lot of files while running
> 
> By that you mean file descriptors as in tcp sockets, right? Or does it open actual files too?

Yes, its socket. Copied the commit message from original kafka commit but missed updating this part. Addison mentioned that there are customers hitting limits with proxy too, hence this change. 

Will update commit message to mention open socket count.

edenhill:
Okay, but 1M open sockets seems kind of extreme.
Unless there is a recommendation somewhere for that high of a value I'd rather see it reduced to 100k or so, which seems more reasonable.

vikasconfluent:
Hi Magnus,

Lowered it to 100K and pushed another commit. There is no reasonable upper limit. What we discussed here was that bump it up to some number so that we don't have to see these errors again. 100K seems reasonable too.

For the Kafka server though, we have seen open fd count of 800K or so, so for that I have changed the limit to 1 million.

PR #566, opened 2019-06-10
ST-1704 and ST-1705: Fix usage of variable

ewencp:
Due to extra spaces, the variable is being interpreted as a command and
the variable is never assigned. This causes later logic to always treat
this as a hotfix version.

### DIFF
--- a/create_spec.sh
+++ b/create_spec.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # create_spec.sh spec.in out.spec
-CONFLUENT_PATCH = "cp"
+CONFLUENT_PATCH="cp"
 set -e
 
 echo "Creating the spec file"


ewencp:
Without this, at least 3.3.x and 4.0.x snapshot builds are failing on package version checks, but source of issue is much earlier in build:

```
./create_spec.sh confluent-kafka-rest.spec.in confluent-kafka-rest-4.0.4/confluent-kafka-rest.spec^M
./create_spec.sh: line 4: CONFLUENT_PATCH: command not found^M
```

This causes it to go into the logic for the hotfix builds even on SNAPSHOT builds.

But also @xli1996, why is this done for kafka-rest but not, e.g., for schema-registry or other packages?

xli1996:
Since I found that some 
* new repos that do not have independent `rpm` branch have different logic; 
* we may change new names for `hotfix` in future; 
* I want to avoid breaking some current packaging, I switch to use my personal fork for that special hotfix. 

Apparently, I should revert some changes I made previous.

ewencp:
@xli1996 So what do we want to do here? Some new repos (or just newer versions of existing repos) won't have independent `rpm` branch that they need updates to, but the logic just changed location, didn't it? e.g. compare `rpm` branch `Makefile` logic https://github.com/confluentinc/schema-registry/blob/rpm/Makefile#L113 vs `master` logic https://github.com/confluentinc/schema-registry/blob/master/debian/Makefile -- some of the code just got reorganized into the `Makefile`.

The logic you added isn't necessarily wrong, and I get possibility of changing, but I was confused because schema-registry and kafka-rest are the same age, I believe had migration from separate packaging branch to in-tree packging code at the same time, etc.

ewencp:
Closing in favor of just reverting the last commit -- that update was made accidentally and is why the different repos diverged. We need to settle on our hotfix naming and then we can file PRs against all the repos that need this adjustment, and include the spacing fix on the variable name.

PR #579, opened 2019-07-31
Move ZkExceptionMapper to the kafka-rest-scala-consumer module

ijuma:
`kafka-rest-common` does not have a direct dependency on `zkclient` and
we only need the mapper when using the legacy scala consumer.

The code compiles currently because `commons-utils` has a `zkclient`
dependency, but we're in the process of eliminating that.


omkreddy:
@ijuma Thanks for the PR. LGTM.

PR #580, opened 2019-08-05
MINOR: Include `zkclient` dependency to kafka-rest-scala-consumer shaded jar

omkreddy:
We have removed `zkclient` dependency from Kafka.  kafka-rest-scala-consumer depends on zk-client. This PR includes `zkclient` dependency to kafka-rest-scala-consumer shade jar.


omkreddy:
Thanks for the review, merging to master..

PR #587, opened 2019-09-12
Add API to get beginning offset, end offset and offset for time.

rigelbm:
CPKAFKA-1526: Extend the topic partition resource to support getting
offset information. The equivalent Java APIs are KafkaConsumer
beginningOffsets, endOffsets and offsetForTimes.

The kafka-rest server will create a randomly generated consumer-group
and consumer to use to query the offsets information. It is possible for
an adversarial user to kill this consumer manually. There are some
measures in place for self-healing if that happens.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetResponse.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetResponse.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2019 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.entities;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonProperty.Access;
+
+/**
+ * Response for GET /topics/(topic)/partitions/(partition)/offsets requests.
+ *
+ * @see io.confluent.kafkarest.resources.v2.PartitionsResource#getOffsets(String, int)
+ */
+public final class TopicPartitionOffsetResponse {
+
+  private final long beginningOffset;
+  private final long endOffset;
+
+  public TopicPartitionOffsetResponse(long beginningOffset, long endOffset) {
+    this.beginningOffset = beginningOffset;
+    this.endOffset = endOffset;
+  }
+
+  /**
+   * The earliest offset in the topic partition.
+   */
+  @JsonProperty(value = "beginning_offset", access = Access.READ_ONLY)
+  public long getBeginningOffset() {
+    return beginningOffset;
+  }
+
+  /**
+   * The latest offset in the topic partition.
+   */
+  @JsonProperty(value = "end_offset", access = Access.READ_ONLY)
+  public long getEndOffset() {
+    return endOffset;
+  }
+}
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/extension/InstantConverterProvider.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/extension/InstantConverterProvider.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2019 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.extension;
+
+import static java.util.Objects.requireNonNull;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import javax.annotation.Nullable;
+import javax.ws.rs.ext.ParamConverter;
+import javax.ws.rs.ext.ParamConverterProvider;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * A {@link ParamConverterProvider} for {@link InstantConverter}.
+ *
+ * <p>By default it uses {@link DateTimeFormatter#ISO_INSTANT} format.</p>
+ *
+ * <p>TODO: If more formats are needed, create an annotation, e.g. DateTimeParam, and an
+ * enum, e.g. DateTimeFormat, and map from enum value to DateTimeFormatter. The annotation will be
+ * passed in {@link #getConverter(Class, Type, Annotation[])}.</p>
+ */
+@Provider
+public final class InstantConverterProvider implements ParamConverterProvider {
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public <T> ParamConverter<T> getConverter(
+      Class<T> rawType, Type genericType, Annotation[] annotations) {
+    if (!rawType.equals(Instant.class)) {
+      return null;
+    }
+    return (ParamConverter<T>) new InstantConverter(DateTimeFormatter.ISO_INSTANT);
+  }
+
+  /**
+   * A {@link ParamConverter} for {@link Instant}.
+   */
+  private static final class InstantConverter implements ParamConverter<Instant> {
+
+    private final DateTimeFormatter formatter;
+
+    private InstantConverter(DateTimeFormatter formatter) {
+      this.formatter = requireNonNull(formatter, "formatter");
+    }
+
+    @Override
+    @Nullable
+    public Instant fromString(@Nullable String value) {
+      // Contrary to JAX-RS specification, if the parameter is not present in the requested URI
+      // Jersey will try to convert `null' instead of bypassing conversion and injecting `null'
+      // directly. Return `null' here so that it can be passed to the injection point as expected.
+      if (value == null) {
+        return null;
+      }
+      return formatter.parse(value, Instant::from);
+    }
+
+    @Override
+    public String toString(Instant value) {
+      return formatter.format(value);
+    }
+  }
+}
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/PartitionsResource.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/PartitionsResource.java
@@ -15,33 +15,15 @@
 
 package io.confluent.kafkarest.resources;
 
-import io.confluent.kafkarest.Utils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static java.util.Collections.emptyList;
 
-import java.util.List;
-import java.util.Vector;
-
-import javax.validation.Valid;
-import javax.validation.constraints.NotNull;
-import javax.ws.rs.Consumes;
-import javax.ws.rs.DefaultValue;
-import javax.ws.rs.GET;
-import javax.ws.rs.POST;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
-import javax.ws.rs.QueryParam;
-import javax.ws.rs.container.AsyncResponse;
-import javax.ws.rs.container.Suspended;
-import javax.ws.rs.core.Response;
-
+import io.confluent.kafkarest.ConsumerReadCallback;
 import io.confluent.kafkarest.Errors;
 import io.confluent.kafkarest.KafkaRestContext;
 import io.confluent.kafkarest.ProducerPool;
 import io.confluent.kafkarest.RecordMetadataOrException;
+import io.confluent.kafkarest.Utils;
 import io.confluent.kafkarest.Versions;
-import io.confluent.kafkarest.ConsumerReadCallback;
 import io.confluent.kafkarest.entities.AvroProduceRecord;
 import io.confluent.kafkarest.entities.BinaryProduceRecord;
 import io.confluent.kafkarest.entities.ConsumerRecord;
@@ -53,6 +35,27 @@ import io.confluent.kafkarest.entities.PartitionProduceRequest;
 import io.confluent.kafkarest.entities.ProduceRecord;
 import io.confluent.kafkarest.entities.ProduceResponse;
 import io.confluent.rest.annotations.PerformanceMetric;
+import java.time.Instant;
+import java.util.List;
+import java.util.Optional;
+import java.util.Vector;
+import javax.annotation.Nullable;
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.container.AsyncResponse;
+import javax.ws.rs.container.Suspended;
+import javax.ws.rs.core.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Path("/topics/{topic}/partitions")
 @Produces({Versions.KAFKA_V1_JSON_BINARY_WEIGHTED_LOW, Versions.KAFKA_V1_JSON_AVRO_WEIGHTED_LOW,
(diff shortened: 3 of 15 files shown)


kafka-rest-common/src/main/java/io/confluent/kafkarest/Errors.java
```diff
@@ -54,11 +52,21 @@ public static RestException partitionNotFoundException() {
...
+
+    private ConsumerInstanceNotFoundException() {
+      super(CONSUMER_INSTANCE_NOT_FOUND_MESSAGE, CONSUMER_INSTANCE_NOT_FOUND_ERROR_CODE);
+    }
+  }
+
+  /**
+   * @see ConsumerInstanceNotFoundException
+   */
   public static RestException consumerInstanceNotFoundException() {
```
mhowlett: I think this helper method is now just pointless complexity, so feel free to remove it and update other code unless there's some javaism here i'm not aware of.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/GetTopicPartitionOffsetResponse.java
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * <ul>
+ *   <li>a summary containing the beginning and end offsets</li>
+ *   <li>an offset and its associated timestamp</li>
+ *   <li>empty</li>
+ * </ul>
+ *
+ * @see io.confluent.kafkarest.resources.v2.PartitionsResource#getOffsets(String, int, Instant)
+ */
+public final class GetTopicPartitionOffsetResponse {
```
mhowlett: omitting the prefix `Get` would be more in line with the existing convention.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -374,6 +458,51 @@ boolean hasNextCached() {
     return consumerRecords.poll();
   }
 
+  /**
+   * A topic partition offset, and its associated timestamp.
+   */
+  public static final class OffsetTimestamp {
```
mhowlett: given the existence of `org.apache.kafka.clients.consumer.OffsetAndTimestamp`, what's the value in having this class as well?

also, internally, I'd be inclined to leave timestamp as `long` for as long as possible (see another conversation about the association with time being a little bit arbitrary / user defined).

kafka-rest/src/main/java/io/confluent/kafkarest/entities/GetTopicPartitionOffsetResponse.java
```diff
@@ -0,0 +1,143 @@
+/*
+ * Copyright 2018 Confluent Inc.
```
mhowlett: 2019

kafka-rest/src/main/java/io/confluent/kafkarest/entities/GetTopicPartitionOffsetResponse.java
```diff
@@ -0,0 +1,143 @@
...
+   * Returns an empty response.
+   */
+  public static GetTopicPartitionOffsetResponse empty() {
+    return EMPTY;
+  }
+
+  /**
+   * The earliest offset in the topic partition.
+   */
+  @JsonProperty(value = "beginning", access = Access.READ_ONLY)
```
mhowlett: i'm personally inclined towards the more verbose `beginning_offset` and `end_offset`. I agree with the choice of words `beginning` and `end` as they are consistent with the java api (i would choose other words if that wasn't a constraint though).

kafka-rest/src/main/java/io/confluent/kafkarest/entities/GetTopicPartitionOffsetResponse.java
```diff
@@ -0,0 +1,143 @@
...
+  public Optional<Long> getOffset() {
+    return Optional.ofNullable(offset);
+  }
+
+  /**
+   * The timestamp associated with {@link #getOffset()} in the topic partition.
+   */
+  @JsonProperty(value = "timestamp", access = Access.READ_ONLY)
+  @JsonInclude(Include.NON_ABSENT)
+  public Optional<Instant> getTimestamp() {
```
mhowlett: I'm personally inclined towards using `long` for timestamp because: 

>  1. consistency with the java API. 2. the meaning of timestamp in kafka is almost arbitrary, so it's conceivable people will not want a standard interpretation of it. 3. more definitely easy to parse by all client languages (also, it's well under the javascript max int so no problems there). 4. unix time is not obscure, so not really confusing or hard to deal with in any language.

Anyone else feel free to give an opinion / reasons here. I'm happy to accept either way, just stating a preference (which is not strongly held, i'd be easily convinced the other way).

kafka-rest/src/main/java/io/confluent/kafkarest/entities/GetTopicPartitionOffsetResponse.java
mhowlett: same comment elsewhere

kafka-rest/src/main/java/io/confluent/kafkarest/extension/InstantConverterProvider.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.ws.rs.ext.ParamConverter;
+import javax.ws.rs.ext.ParamConverterProvider;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * A {@link ParamConverterProvider} for {@link InstantConverter}.
+ *
+ * <p>By default it uses {@link DateTimeFormatter#ISO_INSTANT} format.</p>
+ *
+ * <p>TODO(rigelbm): If more formats are needed, create an annotation, e.g. DateTimeParam, and an
```
mhowlett: we typically don't leave next to todo's.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
```diff
@@ -15,53 +15,69 @@
...
+public final class PartitionsResource {
   private static final Logger log = LoggerFactory.getLogger(PartitionsResource.class);
 
+  private static final int CONSUMER_REQUEST_MAX_ATTEMPTS = 3;
+
   private final KafkaRestContext ctx;
+  private final String consumerGroup;
+
+  @GuardedBy("this")
+  private String consumerInstance = null;
```
mhowlett: `consumerInstanceId` would be a better name i think

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
```diff
@@ -200,13 +216,124 @@ public void onCompletion(
...
+   *                                                            partition} don't exist.
+   */
+  @GET
+  @Path("/{partition}/offsets")
+  public GetTopicPartitionOffsetResponse getOffsets(
+      @PathParam("topic") String topic,
+      @PathParam("partition") int partition,
+      @QueryParam("timestamp") @Nullable Instant timestamp
+  ) throws Exception {
+    checkTopicExists(topic);
```
mhowlett: these will both result in synchronous calls to the cluster which seems undesirable - i think you want to just rely on errors returned by the offset requests.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
```diff
@@ -200,13 +216,124 @@ public void onCompletion(
...
+    String serverId = ctx.getConfig().getString(KafkaRestConfig.ID_CONFIG);
+    if (serverId.isEmpty()) {
+      return String.format("rest-consumer-group-%s", UUID.randomUUID().toString());
+    } else {
+      return String.format("rest-consumer-group-%s-%s", serverId, UUID.randomUUID().toString());
+    }
   }
 
-  private void checkTopicExists(final String topic) throws Exception {
+  private synchronized String getConsumerInstance() {
```
mhowlett: better to call this `getConsumerInstanceId` i think

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
```diff
@@ -200,13 +216,124 @@ public void onCompletion(
...
+    // Consumers expire when not used for some time. They can also be explicitly deleted by a user
+    // using DELETE /consumers/{consumerGroup}/instances/{consumerInstances}. If the consumer does
```
mhowlett: can consumers created here really be explicitly DELETEd explicitly in this way?

kafka-rest/src/main/java/io/confluent/kafkarest/Utils.java
```diff
@@ -67,4 +67,23 @@ public static int errorCodeFromProducerException(Throwable e) {
     }
     return Response.Status.OK;
   }
+
+  /**
+   * Tries up to {@code maxAttempts} times to get the value from the {@code supplier}, retrying if
```
mhowlett: i haven't investigated use of this fully yet, but it smells a bit bad to me.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
```diff
@@ -15,53 +15,69 @@
...
 @Consumes({Versions.KAFKA_V2_JSON})
-public class PartitionsResource {
-
+public final class PartitionsResource {
   private static final Logger log = LoggerFactory.getLogger(PartitionsResource.class);
 
+  private static final int CONSUMER_REQUEST_MAX_ATTEMPTS = 3;
+
   private final KafkaRestContext ctx;
+  private final String consumerGroup;
```
mhowlett: this is out of place in `PartitionsResource` - i think you don't have the abstractions right.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
mhowlett: also, there's no reason for `PartitionsResource` to explicitly know about/manage specific consumers either, this would be better abstracted away.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
```diff
@@ -200,13 +216,124 @@ public void onCompletion(
...
+  }
+
+  /**
+   * Returns the earliest offset in the {@code topic} {@code partition}.
+   */
+  private long getBeginningOffset(String topic, int partition) {
+    try {
+      return retrying(
+          () -> ctx.getKafkaConsumerManager()
+              .getBeginningOffset(consumerGroup, getConsumerInstance(), topic, partition),
```
mhowlett: (furthering comments elsewhere) I think the back-and-forth with consumer instance id with the consumer manager is structurally more complex than needed and also there's no scope for other users of consumer manager to re-use these instances.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
```diff
@@ -200,13 +216,124 @@ public void onCompletion(
...
+          () -> ctx.getKafkaConsumerManager()
+              .getOffsetForTime(consumerGroup, getConsumerInstance(), topic, partition, timestamp),
+          ConsumerInstanceNotFoundException.class,
+          CONSUMER_REQUEST_MAX_ATTEMPTS);
+    } catch (ConsumerInstanceNotFoundException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  private String createConsumerGroup() {
```
mhowlett: i'm thinking incorporate this in consumer manager / invent the concept of a 'utility consumer' or something if there's no such thing already.

kafka-rest/src/main/java/io/confluent/kafkarest/Utils.java
rigelbm: Can you elaborate please? :)

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
rigelbm: I believe the Java API will block until the topic/partition exists. I think that's why the other actions do it as well. The option is to have a timeout. But since this action is not going to resource intensive, having the check before might make it fail faster then relying on the timeout. Re-open if you don't agree.

mhowlett: ahh right, this is following what other resources do, fair enough.

not good, but also not something to try and address in this PR - belongs in a separate PR that fixes it everywhere.

kafka-rest/src/main/java/io/confluent/kafkarest/Utils.java
mhowlett: the purpose is to repeatedly run code (without any backoff) until an error stops occurring. that seems a very dubious idea to me, and unlikely to justifiable.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -486,6 +494,87 @@ public ConsumerCommittedResponse committed(
...
 
+  /**
+   * Returns the beginning offset of the {@code topic} {@code partition}.
+   */
+  public long getBeginningOffset(String topic, int partition) {
+    log.debug("Beginning offset for topic {} and partition {}.", topic, partition);
+    KafkaConsumerState<?, ?, ?, ?> consumer;
+    try {
+      consumer =
+          retrying(
```
mhowlett: can this really fail?
if it can (i don't think so, but good chance I'm missing something), i would question the need for this an make changes so it can't.
if that's justifiable for some reason, the retrying functionality would be better encapsulated lower down i think.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -486,6 +494,87 @@ public ConsumerCommittedResponse committed(
...
+   * timestamp} in the {@code topic} {@code partition}, or empty if such offset does not exist.
+   */
+  public Optional<OffsetAndTimestamp> getOffsetForTime(
+      String topic, int partition, Instant timestamp) {
+    log.debug("Offset for topic {} and partition {} at timestamp {}.", topic, partition, timestamp);
+    KafkaConsumerState<?, ?, ?, ?> consumer;
+    try {
+      consumer =
+          retrying(
+              () -> getConsumerInstance(getAdminConsumerInstance()),
```
mhowlett: I think `getConsumerState` would be a better name for this and we could consider changing it in this PR ...

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -328,6 +333,79 @@ public ConsumerCommittedResponse committed(ConsumerCommittedRequest request) {
...
+    try {
+      if (consumer == null) {
+        throw new IllegalStateException("KafkaConsumerState has been closed.");
+      }
+
+      Map<TopicPartition, Long> response =
+          consumer.beginningOffsets(singletonList(new TopicPartition(topic, partition)));
+
+      if (response.size() != 1) {
+        throw new IllegalStateException(
```
mhowlett: don't immediately have an idea of a better alternative, but IllegalStateException seems a bit semantically wrong here. ideas?

kafka-rest/src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetResponse.java
```diff
@@ -0,0 +1,143 @@
...
+   * Returns an empty response.
+   */
+  public static TopicPartitionOffsetResponse empty() {
+    return EMPTY;
+  }
+
+  /**
+   * The earliest offset in the topic partition.
+   */
+  @JsonProperty(value = "beginning", access = Access.READ_ONLY)
```
mhowlett: you don't like `beginning_offset` / `end_offset`?

kafka-rest/src/test/java/io/confluent/kafkarest/v2/PartitionsResourceTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+    String response =
+        target("/topics/{topic}/partitions/{partition}/offsets")
+            .resolveTemplate("topic", TOPIC)
+            .resolveTemplate("partition", PARTITION)
+            .queryParam("timestamp", DATE_TIME_FORMATTER.format(REQUEST_TIMESTAMP_FOR_OFFSET))
+            .request()
+            .get(String.class);
+
+    assertEquals("{}", response);
```
mhowlett: this is an interesting case... I think `null` is semantically more correct. typically a query would return an array, and if nothing matched you'd return an empty array. here you return a value and if nothing matched you're returning a value of  a different form. as an alternative to using null, you could enclose the value in either an array or another object.

don't have strong conviction on the best thing to do here. do lean towards the `null` idea based on everything I can think of.

rigelbm: Done.

mhowlett: i mean `assertEquals("null", response)` - i think this is worse.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/TopicPartitionOffsetResponse.java
mhowlett: nice. you don't have to agree, but glad you did

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
mhowlett: sounds good from the users point of view, so good enough for me.

mhowlett:
i'm happy with this now once we agree on `{\"offset\":null,\"timestamp\":null}` vs `null` and assuming the tests pass

rigelbm:
> i'm happy with this now once we agree on `{\"offset\":null,\"timestamp\":null}` vs `null` and assuming the tests pass

As discussed offline, I moved the offsetForTime to the consumeFromPartition endpoint. Now we don't have this issue anymore. No offset returns empty list.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/PartitionsResource.java
```diff
@@ -101,44 +104,71 @@ public Partition getPartition(
...
-      final @PathParam("topic") String topicName,
-      final @PathParam("partition") int partitionId,
-      final @QueryParam("offset") long offset,
-      final @QueryParam("count") @DefaultValue("1") long count
+      @Suspended AsyncResponse asyncResponse,
+      @PathParam("topic") String topicName,
+      @PathParam("partition") int partitionId,
+      @QueryParam("offset") @Nullable Long offset,
+      @QueryParam("timestamp") @Nullable Instant timestamp,
+      @QueryParam("count") @DefaultValue("1") long count
```
mhowlett: nice. great idea.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/PartitionsResource.java
```diff
@@ -200,13 +197,56 @@ public void onCompletion(
     );
   }
 
-  private boolean topicExists(final String topic)  throws Exception {
-    return ctx.getAdminClientWrapper().topicExists(topic);
+  /**
+   * Get offsets for the given {@code topic} and {@code partition}.
+   *
+   * <p>If {@code timestamp} is not specified, returns a summary with beginning and end offsets.
```
mhowlett: out of date comment - timestamp can no longer be specified.

rigelbm: Done

PR #588, opened 2019-09-26
Updating slack channel for Jenkins build alerts

manveer:
As per the new Slack changes we want all the notifications from bots
which are not immediately actionable to go to a different channel. Moving
the jenkins build alerts to #kafka-warn.


ijuma:
I missed it the first time around, but we need this in older branches too.

PR #617, opened 2020-02-14
Fix compile error due to KIP-515 (ZK TLS)

rondagostino:
This fixes a compile error that was created when [KIP-515 Enable ZK client to use the new TLS supported authentication](https://cwiki.apache.org/confluence/display/KAFKA/KIP-515%3A+Enable+ZK+client+to+use+the+new+TLS+supported+authentication) changed the name of a method that was not part of the public Kafka clients API.

This PR does not address the fact that REST Proxy will be unable to connect to a TLS-enabled ZooKeeper instance -- that will have to be fixed separately.


rigelbm:
Re-running Jenkins build. I think it broke because of the artifactory migration.

PR #622, opened 2020-02-19
Refactoring: Split V1 and V2 resources and entities.

rigelbm:
Currently the project is structured in the following way:

```
- kafkarest: Classes in this packages are mostly controllers, in addition to the server entrypoint and configurations.
      - entities: POJOs that are used for (1) JSON deserialization/serialization of HTTP request/response bodies of both V1 and V2 APIs, and (2) as data holders for the controllers/business logic.
      - resources: The presentation layer. Classes in this package handle either V1 APIs only, or both V1 and V2 APIs.
            - v2: Classes in this package handle V2 APIs only.
      - v2: Controllers that handle only V2 API business logic.
```

There are a couple of issues with the aforementioned structure, which I try to address in this PR:

- The same (HTTP) serialization format is used for both V1 and V2. Changes to one API inevitably change the other.
- The same resource handles both V1 and V2. Changes to one implementation will after the other.
- The controllers use the same POJO as the presentation. That makes reuse by another API (e.g. V3) harder, as changes to the controller might alter both V1 and V2 API surface.

This PR basically splits the resources and entities in V1 and V2, so that only the controllers, where the business logic resides, is shared between them. This is accomplished with the following changes:

```
- For each resource kafkarest.resources.FooResource:
      - If it handles V1 API, copy it to kafkarest.resources.v1.FooResource.
      - If it handles V2 API, copy it to kafkarest.resources.v2.FooResource.
      - Remove kafkarest.resources.FooResource.
- For each entity kafkarest.entities.Foo:
      - If it is part of a request/response of a V1 API, copy it to kafkarest.entities.v1.Foo.
      - If it is part of a request/response of a V2 API, copy it to kafkarest.entities.v2.Foo.
      - If it is not used by a controller, remove kafkarest.entities.Foo.
```

As part of the refactoring, all entities were made immutable and had `equals`/`hashCode`/`toString` generated.

There's one downside of the refactoring which is that some code duplication was introduced. I wanted to make the changes obviously correct to easy review. Duplicating an existing piece of code is obviously correct, whereas extracting out common code is not. As part of V3 API addition, I will push as much common code down to the controller layer. I kindly ask the reviewer to overlook for now any such duplication.

Last but not least, please do not get scared with the size of the PR. Git is not very good at tracking file copy/move (compared to mercurial). A lot of the additions are actually just copies. I've only created 3 new classes (`ResourcesFeature`, `V1ResourcesFeature` and `V2ResourceFeature`). Then a whole lot of other additions are just the auto-generated `equals`/`hashCode`/`toString`. Other than that, this PR does not change any behaviour. Despite the size of the PR, there's actually very little to review :).

Thanks for reviewing, and sorry for the trouble.


kafka-rest-common/src/main/java/io/confluent/kafkarest/entities/ConsumerRecord.java
```diff
@@ -117,13 +77,44 @@ public boolean equals(Object o) {
     return partition == that.partition
            && offset == that.offset
            && Objects.equals(topic, that.topic)
-           && Objects.equals(key, that.key)
-           && Objects.equals(value, that.value);
+           && genericEquals(key, that.key)
```
mhowlett: good catch(es)

mhowlett:
thanks for starting out by taking the effort to clean up, long overdue - the future is bright.

duplicating code is underrated - good idea.

this is probably harder to review than write I think, but lgtm afaict and the vision looks good.

PR #624, opened 2020-02-21
Add first V3 cluster APIs (+ V3 plumbing).

rigelbm:
The APIs added are:

* GET /clusters -> returns the list of clusters known
* GET /clusters/{clusterId} -> returns the cluster with the given clusterId

### DIFF
--- a/kafka-rest-common/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest-common/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -65,6 +65,13 @@ public class KafkaRestConfig extends RestConfig {
       + " hostname is used";
   public static final String HOST_NAME_DEFAULT = "";
 
+  public static final String ADVERTISED_LISTENERS_CONFIG = "advertised.listeners";
+  protected static final String ADVERTISED_LISTENERS_DOC =
+      "List of advertised listeners. Used when generating absolute URLs in responses. Protocols"
+          + " http and https are supported. Each listener must include the protocol, hostname, and"
+          + " port. For example: http://myhost:8080, https://0.0.0.0:8081";
+  protected static final String ADVERTISED_LISTENERS_DEFAULT = "";
+
   public static final String CONSUMER_MAX_THREADS_CONFIG = "consumer.threads";
   private static final String CONSUMER_MAX_THREADS_DOC =
       "The maximum number of threads to run consumer requests on."
@@ -341,6 +348,13 @@ public class KafkaRestConfig extends RestConfig {
         Type.STRING,
         HOST_NAME_DEFAULT,
         Importance.MEDIUM, HOST_NAME_DOC
+    )
+    .define(
+        ADVERTISED_LISTENERS_CONFIG,
+        Type.LIST,
+        ADVERTISED_LISTENERS_DEFAULT,
+        Importance.MEDIUM,
+        ADVERTISED_LISTENERS_DOC
     )
     .define(
         CONSUMER_MAX_THREADS_CONFIG,
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/backends/BackendsModule.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/backends/BackendsModule.java
@@ -13,28 +13,26 @@
  * specific language governing permissions and limitations under the License.
  */
 
-package io.confluent.kafkarest.resources;
+package io.confluent.kafkarest.backends;
 
 import io.confluent.kafkarest.KafkaRestContext;
-import io.confluent.kafkarest.resources.v1.V1ResourcesFeature;
-import io.confluent.kafkarest.resources.v2.V2ResourcesFeature;
-import io.confluent.kafkarest.resources.v3.V3ResourcesFeature;
+import io.confluent.kafkarest.backends.kafka.KafkaModule;
 import java.util.Objects;
-import javax.ws.rs.core.Feature;
-import javax.ws.rs.core.FeatureContext;
+import org.glassfish.hk2.utilities.binding.AbstractBinder;
 
-public final class ResourcesFeature implements Feature {
+/**
+ * A module to configure access to external dependencies.
+ */
+public final class BackendsModule extends AbstractBinder {
+
   private final KafkaRestContext context;
 
-  public ResourcesFeature(KafkaRestContext context) {
+  public BackendsModule(KafkaRestContext context) {
     this.context = Objects.requireNonNull(context);
   }
 
   @Override
-  public boolean configure(FeatureContext configurable) {
-    configurable.register(new V1ResourcesFeature(context));
-    configurable.register(new V2ResourcesFeature(context));
-    configurable.register(V3ResourcesFeature.class);
-    return true;
+  protected void configure() {
+    install(new KafkaModule(context));
   }
 }
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ClusterManagerImpl.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ClusterManagerImpl.java
@@ -43,53 +43,53 @@ final class ClusterManagerImpl implements ClusterManager {
 
   @Override
   public CompletableFuture<List<Cluster>> listClusters() {
-    DescribeClusterResult describeCluster =
+    DescribeClusterResult describeClusterResult =
         adminClient.describeCluster(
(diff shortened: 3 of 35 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/backends/BackendsModule.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private final KafkaRestContext context;
+
+  public BackendsModule(KafkaRestContext context) {
+    this.context = Objects.requireNonNull(context);
+  }
+
+  @Override
+  protected void configure() {
+    install(new KafkaModule(context));
```
mhowlett: why is this registered via `BackendsModule`? what's the advantage of the extra level?

(note: i may well ask some stupid questions due to ignorance of the frameworks etc., this may well be one of them).

kafka-rest-common/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -65,6 +65,13 @@
       + " hostname is used";
   public static final String HOST_NAME_DEFAULT = "";
 
+  public static final String ADVERTISED_LISTENERS_CONFIG = "advertised.listeners";
```
mhowlett: I have not spent any time thinking about the advantages of including absolute URLs in responses, but for your consideration, want to note that requiring this additional config seems like a material downside.

rigelbm: The configuration is not required. See UrlFactoryImpl for how the absolute URLs are generated. This configuration is to allow for generating URLs using a LB.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ClusterManagerImpl.java
```diff
@@ -0,0 +1,119 @@
...
+  private final Admin adminClient;
+
+  @Inject
+  ClusterManagerImpl(Admin adminClient) {
+    this.adminClient = Objects.requireNonNull(adminClient);
+  }
+
+  @Override
+  public CompletableFuture<List<Cluster>> listClusters() {
+    DescribeClusterResult describeCluster =
```
mhowlett: maybe `describeCluster` -> `describeClusterResult` is better?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ClusterManagerImpl.java
```diff
@@ -0,0 +1,119 @@
...
+        adminClient.describeCluster(
+            new DescribeClusterOptions().includeAuthorizedOperations(false));
+
+    return CompletableFuture.completedFuture(new Cluster.Builder())
+        .thenCombine(
+            toCompletableFuture(describeCluster.clusterId()),
+            (cluster, clusterId) -> {
```
mhowlett: maybe `cluster` -> `clusterBuilder` is better?

kafka-rest/src/main/java/io/confluent/kafkarest/response/FakeAsyncResponse.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.Date;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+import javax.ws.rs.container.AsyncResponse;
+import javax.ws.rs.container.TimeoutHandler;
+
+/**
+ * A fake {@link AsyncResponse} to be used in tests.
```
mhowlett: if it's only for use in tests, does it belong under `/main`?

kafka-rest/src/main/java/io/confluent/kafkarest/response/FakeUrlFactory.java
```diff
@@ -0,0 +1,35 @@
...
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.response;
+
+import java.util.StringJoiner;
+
+/**
+ * A fake {@link UrlFactory} to be used in tests.
```
mhowlett: better under `/test`?

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/ClusterManagerImplTest.java
```diff
@@ -0,0 +1,207 @@
...
+  private ClusterManagerImpl clusterManager;
+
+  @Before
+  public void setUp() {
+    clusterManager = new ClusterManagerImpl(adminClient);
+  }
+
+  @Test
+  public void listClusters_returnListWithOwnCluster() throws Exception {
+    expect(adminClient.describeCluster(anyObject())).andReturn(describeClusterResult);
```
agg111: Why do we specify in the test name - `returnListWithOwnCluster` ? Don't we simply list all the clusters ?

agg111:
Trying to follow the PR, since it has quite a few changes, I thought of following through tests to begin with, also I am pretty new to REST, might come up with simple questions. I notice the existing repo has folders as `v1`, `v2`, under resources and entities. I read in an older PR (#622), we would want to separate these but would have shared controllers. I see this PR creates separate `/controller` folder with this PR, and we still have some classes like Broker, Cluster under /entity (not /entity/v3). Can you explain why we do this ?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ClusterManagerImpl.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ClusterManagerImpl.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/response/FakeAsyncResponse.java
rigelbm: There's a difference between code which is used by tests, and tests. I'm used to keep code that is used by tests on main. That kind of forces me to do it properly and to test it. Keeping it as is.

kafka-rest/src/main/java/io/confluent/kafkarest/response/FakeUrlFactory.java
rigelbm: Ditto.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/ClusterManagerImplTest.java
rigelbm: So, there's catch currently which is that Kafka (and the REST Proxy) does not know about other clusters. It only knows about the Cluster pointed by the `bootstrap.servers` property. I created the API this way so that in the future, when/if we do have information about other clusters, we can return them in the list as well. But right now, listing clusters always returns only one cluster.

rigelbm:
The entities that are in `entities/` but not in `v?/` are used the controllers as input-output. The entities in `entities/v?/` are used by the resources as request/response types. You see the difference by noting that they are annotated with json-mapping annotations.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/ClusterManagerImplTest.java
agg111: Okay, thank you.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/BrokersResource.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response.Status;
+
+@Path("/v3/clusters/{clusterId}/brokers")
+public final class BrokersResource {
+
+  @GET
+  @Produces(Versions.JSON_API)
+  public void listBrokers(@PathParam("clusterId") String clusterId) {
```
agg111: I see in v2 we were using @PerformanceMetric, why don't we use it now ?

rigelbm: We can add performance metrics to all endpoints later.

PR #631, opened 2020-03-06
CLIENTS-1475: Add LIST/GET Brokers V3 endpoints.

rigelbm:
This PR adds endpoints to list and get brokers:

* /v3/clusters/{clusterId}/brokers
* /v3/clusters/{clusterId}/brokers/{brokerId}

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerManagerImpl.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerManagerImpl.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.controllers;
+
+import io.confluent.kafkarest.entities.Broker;
+import io.confluent.kafkarest.entities.Cluster;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import javax.inject.Inject;
+import javax.ws.rs.NotFoundException;
+
+final class BrokerManagerImpl implements BrokerManager {
+
+  private final ClusterManager clusterManager;
+
+  @Inject
+  BrokerManagerImpl(ClusterManager clusterManager) {
+    this.clusterManager = Objects.requireNonNull(clusterManager);
+  }
+
+  @Override
+  public CompletableFuture<List<Broker>> listBrokers(String clusterId) {
+    return clusterManager.getCluster(clusterId)
+        .thenApply(cluster ->
+            cluster.map(Cluster::getBrokers)
+                .orElseThrow(
+                    () -> new NotFoundException(
+                        String.format("Cluster %s cannot be found.", clusterId))));
+  }
+
+  @Override
+  public CompletableFuture<Optional<Broker>> getBroker(String clusterId, int brokerId) {
+    return listBrokers(clusterId)
+        .thenApply(brokers ->
+            brokers.stream().filter(broker -> broker.getBrokerId() == brokerId).findAny());
+  }
+}
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerManagerImpl.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerManagerImpl.java
@@ -15,6 +15,9 @@
 
 package io.confluent.kafkarest.controllers;
 
+import static io.confluent.kafkarest.controllers.Entities.checkEntityExists;
+import static io.confluent.kafkarest.controllers.Entities.findEntityByKey;
+
 import io.confluent.kafkarest.entities.Broker;
 import io.confluent.kafkarest.entities.Cluster;
 import java.util.List;
@@ -22,7 +25,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import javax.inject.Inject;
-import javax.ws.rs.NotFoundException;
 
 final class BrokerManagerImpl implements BrokerManager {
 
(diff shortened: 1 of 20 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/controllers/KafkaFutures.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import org.apache.kafka.common.KafkaFuture;
+
+final class KafkaFutures {
+
+  private KafkaFutures() {
+  }
+
+  static <T> CompletableFuture<T> toCompletableFuture(KafkaFuture<T> kafkaFuture) {
```
gwenshap: This looks useful. Did you consider contributing to Apache Kafka, so KafkaFuture object will support this conversion out of the box?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerManagerImpl.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public CompletableFuture<List<Broker>> listBrokers(String clusterId) {
+    return clusterManager.getCluster(clusterId)
+        .thenApply(cluster ->
+            cluster.map(Cluster::getBrokers)
+                .orElseThrow(
+                    () -> new NotFoundException(
+                        String.format("Cluster %s cannot be found.", clusterId))));
```
rnpridgeon: Is it that the cluster can't be found or no brokers. If it's the former shouldn't `clusterManager.getCluster` throw this instead?

kafka-rest/src/main/java/io/confluent/kafkarest/entities/Broker.java
```diff
@@ -25,25 +25,36 @@
  */
 public final class Broker {
 
-  private int brokerId;
+  private final String clusterId;
```
rnpridgeon: Would it make sense to hold a reference back to the cluster which owns this broker as opposed to the cluster id. It may(or may not) be nice (or one day be nice) for the broker to be able to introspect on its status within the cluster. i.e. isControler.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/BrokerData.java
```diff
@@ -0,0 +1,237 @@
...
+    public String getClusterId() {
+      return clusterId;
+    }
+
+    @JsonProperty("broker_id")
+    public Integer getBrokerId() {
+      return brokerId;
+    }
+
+    @JsonProperty("host")
```
rnpridgeon: Perhaps I'm misunderstanding the intent of this class but should we be using 'listener_host' and 'listener_post' here ? The one -pager uses 9093 as an example port which is pretty typical for an advertised listener. Furthermore should this actually be an array as brokers may have multiple listeners which may be bound to different interfaces? And does it make sense to advertise the security protocol as well? i.e. SASL_PLAINTEXT, SASL_SSL etc. The sasl mechanism itself isn't necessary as that is negotiated by the client anyway. If the idea is to leave that up to the configuration resource then perhaps the port doesn't belong here at all.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerManagerImpl.java
rigelbm: No cluster has been found. No brokers found is signalled by empty list. The ClusterManager signals no cluster found with Optional.empty(). Basically, empty-value for the thing you are looking for not found. Exception when you are looking for the wrong thing (e.g. for brokers of a non-existing cluster).

kafka-rest/src/main/java/io/confluent/kafkarest/entities/Broker.java
rigelbm: I'm trying to keep the object-ref graph as acyclic as I can. We can add things like you said (e.g. isController()) later. I think it's possible to do it without adding a back-reference. But if one is needed I'm happy to add it. In the future. :)

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerManagerImpl.java
mhowlett: [pre-note: all comments with the disclaimer that I don't know Java..]

@rnpridgeon has a good point here.

1. you're applying map to a single value, which apparently works but is not a great idea since map is something typically applied to a collection, so someone skimming it is going to pick up on the wrong signals, particularly if they don't use java because the code is just weird.
2. logically, it's more straightforward to check for the cluster existing first, and throw the exception rather than propagate this information through the map function, so you should prefer that. It's also more straightforward to write. i.e.:

```
    return clusterManager.getCluster(clusterId).thenApply(
      cluster -> cluster
        .orElseThrow(() -> new NotFoundException(String.format("Cluster %s not found.", clusterId)))
        .getBrokers());
```

further notes:

1. I'd phrase the error message differently as above.
2. indentation. i'm not sure what's javery but one thing is it's most clear if `.orElseThrow` & `.getBrokers` (or `.map` in your original version) are at the same level imo.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/Broker.java
mhowlett: given the DI setup and how this is used, what @rigelbm has lgtm

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/BrokerData.java
mhowlett: the components of the result are pretty much dictated by `DescribeClusterResult` and the result of that i'm guessing is dependent on the connection used to call it. but your point seems good - why not list listeners? seems better? maybe @cmccabe has input as the root of the decision is pretty much the describe cluster call. But given the APIs, what's here sgtm.

i think the word listener is great if encapsulating everything @rnpridgeon is talking about, but if there's not more than one, it's implicit.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/BrokersResource.java
```diff
@@ -16,33 +16,129 @@
...
+      @Suspended AsyncResponse asyncResponse,
+      @PathParam("clusterId") String clusterId,
+      @PathParam("brokerId") Integer brokerId
+  ) {
+    CompletableFuture<GetBrokerResponse> response =
+        brokerManager.getBroker(clusterId, brokerId)
+            .thenApply(
+                broker ->
+                    broker.map(this::toBrokerData)
+                        .map(GetBrokerResponse::new)
```
mhowlett: same comment as before, re `map`, ordering, etc.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/KafkaFutures.java
rigelbm: I've never done an AK PR. Sounds exciting :)

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerManagerImpl.java
rigelbm: I think a way to summarise what you are saying is: "get rid of the optional as soon as possible", which I completely agree. Sometimes it's not possible or convenient, but that's not the case here. You are completely correct. About the `map`, that's what you use to transform optionals in Java. I think they designed the optional [monad](https://en.wikipedia.org/wiki/Monad_(functional_programming)) as a [functor](https://en.wikipedia.org/wiki/Functor_(functional_programming)), which is why you can use `[map](https://en.wikipedia.org/wiki/Map_(higher-order_function))` on it. Regardless, I will do the change. Thanks!

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/BrokerData.java
rigelbm: This is what is currently exposed by AdminClient, which I assume to be the advertised listener, which is what the clients use. I can't see a mechanism to find the security information from the AdminClient currently. This is definitely something we will have to improve on later. But for now, keeping as is, as it will feel familiar to whoever knows the AdminClient.

PR #632, opened 2020-03-09
CPKAFKA-4617: Change Admin binding to be request-scoped.

rigelbm:
The request-scope is needed because the Admin creation logic uses request-scoped auth information. KafkaRestContext itself is a proxied instance, so the result will be cached there appropriately, after it is created.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
@@ -18,7 +18,9 @@ package io.confluent.kafkarest.backends.kafka;
 import io.confluent.kafkarest.KafkaRestContext;
 import java.util.Objects;
 import org.apache.kafka.clients.admin.Admin;
+import org.glassfish.hk2.api.Factory;
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
+import org.glassfish.jersey.process.internal.RequestScoped;
 
 /**
  * A module to configure access to Kafka.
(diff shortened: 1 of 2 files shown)


rigelbm:
> @rigelbm Will this avoid an admin client being created when the REST proxy starts up? Can we add a test that uses invalid client configs on the REST proxy to ensure that the proxy still starts up successfully without creating any client?

Done

kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
```diff
@@ -37,7 +39,33 @@ public KafkaModule(KafkaRestContext context) {
...
+    }
+
+    @Override
+    public Admin provide() {
+      return context.getAdmin();
+    }
+
+    @Override
+    public void dispose(Admin instance) {
+      // Do nothing.
```
cmccabe: Where is the AdminClient closed?

rigelbm: AdminFactory has no ownership of Admin. It only proxies it to KafkaRestContext. KafkaRestContext owns and disposes of it.

https://github.com/confluentinc/kafka-rest/blob/master/kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java#L112
https://github.com/confluentinc/kafka-rest/blob/master/kafka-rest/src/main/java/io/confluent/kafkarest/AdminClientWrapper.java#L157

PR #645, opened 2020-03-23
Fix test due to changes in MockAdminClient.

rigelbm:
`MockAdminClient` changed behaviour in https://github.com/apache/kafka/commit/56051e763965d439f11f20f876475732eed7b307 to not allow partitions with `null` leader. That breaks a test in kafka-rest that test specifically the behaviour of the server when Kafka returns a topic with a partition with null leader.

This commit changes the test to use a `@Mock AdminClient` instead of `MockAdminClient`, so that we can keep testing defensive behaviour against this corner case.


cmccabe:
Thanks for looking at this, but I don't think this is the right fix.  We should be able to use MockAdminClient in kafka-rest.  If we need the ability to have null leaders we should add that in.  For a short-term fix, we can put "ignore" on the test.

rigelbm:
Fixed upstream: https://github.com/apache/kafka/pull/8345

PR #647, opened 2020-03-30
Remove V1 APIs and all their dependencies.

rigelbm:


mhowlett:
at a high level looks fine to me, and i trust you've got it right. if you believe it would be valuable for me to check this in some depth, happy to do so (let me know), else LGTM.

ijuma:
Nice!

rigelbm:
Ignoring Jenkins failure, because it's due to a dependend build. A PR to fix the dependent build is ready, and will be merged after this one.

brianbushree:
@rigelbm I think we need to refactor some stuff [here](https://github.com/confluentinc/confluent-cloud-plugins/blob/master/kafka-rest-healthcheck-plugin/src/main/java/io/confluent/cloud/kafka/rest/health/HealthCheckResource.java) that depends on some of this code

PR #663, opened 2020-05-05
Remove RestConfigException from constructor signatures

rnpridgeon:
All KafkaRestConfig constructors throw RestConfigException despite this exception only being thrown in `public KafkaRestConfig(String propsFile)`. This makes for an undesirable developer experience.


rnpridgeon:
Test failure seems unrelated, looking into it though

PR #665, opened 2020-05-06
CPKAFKA-4979: Remove non-determinism from searchReplicasByBroker.

rigelbm:


kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -404,6 +404,23 @@ protected final void createTopic(String topicName, int numPartitions, short repl
     }
   }
 
+  protected final void createTopic(
```
agg111, line 407: So using the older way to createTopic by specifying the partitions and replicationFactor, it was not known what partitions the broker would be leader for, is that right ?

rigelbm: Yes. This way we can be sure.

PR #670, opened 2020-05-11
Add JsonCreator to v3 entities

rnpridgeon:
Makes v3 entities reader friendly.


kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/BrokersResourceIntegrationTest.java
```diff
@@ -115,12 +114,16 @@ public void listBrokers_existingCluster_returnsBrokers() throws Exception {
...
+                                + "/partition-replicas"))));
 
     Response response =
         request("/v3/clusters/" + clusterId + "/brokers").accept(Versions.JSON_API).get();
     assertEquals(Status.OK.getStatusCode(), response.getStatus());
-    assertEquals(expected, response.readEntity(String.class));
+
+    ListBrokersResponse actual = OBJECT_MAPPER.readValue(
+            response.readEntity(String.class),
+            ListBrokersResponse.class);
```
rigelbm, line 125: Here and everywhere else: 

I think this can be simplified as `ListBrokerResponse actual = response.readEntity(ListBrokerResponse.class);`

PR #678, opened 2020-05-23
ST-1323: Add license file so sources get attached

ewencp:


ewencp:
See https://github.com/confluentinc/common/pull/257 for explanation, but basically the naming is a somewhat hacky, but usable ootb, way of controlling whether sources are included given common is a parent for both open and proprietary code. I went attempted some other options that didn't rely on the limited default set of profile activations that maven supports, but they require changes to developer workflow (custom plugins that need to be installed and used for every affected repo). This was a reasonable compromise. We actually already have the multiple file layout for at least schema-registry since it contains both Apache and CCL components in that repo, so requires multiple license files.

A `licenses/` directory would be feasible, but also conflicts with the use of a `licenses/` directory for licenses of dependencies.

I don't love the solution, but it works and since most repos usually only need one license, I'm not too worried about clutter. Additionally, really only open repos need the extra license since it's used to turn on sources publishing.

rigelbm:
SGTM

PR #679, opened 2020-05-23
Migrate internal entities to Google's Auto-Value.

rigelbm:
The amount of boilerplate code I had to write in my last PR (https://github.com/confluentinc/kafka-rest/pull/677) to add two fields in a response, was just too much. Enough is enough, so I migrated our internal entities to use Google's Auto-Value, that auto-generates a lot of this boilerplate automatically.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerInstanceConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerInstanceConfig.java
@@ -15,36 +15,52 @@
 
 package io.confluent.kafkarest.entities;
 
+import com.google.auto.value.AutoValue;
 import io.confluent.kafkarest.KafkaRestConfig;
-import java.util.Objects;
 import java.util.Properties;
-import java.util.StringJoiner;
 import javax.annotation.Nullable;
 
-public final class ConsumerInstanceConfig {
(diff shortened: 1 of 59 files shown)


rnpridgeon:
For the uninitiated (myself) https://www.baeldung.com/introduction-to-autovalue

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerInstanceConfig.java
```diff
@@ -63,96 +79,7 @@ public ConsumerInstanceConfig(
...
-
-  @Override
-  public int hashCode() {
-    return Objects.hash(
+    return new AutoValue_ConsumerInstanceConfig(
         id, name, format, autoOffsetReset, autoCommitEnable, responseMinBytes, requestWaitMs);
   }
-
-  @Override
-  public String toString() {
```
rnpridgeon, line 147: What's the effect of a change like this on logging?

rigelbm: AutoValue will generate reasonably similar toStrings for these entities. Most of these entities, and their toStrings were added as part of 5.5 preview. I would not expect people to be depending on them.

Issue #538, opened 2019-02-25
LOG_DIR not set causing FileNotFoundException: /kafka-rest.log (Permission denied)


kijanowski:
To fix it, edit `bin/kafka-rest-run-class` and add:
```if [ "x$LOG_DIR" = "x" ]; then
  LOG_DIR="$base_dir/logs"
fi
```
just before
```
if [[ -n $LOG_DIR ]]; then
    [[ -d $LOG_DIR ]] || mkdir -p "$LOG_DIR"
    KAFKAREST_LOG4J_OPTS="-Dkafka-rest.log.dir=$LOG_DIR ${KAFKAREST_LOG4J_OPTS}"
fi```

MarkC0x:
Thanks for that information @kijanowski 

As this behavior was still happening in the latest 5.5.x version of Rest Proxy, I created a simple PR based on your suggestion.

PR #688, opened 2020-06-03
#538 - LOG_DIR not set causing FileNotFoundException

MarkC0x:
Noticed that this same `LOG_DIR` not set causing `FileNotFoundException` behavior was still happening in the latest 5.5.x version of Rest Proxy.

Created a simple PR to correctly set the base log directory for the logs.

### DIFF
--- a/bin/kafka-rest-run-class
+++ b/bin/kafka-rest-run-class
@@ -39,6 +39,11 @@ if [ "x$KAFKAREST_LOG4J_OPTS" = "x" ]; then
   fi
 fi
 
+# Log directory to use
+if [ "x$LOG_DIR" = "x" ]; then
+  LOG_DIR="$base_dir/logs"
+fi
+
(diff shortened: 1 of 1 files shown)


bin/kafka-rest-run-class
```diff
@@ -39,6 +39,11 @@ if [ "x$KAFKAREST_LOG4J_OPTS" = "x" ]; then
   fi
 fi
 
+# Log directory to use
+if [ "x$LOG_DIR" = "x" ]; then
```
rigelbm, line 43: [[ -z "${LOG_DIR+x}" ]] && LOG_DIR="$base_dir/logs"

rigelbm: BTW, not sure if `$base_dir/logs` is a better default log folder than, for example, `/var/log/kafka-rest`. AK uses `/var/log/kafka` by default, I think.

PR #689, opened 2020-06-03
KIP-606 MetricsContext

rnpridgeon:


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestMetricsContext.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package io.confluent.kafkarest;
+
+import io.confluent.rest.metrics.RestMetricsContext;
+import org.apache.kafka.common.utils.AppInfoParser;
+
+import java.util.Map;
+
+public final class KafkaRestMetricsContext {
```
rigelbm, line 23: Should this implement `MetricsContext`?

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestMetricsContext.java
```diff
@@ -0,0 +1,71 @@
...
+      setResourceLabel(labelKey, labelValue);
+    }
+
+    metricsContext.setLabel(labelKey, labelValue);
+  }
+
+  /**
+   * Sets {@link RestMetricsContext} resource label if not previously set.
+   */
+  protected void setResourceLabel(String resource, String value) {
```
rigelbm, line 59: Why protected if the class is `final`?

rnpridgeon: Thanks, missed that when I made the class final.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestMetricsContext.java
rnpridgeon: Since KafkaRest itsself does not manage any metrics reporters and we can only propagate metrics contexts through configs it seemed appropriate to just expose the MetricsContext as opposed to implementing the interface.

PR #700, opened 2020-06-17
Add Batch Alter (UPDATE/DELETE) Broker/Cluster/Topic configs.

rigelbm:
New APIs:

* `POST /v3/clusters/{clusterId}/broker-configs:alter`
* `POST /v3/clusters/{clusterId}/topic-configs:alter`
* `POST /v3/clusters/{clusterId}/brokers/{brokerId}/configs:alter`
* `POST /v3/clusters/{clusterId}/topics/{topicName}/configs:alter`

Request body:

```
{
  "data": [
    {
      "name": "compression.type",
      "value": "gzip",
      "operation": "UPDATE"
    },
    {
      "name": "max.connections",
      "operation": "DELETE"
    }
  ]
}
```

`operation` can be `UPDATE` and `DELETE`. It's optional and defaults to `UPDATE` if not specified. `value` defaults to `null` and is ignored if `operation` is `DELETE`.

HTTP 204 No Content is returned on success. The update is atomic: either the whole batch succeeds or fails.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/AlterConfigBatchRequestData.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/AlterConfigBatchRequestData.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.entities.v3;
+
+import static java.util.Collections.unmodifiableList;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import io.confluent.kafkarest.entities.AlterConfigCommand;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+@AutoValue
+public abstract class AlterConfigBatchRequestData {
+
+  AlterConfigBatchRequestData() {
+  }
+
+  @JsonProperty("data")
+  public abstract ImmutableList<AlterEntry> getData();
+
+  public static AlterConfigBatchRequestData create(List<AlterEntry> data) {
+    return new AutoValue_AlterConfigBatchRequestData(ImmutableList.copyOf(data));
+  }
+
+  @JsonCreator
+  static AlterConfigBatchRequestData fromJson(@JsonProperty("data") List<AlterEntry> data) {
+    return create(data);
+  }
+
+  public final List<AlterConfigCommand> toAlterConfigCommands() {
+    ArrayList<AlterConfigCommand> commands = new ArrayList<>();
+    for (AlterEntry entry : getData()) {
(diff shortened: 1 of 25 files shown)


javi-redondo:
What does the response look like?

rigelbm:
> HTTP 204 No Content is returned on success. The update is atomic: either the whole batch succeeds or fails.

jaykimc3:
@rigelbm , just curious, what's the difference between these two?
* `UPDATE` a config with `null` value
vs.
* `DELETE` a config

rigelbm:
`null` is a valid value for some configs. `UPDATE` to `null` will set the value to `null`, while `DELETE` will reset the config to its default value, which might not be `null`.

jaykimc3:
@rigelbm , does the value type always `string`? (or, `null`).
for example, which one do you expect?

confluent.balancer.throttle.bytes.per.second=1048576
confluent.balancer.enable=true

1.
```
    {
      "name": "confluent.balancer.enable",
      "value": "true",
      "operation": "UPDATE"
    },
    {
      "name": "confluent.balancer.throttle.bytes.per.second",
      "value": "10485760",
      "operation": "UPDATE"
    },
```
 OR
2.
```
    {
      "name": "confluent.balancer.enable",
      "value": true,
      "operation": "UPDATE"
    },
    {
      "name": "confluent.balancer.throttle.bytes.per.second",
      "value": 10485760,
      "operation": "UPDATE"
    },
```

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/AlterConfigBatchRequestData.java
```diff
@@ -0,0 +1,116 @@
...
+        case UPDATE:
+          commands.add(AlterConfigCommand.update(entry.getName(), entry.getValue().orElse(null)));
+          break;
+
+        case DELETE:
+          commands.add(AlterConfigCommand.delete(entry.getName()));
+          break;
+
+        default:
+          throw new AssertionError("unreachable");
```
srpanwar-confluent, line 61: Would this make `operation` required property in the request json?

rigelbm:
@jaykimc3 value is always string or `null`.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/AlterConfigBatchRequestData.java
rigelbm: Not really. See line 93 where we default to UPDATE.

jaykimc3:
@rigelbm , could you provide some sample error responses? I'd like to know the error response structure for the following cases (if they would be different) -
* errors due to some invalid config values
* errors due to failures on one or more config updates
* errors due to any system failures
* any other unexpected/unknown errors or exceptions thrown

rigelbm:
> * errors due to some invalid config values

```
HTTP 400

{
    "error_code": 40002,
    "message": "Invalid config value for resource ConfigResource(type=BROKER, name='0'): Invalid value foobar for configuration max.connections: Not a number of type INT"
}
```

> * errors due to failures on one or more config updates
> * errors due to any system failures
> * any other unexpected/unknown errors or exceptions thrown

Just like above, but another error code, possible on the 500 series.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AlterTopicConfigBatchAction.java
```diff
@@ -0,0 +1,65 @@
...
+public final class AlterTopicConfigBatchAction {
+
+  private final Provider<TopicConfigManager> topicConfigManager;
+
+  @Inject
+  public AlterTopicConfigBatchAction(Provider<TopicConfigManager> topicConfigManager) {
+    this.topicConfigManager = requireNonNull(topicConfigManager);
+  }
+
+  @POST
```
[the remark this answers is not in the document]
rigelbm: Sorry, completely forgot to reply to this comment. So, this `:alter` operation is not a normal method in the resource. It's a custom methods. We use `POST` for all the custom methods, since they are basically RPCs. `PATCH` would be used to update a single resource (we actually use `PUT` for config, but we will support `PATCH`).

PR #708, opened 2020-07-08
SEC-1034: log4j migration to confluent repackaged version

niteshmor:
- Confluent repackaged version fixes CVE-2019-17571
- `common/pom.xml` excludes default log4j as a transitive dependency, and the repackaged version must be explicitly added.

Ref: https://github.com/confluentinc/common/pull/270

### DIFF
--- a/kafka-rest/pom.xml
+++ b/kafka-rest/pom.xml
@@ -67,6 +67,12 @@
             <artifactId>slf4j-log4j12</artifactId>
             <version>${log4j.version}</version>
         </dependency>
+        <!-- Use a repackaged version of log4j with security patches. Default log4j v1.2 is a transitive dependency of slf4j-log4j12, but it is excluded in common/pom.xml -->
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>confluent-log4j</artifactId>
+            <version>${confluent-log4j.version}</version>
+        </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


niteshmor:
Note that there is a cyclic dependency here. There's a PR in common which needs to get merged first. Once that PR gets merged, the 6.0.x + master branches in this repository will start seeing a failure. This PR will be the fix for those failures, and it should be merged once common has been updated.
Context: https://github.com/confluentinc/common/pull/270#issuecomment-655643974

kafka-rest/pom.xml
```diff
@@ -67,6 +67,12 @@
             <artifactId>slf4j-log4j12</artifactId>
             <version>${log4j.version}</version>
         </dependency>
+        <!-- Use a repackaged version of log4j with security patches. Default log4j v1.2 is a transitive dependency of slf4j-log4j12, but it is excluded in common/pom.xml -->
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>confluent-log4j</artifactId>
+            <version>${confluent-log4j.version}</version>
```
ewencp, line 74: does this need this or would version just be inherited?

niteshmor: not strictly necessary. But doesn't hurt either. I just followed what every other dependency block had in this file.

PR #711, opened 2020-07-08
CLIENTS-1613: Add configuration to enable/disable V2 and V3 APIs.

rigelbm:
The configurations are, respectively, `api.v2.enable` and `api.v3.enable`. Both default to true.


rnpridgeon:
Test failed with a registry timeout; seems unrelated but maybe give it a quick pass

PR #715, opened 2020-07-22
Add config endpoints to openapi spec.

rigelbm:


api/v3/kafka-rest-v3-spec.yaml
```diff
@@ -515,6 +909,45 @@ components:
...
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              value:
+                type: string
+                nullable: true
+          example:
+            value: 'gzip'
```
agg111, line 923: What does this value `gzip` mean ?

PR #717, opened 2020-07-23
CNKAF-1017: Update v3 error response format

agg111:
### What
For kafka-rest v3, we first thought of going the JSON:API way, but then decided against it. We updated the all APIs but forgot to update the following - [Illegal broker removal](https://github.com/confluentinc/ce-kafka-rest/blob/master/ce-kafka-rest-extensions/src/main/java/io/confluent/kafkarest/exceptions/IllegalBrokerRemovalException.java), [Non-existing broker removal](https://github.com/confluentinc/ce-kafka-rest/blob/master/ce-kafka-rest-extensions/src/main/java/io/confluent/kafkarest/exceptions/BrokerNotFoundException.java) and [Unsupported kafka protocols](https://github.com/confluentinc/kafka-rest/blob/master/kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/UnsupportedProtocolException.java). This PR modifies the error format for these Apis.

### Testing 
`mvn test`


agg111:
Opened against 60.x - #718

PR #720, opened 2020-07-23
Propagate Confluent Telemetry Reporter configuration properties to managed clients.

rnpridgeon:
KafkaRest managed clients should inherit the KafkaRestApplication's metrics reporter configs when not configured as a client override. 

There is a new metrics reporter, ConfluentTelemetryReporter, who's configurations need to be propagated as well. 

This PR addresses both concerns.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/KafkaRestConfigTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/KafkaRestConfigTest.java
@@ -52,14 +52,14 @@ public class KafkaRestConfigTest {
   @Test
   public void getProducerProperties_propagateMetricsProperties() {
     Properties properties = new Properties();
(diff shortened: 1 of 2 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/KafkaRestConfigTest.java
```diff
@@ -94,14 +94,14 @@ public void getConsumerProperties_propagateMetrics_defaultClusterId() {
   @Test
   public void getConsumerProperties_propagateMetricsProperties() {
     Properties properties = new Properties();
-    properties.put(reporter_config("telemetry.bootstrap.servers"), "broker1:9092");
+    properties.put(reporter_config("bootstrap.servers"), "broker1:9092");
```
xvrl, line 97: nit, bootstrap server might be a bit confusing since it's not an actual config, instead maybe use and actual config such as `api.key` or alternatively some placeholder such as `some.config`

PR #724, opened 2020-07-28
DOCS-4640 - Edit V3 Kafka REST config

joel-hamill:
Grammar edits for the V3 config.

### DIFF
--- a/api/v3/kafka-rest-v3-spec.yaml
+++ b/api/v3/kafka-rest-v3-spec.yaml
@@ -14,10 +14,10 @@ paths:
   /clusters:
     get:
       summary: 'List Clusters'
-      description: 'Returns the list of known Kafka clusters. Right now both Kafka and Kafka REST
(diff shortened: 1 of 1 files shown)


api/v3/kafka-rest-v3-spec.yaml
```diff
@@ -289,7 +289,7 @@ paths:
 
     get:
       summary: 'List Consumers'
-      description: 'Returns the list of consumers belonging to the given consumer group.'
+      description: 'Returns a list of consumers that belong to the specified consumer group.'
```
agg111, line 292: nit: are we referring the list returned with article `a` or `the` throughtout the doc ?

joel-hamill: @agg111 "a list" is more appropriate here. "the list" could be interpreted as a static unchanging list, rather than a current snapshot of consumers.

PR #728, opened 2020-07-31
MINOR: Rename open api spec file

agg111:
### What
Rename OpenApi spec file so that consolidating all the spec files is easier from docs repo.


rnpridgeon:
Spec files will never have to consolidated across projects right? I.e. We aren't risking a collision with telemetry reporter or mds openapi spec files.

agg111:
Rebased to resolve a conflict, hence the force push. 

> Spec files will never have to consolidated across projects right? I.e. We aren't risking a collision with telemetry reporter or mds openapi spec files.

While consolidating we would mention the project directories for spec yamls. For eg, we now have - `API_SPEC_DIRS ?= .hidden/kafka-rest/api/v3 .hidden/ce-kafka-rest/api/v3`

PR #729, opened 2020-08-05
METRICS-1988: add confluent-telemetry directory to kafka-rest-run-class

brianbushree:
### what
Add `share/java/confluent-telemetry` to kafka-rest's classpath.

### why
This is where we'll install the confluent-metrics jar (as part of packaging) and we need this jar for telemetry to work.

### DIFF
--- a/bin/kafka-rest-run-class
+++ b/bin/kafka-rest-run-class
@@ -22,7 +22,7 @@ for dir in $base_dir/kafka-rest/target/kafka-rest-*-development; do
 done
 
 # Production jars
-for library in "confluent-security/kafka-rest" "confluent-common" "rest-utils" "kafka-rest" "monitoring-interceptors"; do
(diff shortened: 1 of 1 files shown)


rnpridgeon:
We will need to update the RPM spec and Deb files to mark the telemetry reporter package as a dependency as well. 


https://github.com/confluentinc/kafka-rest/blob/master/debian/confluent-kafka-rest.spec.in#L14

Also make sure all files are accounted for in the spec file as well 
https://github.com/confluentinc/kafka-rest/blob/master/debian/confluent-kafka-rest.spec.in#L86

https://github.com/confluentinc/kafka-rest/blob/master/debian/control#L12

rnpridgeon:
Spoke with @brianbushree offline we will be requiring users to install this separately meaning it will not be enabled by default 👍

PR #737, opened 2020-08-18
KC-613: Split share/java/kafka-rest in kafka-rest-bin and kafka-rest-lib.

rigelbm:
This PR splits the packaged contents of kafka-rest in two folders share/java/kafka-rest-bin and share/java/kafka-rest-lib. The later is to be used when including kafka-rest as library in another artifact, while both should be used when running kafka-rest as a standalone service.

This PR also changes the test environments to use the latest packaged cp-kafka-rest image as base image. This is to make sure the changes in this PR work with cp-kafka-rest image.


rnpridgeon:
Overall seems fine but you may want to fork the packaging repo to build this package first to ensure it doesn't break any of the heuristics tests before merging.

ijuma:
This looks good, the only question I have is about the naming convention. @ewencp do we have something similar elsewhere? Does this naming convention seem OK to you?

ewencp:
I can't think of anywhere else we have this type of split. Schema registry comes to mind since it has both libs and a service but I think that's still all packaged in one directory. The location and naming is just based on what seems to be convention for debian java packages. That means that normally the directory would be named after the package itself, but I'm not sure what convention would be if one package needed to be split into two directories. As long as we've got unique naming, I can't think of a reason it would be a problem.

PR #739, opened 2020-08-21
ST-3461: Nano versioning

elismaga:
Setting the jenkins config to use nano versioning. No longer need to define upstream repos as we now define downstream repos which will be automated in the future. Removed all of the snapshot versions and replaced with nano versions. Updated all of the dependency versions. Updated the make file so the rpm version parsing works with nano versioning, and applied the batch mode option to all maven commands so we don't get download progress and color codes in the CI logs.

These changes will be merged during phase two of the nano versioning roll out.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,7 @@
 #!/usr/bin/env groovy
 common {
   slackChannel = '#rest-proxy-eng'
-  upstreamProjects = 'confluentinc/schema-registry'
+  downStreamRepos = ["confluent-security-plugins", "ce-kafka-rest",
+    "confluent-cloud-plugins"]
+  nanoVersion = true
 }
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env groovy
 common {
   slackChannel = '#rest-proxy-eng'
-  upstreamProjects = 'confluentinc/schema-registry'
+  downStreamRepos = ["confluent-security-plugins", "ce-kafka-rest",
```
mohnishbasha, line 4: any thoughts on why we replaced upstream with downstream here?

mohnishbasha (approved):

Jenkinsfile
elismaga: @mohnishbasha We are moving to a downstream model instead of upstream because it enables us to automate this in the future by using the dependency graph. Using the dependency graph we will no longer need to define these dependencies as we can just determine it from the graph. Having these build dependencies correct becomes more critical with nano versioning so we don't want to rely on a manual process to get it right and keep it updated.

PR #744, opened 2020-09-01
KREST-7: Make advertised.listeners and listeners precede host.name:port.

rigelbm:
For V3 APIs, the preference order for returned URLs should be:

1. advertised.listeners config
2. listeners config
3. host.name and port config
4. request base URL

Currently the order is (3, 1, 2, 4). Reorder them to the correct order.


rigelbm:
> Is ordering a breaking change? It's not clear to me this is worth blocking on but maybe I'm missing something. That said the change itself seems find so I've conditionally approved for 6.0.x pending blocker team approval and unconditionally approved for master

The order I am talking here is the order in which configurations are applied, which is, some would argue, a breaking change.

Issue #748, opened 2020-09-07
Not possible to increase default value consumer max.poll.records on kafkaRestConfig


cnwme:
Hi All,
I'm testing to increase the default value of max.poll.records = 30 to 500 version 2.4.2. when i force on kafka rest proxy config file like this : 
consumer.max.poll.records=500
at startup server, i hav i WARN : 
`WARN Property consumer.max.poll.records is not valid (io.confluent.kafka.utils.VerifiableProperties)`
all consumer still getting connection with default config : 
`max.poll.records = 30`

How to increase this property for all consumers ?

Regards
Martin

cnwme:
Thank you for this commit :)

kasarlaharish1:
Hi Mark,

Is this issue fixed in rest-proxy 7.5.0 version? When I am trying with KAFKA_REST_CONSUMER_MAX_POLL_RECORDS_CONFIG 
in rest proxy config it is not getting applied.

Thanks

PR #750, opened 2020-09-18
#748 - Allow config consumer.max.poll.records

MarkC0x:
Currently the consumer `max.poll.records` configuration cannot be altered from the internally coded value of `30`.  This PR allows the consumer value to be changed to the users choice by setting `consumer.max.poll.records` in the `kafka-rest.properties` file.  If a different value is not set by the user, then the previous internally coded value of `30` will be used.


kasarlaharish1:
Hi @MarkC0x ,

Why this pull request is still open from last 3 years? Are we not planning to push this change?

PR #754, opened 2020-10-02
KREST-20: Load tests for V2 and V3 APIs.

rigelbm:
This load tests where created using the k6 load-testing tool. See https://k6.io/docs/ for more information.

The results of a run of the tests on my machine can be seen here: https://docs.google.com/spreadsheets/d/1ay8xOX6daZK68Gtn3uOLpqg7MQtk0oUS65_5Qo6roWc/edit#gid=0.

This tests where executed on bare metal, with the following specs:

* AMD Ryzen 9 3900X 3.8 GHz 12-Core Precessor
* Kingston HyperX Predator 32 GB (2 x 16 GB) DDR4-3200 CL16
* Samsung 970 Evo Plus 1TB M.2-2280 NVME SSD
* Gigabyte X570 Aorus Master ATX AM4 Motherboard
* Arch Linux (Linux 5.8.13-arch1-1) x86-64

Except for `v2/consumer-binary-test.js`, all tests where executed with the default JVM options for REST Proxy.

For the `v2/consumer-binary-tests.js`, the JVM options used were:

```
-Xms1g
-Xmx1g
-XX:MetaspaceSize=96m
-XX:+UseG1GC
-XX:MaxGCPauseMillis=20
-XX:InitiatingHeapOccupancyPercent=35
-XX:G1HeapRegionSize=16M
-XX:MinMetaspaceFreeRatio=50
-XX:MaxMetaspaceFreeRatio=80
```

I added an experimental script to run all tests and collect all results into a spreadsheet. Here is an example spreadsheet summary of the v3 tests:
https://drive.google.com/file/d/1BzOMYxaxERjof477JouuGiIRVZ1nEQz3/view?usp=sharing


testing/environments/minimal/docker-compose.yml
```diff
@@ -91,6 +100,9 @@ services:
       KAFKA_REST_API_ENDPOINTS_BLOCKLIST: "api.v3.acls.*"
       KAFKA_REST_BOOTSTRAP_SERVERS: "kafka-1:9191"
       KAFKA_REST_LISTENERS: "http://0.0.0.0:9391"
+      KAFKAREST_HEAP_OPTS: "-Xms1g -Xmx1g -XX:MetaspaceSize=96m -XX:+UseG1GC \
```
mhowlett, line 103: optional: if it's easy to comment or link to somewhere to better understand the rational for these settings, that'd be good. I'm assuming you're optimizing things.

testing/load-test/README.md
```diff
@@ -0,0 +1,29 @@
...
+$ k6 run <my-test-script.js>
+```
+
+For example:
+
+```shell script
+$ k6 run v3/get-broker-test.js
+```
+
+All tests will execute against a REST Proxy instance running on `http://localhost:9391` (this is
```
mhowlett, line 18: optional: it might be worth making this overridable with an env variable.

testing/load-test/collect-results.py
```diff
@@ -0,0 +1,123 @@
...
+                parsed_value = float(value)
+            else:
+                parsed_value = value
+            worksheet.write(row, i, parsed_value, value_format)
+
+        col += len(columns[column_name])
+
+    row += 1
+
+workbook.close()
```
mhowlett, line 123: fancy. is csv not adequate?

testing/load-test/v2/common.js
```diff
@@ -0,0 +1,229 @@
...
+import http from 'k6/http';
+import {Counter, Trend} from 'k6/metrics';
+
+import {randomIntBetween} from "https://jslib.k6.io/k6-utils/1.0.0/index.js";
+
+export let randomByteString = function (size) {
+    let str = '';
+    for (let i = 0; i < size; i++) {
+        // First 128 chars (US-ASCII) take 1 byte to UTF-8 encode.
+        str += String.fromCharCode(randomIntBetween(0, 128));
```
mhowlett, line 11: this is a stupid way of representing binary data, but it's k6's fault.

testing/load-test/v2/common.js
```diff
@@ -0,0 +1,229 @@
+import {check} from 'k6';
+import http from 'k6/http';
+import {Counter, Trend} from 'k6/metrics';
+
+import {randomIntBetween} from "https://jslib.k6.io/k6-utils/1.0.0/index.js";
+
+export let randomByteString = function (size) {
```
mhowlett, line 7: maybe this function should be a level up, since you will probably want it when v3 gets produce.

testing/load-test/v2/common.js
```diff
@@ -0,0 +1,229 @@
...
+    for (let i = 0; i < size; i++) {
+        // First 128 chars (US-ASCII) take 1 byte to UTF-8 encode.
+        str += String.fromCharCode(randomIntBetween(0, 128));
+    }
+    return str;
+}
+
+const baseUrl = 'http://localhost:9391';
+
+let produceBinaryToTopicRequestLatency = new Trend('ProduceBinaryToTopicRequestLatency', true);
```
mhowlett, line 18: it seems a bit odd how you have all the test specific stuff up in common, but it's perfectly understandable and i'm guessing has some justification in whatever magic is going on with k6 to do it's thing. if not, might consider moving stuff down to the specific test files, where appropriate.

testing/load-test/v2/consume-binary-test.js
```diff
@@ -0,0 +1,51 @@
...
+    ],
+    setupTimeout: '10m',
+    teardownTimeout: '1m'
+};
+
+export function setup() {
+    let listClustersResponse = listClusters();
+    let clusterId = randomItem(listClustersResponse.json().data).cluster_id;
+
+    // Use the topics/messages from produce-binary-to-topic-test.
```
mhowlett, line 20: i'd have made them independent i think (produce the messages here).

testing/load-test/v2/common.js
mhowlett: oh the methods are exported, makes sense.

testing/load-test/v2/common.js
```diff
@@ -0,0 +1,229 @@
...
+let produceJsonToTopicByteCount = new Counter('ProduceJsonToTopicByteCount');
+export let produceJsonToTopic = function (topicName, records) {
+    const url = `${baseUrl}/topics/${topicName}`;
+    const body = {records};
+    const headers = {
+        'Accept': 'application/vnd.kafka.v2+json',
+        'Content-Type': 'application/vnd.kafka.json.v2+json'
+    };
+    const tags = {name: 'ProduceJsonToTopic'};
+    let response = http.post(url, JSON.stringify(body), {headers, tags});
```
mhowlett, line 54: possibly naive thoughts:

wondering various things - why isn't this async?, does k6 use worker threads?, what is the perf cost of various bits of this JS code?. generally, I'm wondering if there is a bottleneck and if that is k6 or REST Proxy? network? I guess i'm thinking i wouldn't be surprised to find REST Proxy more performant than what this test framework can do (and if it's not, why isn't it), so i'm wondering if you're implicitly testing the performance of the test framework to any degree. also, if you're testing over a network other than localhost, I'm guessing you may be able to saturate the network, unless it's a really fast one, so wondering what throughput you're seeing relative to network maximum.

given you seem to be holding VUs constant in the tests, I assume the idea is for latency in request responses to provide back pressure on throughput making this a meaningful thing to measure. you can expect throughput characteristics to change quite a bit with number of concurrent connections I think though, so i think you definitely want to be testing a range. i guess you have probably at least experimented with this on an ad-hoc basis. and it would of course make the tests more time consuming / burdensome to run.

testing/load-test/v2/consume-binary-test.js
```diff
@@ -0,0 +1,51 @@
...
+
+    return {clusterId, consumers, topics};
+}
+
+export default function (data) {
+    let consumer = data.consumers[__VU % data.consumers.length];
+    consumeBinary(consumer.consumerGroupId, consumer.consumerId);
+}
+
+export function teardown(data) {
```
mhowlett, line 48: maybe adopt the style of putting an underscore in front of unused variables / parameters.

mhowlett: (also the teardown method elsewhere)

testing/load-test/v2/produce-binary-to-partition-test.js
```diff
@@ -0,0 +1,58 @@
...
+        listPartitionsResponse.json().data.forEach(
+            partition => partitions.push({topicName, partitionId: partition.partition_id}));
+    }
+
+    return {clusterId, topics, partitions};
+}
+
+export default function (data) {
+    let records = [];
+    let size = randomIntBetween(1, 100);
```
mhowlett, line 43: what is the goal of these tests? if performance regression detection is the goal, I guess using a distribution of batch sizes detects broader based changes in performance across the parameter space, and if you had to choose one measure for this specific goal, this is probably what you want. an alternative would be to choose a batch size that optimizes for throughput/latency. you could also argue that's important because it's a better number to brag about. a change in this would also be a pretty good indicator of change across other parts of the parameter space i think.

(no changes requests, just giving some commentary).

testing/load-test/v2/produce-binary-to-topic-test.js
```diff
@@ -0,0 +1,43 @@
...
+
+import {randomIntBetween, randomItem, uuidv4} from "https://jslib.k6.io/k6-utils/1.0.0/index.js";
+
+import {produceBinaryToTopic, randomByteString} from "./common.js";
+import {createTopic} from "../v3/common.js";
+
+export let options = {
+    stages: [
+        {duration: '10s', target: 2000},
+        {duration: '40s', target: 2000},
```
mhowlett, line 11: any reason the targets are much higher? what i'm getting from this is that the other tests have REST Proxy as bottle neck. what about in this case?

testing/load-test/v2/produce-binary-to-partition-test.js
```diff
@@ -0,0 +1,58 @@
...
+    for (let i = 0; i < size; i++) {
+        let message = uuidv4();
+        records.push({
+            'key': encoding.b64encode(message),
```
mhowlett, line 47: what's wrong with `randomByteString`?

PR #756, opened 2020-10-08
KREST-26: Add a blocklist config to allow disabling individual endpoints.

rigelbm:
Adds a config `api.endpoints.blocklist` to allow disabling individual endpoints. Each resource class/method is now annotated with a `@ResourceName` annotation that uniquely identifies a endpoints or set of endpoints. The config receives a list of such names to disabled.

If an endpoint is disabled, it will return HTTP 404 Not Found for GET requests, and HTTP 405 Method Not Allowed for everything else.


kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/DisabledOperationException.java
```diff
@@ -0,0 +1,13 @@
+package io.confluent.kafkarest.exceptions;
+
+import javax.ws.rs.core.Response.Status;
+
+public final class DisabledOperationException extends StatusCodeException {
+
+  public DisabledOperationException() {
+    super(
+        Status.METHOD_NOT_ALLOWED,
```
rnpridgeon, line 9: I'm not sure 405 is the most appropriate response. The method(GET?) may be allowed but the resource is inaccessible. By saying the method isn't allowed it implies the resource is still valid you just need to use a different method. Perhaps 404 or even 410 would be more precise. It's a bit of a stretch but even 501 seems more appropriate.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AclsResource.java
```diff
@@ -56,17 +58,87 @@
...
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.ConfigDef.Importance;
+import org.apache.kafka.common.config.ConfigDef.Type;
 
 @Path("/v3/clusters/{clusterId}/acls")
 public final class AclsResource {
 
+  private static final String API_V3_ACLS_ENABLE_CONFIG = "api.v3.acls.enable";
```
rnpridgeon, line 68: What if you took a single config that takes a list of resource apis like a feature lst. 

ap1.v3.enable = [CLUSTER, ACL, TOPIC]

And a second one for methods, although perhaps it would be better for this to be  negative list. 

api.v3.<resource_name>.disallow = [DELETE]

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/DisabledOperationException.java
rigelbm: I think 404 for GET, and 405 for everything else is the most appropriate. That's what the documentation for 501 leads me to believe: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501

rnpridgeon: That's fair I actually made this comment before I realized you were restricting resources as well as methods. 👍 In agreement.

rnpridgeon (approved):

rnpridgeon:
LGTM

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -336,6 +336,12 @@
           + "should be delegated to. Examples: confluent.cloud, mds-01.example.com.";
   private static final String CONFLUENT_RESOURCE_NAME_AUTHORITY_DEFAULT = "";
 
+  public static final String API_ENDPOINTS_BLOCKLIST_CONFIG = "api.endpoints.blocklist";
+  public static final String API_ENDPOINTS_BLOCKLIST_DOC =
+      "List of endpoints to disable in this server. For example: \"api.v3.acls.*\" or "
```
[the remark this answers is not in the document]
rnpridgeon: That is better!

PR #762, opened 2020-10-13
KREST-91: Fix NPE in /topics when a partition has no leader.

rigelbm:


rnpridgeon (approved):
That's unfortunate we don't have `None` node with id `-1`

PR #767, opened 2020-10-28
KREST-155: Make partitions_count and replication_factor optional in Create Topic V3 API.

rigelbm:
They default to `num.partitions` and `default.replication.factor` broker configs, respectively.


mhowlett (approved):
what are nullable configs for?

rigelbm:
> what are nullable configs for?

Discussed offline. The config stuff is just a code cleanup. No related changes to the API.

PR #768, opened 2020-10-29
KREST-147: Make absolute URL generation consistent in V2 and V3.

rigelbm:
Config advertised.listeners was added in 6.0.0 to specify which base path to use when generation absolute URLs in V3 API responses. Consume V2 API has a similar mechanism, that uses the host.name and port configs. This PR unifies both mechanisms, so that the URLs generated in both V2 and V3 are based on the same base path.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -27,8 +27,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.FileInputStream;
 import java.io.IOException;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.util.Map;
 import java.util.Properties;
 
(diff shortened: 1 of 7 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -838,18 +837,6 @@ public RestMetricsContext getMetricsContext() {
     return metricsContext.metricsContext();
   }
 
-  public int consumerPort(String scheme) throws URISyntaxException {
```
rnpridgeon, line 841: Do we need to deprecate this?

rigelbm: What do you mean? This is an internal method that is not used anymore?

rnpridgeon (approved):
absoluteUriBuilder and consumerPort feel like part of a public API and changing them may break someone's tests but I'll defer to your better judgement on that. Otherwise LGTM

rigelbm:
> absoluteUriBuilder and consumerPort feel like part of a public API and changing them may break someone's tests but I'll defer to your better judgement on that. Otherwise LGTM

Kafka REST Proxy is an HTTP service. It's is not a library. No code in it makes part of the public API (otherwise we would have hard time evolving it in any way). Only things that make part of the public API is the HTTP API itself, and the config options.

PR #769, opened 2020-10-30
KREST-12: Add seek to timestamp V2 API.

rigelbm:
In 6.0.0, we removed the only method of reading a message by timestamp when we deleted the V1 API. This PR adds a similar, albeit different API, to seek a V2 consumer instance to a certain timestamp.

The API is the same as the one used to seek a consumer instance to a offset: /consumers/{consumerGroupId}/instances/{consumerId}/positions.

That API's request body is extended to support specifying timestamps:

```
{
  "offsets": [
    {
      "topic": "topic-1",
      "partition": 0,
      "offset": 100,
      "metadata": "foobar"
    }
  ],
  "timestamps": [
    {
      "topic": "topic-2",
      "partition": 1,
      "timestamp": "2020-10-30T19:39:45+0000",
      "metadata": "fozbaz"
    }
  ]
}
```

ISO-8601 format is used to specify the timestamp.

This PR also fixes a bug, whereas a seek to an offset would not send the metadata field to Kafka.


mhowlett (approved):

reddybhargava:
Could you please take a look at #951 ?

PR #770, opened 2020-11-02
KREST-200: Never user listeners for URL generation in V2 and V3.

rigelbm:
https://github.com/confluentinc/kafka-rest/pull/768 introduced a regression in V2 absolute URL generation, whereas before listeners was never used for generation, and now it takes precedence over request URI. This PR fixes the regression by never using listeners for URL generation (both for V2 and V3 APIs). Request URI is always a better guess then listeners, and people should be able to override it using advertised.listeners or host.name and port as before.


rnpridgeon (approved):
LGTM

PR #771, opened 2020-11-03
Allow translation of enum path params based on @JsonValue.

rigelbm:
Jackson has a nice feature which allows it to parse JSON values as enums based on a method annotated in the enum class with `@JsonValue`. Currently we only support parsing enums from path params via name (case-insensitive) matching. This PR extends the parsing logic to also parse based on `@JsonValue` annotated method, if one is available in the class.


agg111 (approved):
LGTM

PR #773, opened 2020-11-06
KREST-287: Syncronize resource pools/managers lazy instantiation.

rnpridgeon:
It is possible for multiple requests to arrive prior to resource initialization which can result in a multiple pools being created/leaked. This PR intends to synchronize the process of obtaining these resources to avoid such a situation.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
@@ -54,7 +54,7 @@ public class DefaultKafkaRestContext implements KafkaRestContext {
   }
 
   @Override
-  public ProducerPool getProducerPool() {
+  public synchronized ProducerPool getProducerPool() {
     if (producerPool == null) {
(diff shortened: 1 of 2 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
```diff
@@ -54,7 +54,7 @@ public KafkaRestConfig getConfig() {
   }
 
   @Override
-  public ProducerPool getProducerPool() {
+  public synchronized ProducerPool getProducerPool() {
```
ijuma, line 57: We should use this pattern instead:

https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom

rnpridgeon: Only problem with that approach is that client Pools may only be a singleton for this use case. Our extensions for example create a KafkaRestContext per authenticated user in order propagate identities from the HTTP request to Kafka.

rigelbm (approved):
LGTM

PR #774, opened 2020-11-09
Remove kafka-rest-console-scripts.

rigelbm:
Reason: We don't need a separate target for scripts, specially because the only scripts we have are to start and stop the server, which are useless without the rest of the server's artefacts.

Real reason: I have no idea what this is needed for, and I mostly want to see what breaks (if anything) if I delete it.


rnpridgeon:
Did you verify the confluent docker images don't rely on this as the comment in this module's pom implies?

rigelbm:
> Did you verify the confluent docker images don't rely on this as the comment in this module's pom implies?

I did check `kafka-rest-images` (not thoroughly I must say), but could not find anything. I will checkout the repo and the parent docker repo to grep (thanks Github for the very suboptimal UI).

rnpridgeon (approved):
LGTM pending docker verification

rigelbm:
> LGTM pending docker verification

grepped "console-scripts" in `common-docker`, `confluent-docker-utils`, `cp-docker-images`, `docker-images`, `kafka-images`, `kafka-rest-images` and `packaging`. Not hits.

PR #776, opened 2020-11-09
KREST-297: Correctly propage schema.registry.url config to clients.

rigelbm:
For the Consume V2 API, client.schema.resgistry.url would always be overridden by schema.registry.url, even when not set.

Also cleanup KafkaRestConfig a bit. Mostly remove KafkaRestConfig.time. I mean, why does KafkaRestConfig has clock?!?!


rigelbm:
The failure is because a class I deleted is used somewhere else. Here is a PR that fixes that: https://github.com/confluentinc/ce-kafka-rest/pull/94.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -61,8 +49,6 @@
   public static final String ID_DEFAULT = "";
 
   public static final String MAX_POLL_RECORDS_CONFIG = "max.poll.records";
-  private static final String MAX_POLL_RECORDS_DOC =
-          "The maximum number of records returned in a single call to poll().";
```
mhowlett, line 65: why remove this doc string?

mhowlett: while we're in the rest proxy config, do you want to depreciate `PRODUCER_THREADS_CONFIG` given its not used? and maybe add to the doc string saying that it's not? (maybe a separate PR)

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -709,106 +677,63 @@ private static Properties getPropsFromFile(String propsFile) throws RestConfigEx
...
+  public KafkaRestConfig(ConfigDef configDef, Properties props) {
     super(configDef, props);
-    this.originalProperties = props;
-    this.time = time;
-    metricsContext = new KafkaRestMetricsContext(
-            getString(METRICS_JMX_PREFIX_CONFIG),
-            originalsWithPrefix(METRICS_CONTEXT_PREFIX));
-  }
-
-  public Time getTime() {
```
mhowlett, line 733: this seems odd and a good thing not to be here anymore.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/ConsumersResource.java
```diff
@@ -182,7 +183,7 @@ public void readRecordBinary(
         asyncResponse,
         group,
         instance,
-        timeout,
+        Duration.ofMillis(timeout),
```
mhowlett, line 186: `timeout_ms`, not `timeout` is a better query param name imo.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -129,7 +126,7 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
                   + "The thread pool must be satured or shutiing down", r);
```
mhowlett, line 126: maybe: `The runnable {} was rejected execution because the thread pool is saturated or shutting down`

mhowlett, line 126: replace this single log message (which has a typo and is not well worded) with a different one for each case

mhowlett: maybe:

`The runnable {} was rejected execution because the thread pool is saturated. Retrying in {}ms`.

`The runnable {} was rejected execution because the thread pool is terminating`.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -437,8 +425,8 @@ public void run() {
 
     @Override
     public long getDelay(TimeUnit unit) {
```
mhowlett, line 427: `getDelayMs`?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -437,8 +425,8 @@ public void run() {
...
-      long delayMs = waitExpirationMs - config.getTime().milliseconds();
-      return unit.convert(delayMs, TimeUnit.MILLISECONDS);
+      Duration delay = Duration.between(clock.instant(), waitExpiration);
```
mhowlett, line 428: i dunno, i'd probably wouldn't bother with the `Duration` etc stuff, i'd just work in ms internally. comment not intended to be actioned.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -363,21 +375,12 @@ synchronized long getEndOffset(String topic, int partition) {
         .map(OffsetAndTimestamp::offset);
   }
 
-  public synchronized boolean expired(long nowMs) {
-    return expiration <= nowMs;
+  public synchronized boolean expired(Instant nowMs) {
```
mhowlett, line 378: nowMs -> now

mhowlett (approved):
lgtm with minor comments

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: 1. max.poll.records is not configurable in REST Proxy. This value is used as is.
2. Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v2/ConsumersResource.java
rigelbm: Changed the names of the variables. Unfortunately we can't change the name of the query params, since that would not be backwards compatible.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
rigelbm: This implements `java.util.concurrent.Delayed#getDelay(TimeUnit)`. Can't rename.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
rigelbm: That's an anti-pattern in Java (8+), due to the excellent standard library support for semantic date/times.

PR #779, opened 2020-11-12
Configure ISO 8601 serialization of instants.

rigelbm:


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -135,7 +138,10 @@ protected ObjectMapper getJsonMapper() {
     return super.getJsonMapper()
         .registerModule(new GuavaModule())
         .registerModule(new Jdk8Module())
-        .registerModule(new JavaTimeModule());
+        .registerModule(new JavaTimeModule())
+        .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
+        .setDateFormat(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'"))
```
agg111, line 143: Do we want the time to mention seconds as well - `yyyy-MM-dd HH:mm:ss`?  Some sbk operation can be seconds apart.
An example I think of is tie spent by sbk on metric collection is mentioned in seconds precision to the user - https://github.com/confluentinc/ce-kafka/blob/350c93ba4c3a7683cbc66ca5d3002f8530189ca6/ce-sbk/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControl.java#L461

agg111 (approved):
LGTM, left one small comment.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
rigelbm: Good catch! Done. Thanks!

PR #784, opened 2020-12-04
Consumer Lag Metrics 

ahuang98:
Adding support for consumer group summary (max, total lag) and partition specific lag metrics to kafka-rest.
Modifying and using blueway's ConsumerOffsets* and ConsumerGroupOffsets files.


ahuang98:
openapi.yaml changes are being discussed here: https://github.com/confluentinc/kafka-rest/pull/765/

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDao.java
```diff
@@ -0,0 +1,243 @@
...
+
+public class ConsumerOffsetsDao {
+
+  private final Admin kafkaAdminClient;
+  private final int consumerMetadataTimeout;
+
+  @Inject
+  public ConsumerOffsetsDao(
+      Admin kafkaAdminClient,
+      @OffsetsTimeoutMsConfig int consumerMetadataTimeout
```
ahuang98, line 41:
```suggestion
      @OffsetsTimeoutMsConfig Duration consumerMetadataTimeout
```

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDao.java
```diff
@@ -0,0 +1,243 @@
...
+import org.apache.kafka.clients.admin.OffsetSpec;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.IsolationLevel;
+import org.apache.kafka.common.KafkaFuture;
+import org.apache.kafka.common.TopicPartition;
+
+public class ConsumerOffsetsDao {
+
+  private final Admin kafkaAdminClient;
+  private final int consumerMetadataTimeout;
```
ahuang98, line 36:
```suggestion
  private final long consumerMetadataTimeout;
```

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDao.java
```diff
@@ -0,0 +1,243 @@
...
+
+  @Inject
+  public ConsumerOffsetsDao(
+      Admin kafkaAdminClient,
+      @OffsetsTimeoutMsConfig int consumerMetadataTimeout
+  ) {
+    this.kafkaAdminClient = kafkaAdminClient;
+    this.consumerMetadataTimeout = consumerMetadataTimeout;
```
ahuang98, line 44:
```suggestion
    this.consumerMetadataTimeout = consumerMetadataTimeout.toMillis();
```

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -195,6 +195,13 @@
       + " Use 0 for no timeout";
   public static final String SIMPLE_CONSUMER_POOL_TIMEOUT_MS_DEFAULT = "1000";
 
+  public static final String OFFSETS_TIMEOUT_MS_CONFIG =
+      "offsets.timeout.ms";
+  private static final String OFFSETS_TIMEOUT_MS_DOC =
+      "Timeout for consumer group metadata requests when listing consumer group information "
+          + "for ConsumerOffsetsDao operations.";
+  public static final String OFFSETS_TIMEOUT_MS_DEFAULT = "15000";
```
ahuang98, line 203:
```suggestion
  public static final Duration OFFSETS_TIMEOUT_MS_DEFAULT = Duration.ofSeconds(15);
```

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDao.java
ahuang98, line 36: private final long consumerMetadataTimeout;

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDao.java
ahuang98, line 41: @OffsetsTimeoutMsConfig Duration consumerMetadataTimeout

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDao.java
ahuang98, line 44: this.consumerMetadataTimeout = consumerMetadataTimeout**.toMillis()**;

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -475,6 +482,13 @@ protected static ConfigDef baseKafkaRestConfigDef() {
         Importance.LOW,
         SIMPLE_CONSUMER_POOL_TIMEOUT_MS_DOC
     )
+    .define(
+        OFFSETS_TIMEOUT_MS_CONFIG,
+        Type.INT,
+        OFFSETS_TIMEOUT_MS_DEFAULT,
```
rigelbm, line 488: .getMillis()

kafka-rest/src/main/java/io/confluent/kafkarest/config/ConfigModule.java
```diff
@@ -119,4 +123,15 @@ protected void configure() {
   private static final class PortConfigImpl
       extends AnnotationLiteral<PortConfig> implements PortConfig {
   }
+
+  // ConsumerMetadataTimeout in blueway
+  @Qualifier
+  @Retention(RetentionPolicy.RUNTIME)
+  @Target({ ElementType.TYPE, ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER })
+  public @interface OffsetsTimeoutMsConfig {
```
rigelbm, line 131: We don't need the Me here

kafka-rest/src/main/java/io/confluent/kafkarest/config/ConfigModule.java
```diff
@@ -63,6 +63,10 @@ protected void configure() {
         .qualifiedBy(new ListenersConfigImpl())
         .to(new TypeLiteral<List<String>>() { });
 
+    bind(config.getInt(KafkaRestConfig.OFFSETS_TIMEOUT_MS_CONFIG))
+        .qualifiedBy(new OffsetsTimeoutMsConfigImpl())
+        .to(Integer.class);
```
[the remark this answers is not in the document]
ahuang98: @rigelbm 
Is there a way to get around AbstractConfig not having a `getDuration` defined? In order to change the above to 
```
bind(config.getDuration...
    .qualifiedBy...
    .to(new TypeLiteral<Duration>() { });
```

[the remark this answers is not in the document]
ahuang98: figured it out :)

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,234 @@
...
+
+  public ConsumerGroupDescription getConsumerGroupDescription(
+      String consumerGroupId
+  ) throws InterruptedException, ExecutionException, TimeoutException {
+    Map<String, ConsumerGroupDescription> allCgDesc =
+        getAllConsumerGroupDescriptions(ImmutableSet.of(consumerGroupId));
+    return allCgDesc.get(consumerGroupId);
+  }
+
+  public Map<String, ConsumerGroupDescription> getAllConsumerGroupDescriptions(
```
hachikuji, line 167: Which API does this satisfy? It doesn't seem too useful to get detailed member information for multiple groups at the same time.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -183,6 +184,13 @@
       + " Use 0 for no timeout";
   public static final String SIMPLE_CONSUMER_POOL_TIMEOUT_MS_DEFAULT = "1000";
 
+  public static final String OFFSETS_TIMEOUT_CONFIG =
+      "offsets.timeout";
+  private static final String OFFSETS_TIMEOUT_DOC =
+      "Timeout for consumer group metadata requests when listing consumer group information "
+          + "for ConsumerOffsetsDao operations.";
```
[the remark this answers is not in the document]
ahuang98: changed to 
`"Timeout for consumer group metadata requests for lag calculations. Can not exceed Integer.MAX_VALUE";`
(the timeout is associated with 3 different urls, would make for lengthy description)

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
ahuang98: It's currently only used to grab the description of single consumer groups (called by `getConsumerGroupDescription`). I was thinking of leaving it as is since it seems blueway uses this for c3/cloud UI's lag display (which the rest proxy needs to support in the future)?

api/v3/openapi.yaml
```diff
@@ -1183,6 +1230,92 @@ components:
...
+              type: array
+              items:
+                $ref: '#/components/schemas/ConsumerLagData'
+
+    ConsumerGroupLagData:
+      allOf:
+        - $ref: '#/components/schemas/Resource'
+        - type: object
+          required:
+            - cluster_id
```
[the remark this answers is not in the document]
ahuang98: It's mostly relevant for multi-tenant clusters

kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
```diff
@@ -96,6 +96,10 @@ protected String consumerNameFromInstanceUrl(String url) {
    * @return the new consumer instance's base URI
    */
   protected String startConsumeMessages(String groupName, String topic, EmbeddedFormat format,
+      String expectedMediatype) {
+    return startConsumeMessages(groupName, Collections.singletonList(topic), format, expectedMediatype);
+  }
+  protected String startConsumeMessages(String groupName, List<String> topics, EmbeddedFormat format,
```
ahuang98, line 102: added overloaded method to allow for a consumer to consume from multiple topics (initially needed for lag-related integration tests, but eventually removed the dependency on AbstractConsumerTest)

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,354 @@
...
+  @Inject
+  ConsumerOffsetsDaoImpl(
+      Admin kafkaAdminClient,
+      @OffsetsTimeoutConfig Duration consumerMetadataTimeout
+  ) {
+    this.kafkaAdminClient = kafkaAdminClient;
+    // ahu todo: figure out best way to safely cast to int (required by AbstractOptions.timeoutMs())
+    // this.consumerMetadataTimeout = Math.toIntExact(Long.MAX_VALUE);
+    this.consumerMetadataTimeout = Math.toIntExact(
+        Math.min(consumerMetadataTimeout.toMillis(), Integer.MAX_VALUE));
```
ahuang98, line 69: capped at Integer.MAX_VALUE to avoid ArithmeticException
(OFFSETS_TIMEOUT_DOC in KafkaRestConfig also notes Integer.MAX_VALUE is the largest value)

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ListConsumerLagsResource.java
```diff
@@ -0,0 +1,113 @@
...
+                                            "consumer-groups",
+                                            consumerGroupId,
+                                            "lags"))
+                                    .build())
+                            .setData(
+                                lags.stream()
+                                    .map(
+                                        lag ->
+                                            GetConsumerLagResource.toConsumerLagData(
+                                                lag, urlFactory, crnFactory))
```
ahuang98, line 103: calls `GetConsumerLagResource`'s `toConsumerLagData` to avoid having a duplicate of the method in this class

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,354 @@
...
+      ConsumerGroupDescription cgDesc,
+      Map<TopicPartition, OffsetAndMetadata> fetchedCurrentOffsets,
+      Map<TopicPartition, ListOffsetsResultInfo> latestOffsets
+  ) {
+    ConsumerGroupLag.Builder cgOffsets =
+        ConsumerGroupLag.builder().setConsumerGroupId(cgDesc.groupId());
+
+    // build map of topic partition -> consumer, client, instance ids
+    // Map<TopicPartition, String> tpConsumerIds = new HashMap<>();
+    // Map<TopicPartition, String> tpClientIds = new HashMap<>();
```
ahuang98, line 180: Comments kept temporarily for reference, blueway used two separate Maps to keep track of topic-partition to consumerId and clientId mappings. I created MemberId to keep track of these mappings (with the addition of instanceIds) in a single Map

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerLagManagerImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    this.adminClient = requireNonNull(adminClient);
+    this.consumerOffsetsDao = requireNonNull(consumerOffsetsDao);
+  }
+
+  @Override
+  public CompletableFuture<List<ConsumerLag>> listConsumerLags(
+      String clusterId, String consumerGroupId) {
+    try {
+      return CompletableFuture.completedFuture(
+          consumerOffsetsDao.getConsumerLags(
```
hachikuji, line 55: It looks like we have some async apis and some sync apis. Any reason they are not all async?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerLagManagerImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    CompletableFuture<ListOffsetsOptions> listOffsetsOptions =
+        CompletableFuture.supplyAsync(
+            () -> new ListOffsetsOptions(isolationLevel));
+    CompletableFuture<Map<TopicPartition, OffsetSpec>> latestOffsetSpecs = fetchedCurrentOffsets
+        .thenApply(map ->
+            map.keySet().stream()
+                .collect(Collectors.toMap(Function.identity(), tp -> OffsetSpec.latest())));
+  }
+
+//  @Override
```
hachikuji, line 83: nit: can we clean up the commented code? Do we still need to figure out this method or is it just not needed?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,362 @@
...
+          tp, MemberId.builder()
+              .setConsumerId("")
+              .setClientId("")
+              .setInstanceId(Optional.empty())
+              .build());
+
+      long currentOffset = getCurrentOffset(fetchedCurrentOffsets, tp);
+      long latestOffset = getOffset(latestOffsets, tp);
+      if (currentOffset < 0 || latestOffset < 0) {
+        // log.debug("invalid offsets for topic={} consumerId={} current={} latest={}",
```
hachikuji, line 220: nit: can we fix these logging comments?

kafka-rest/src/main/java/io/confluent/kafkarest/config/ConfigModule.java
```diff
@@ -144,4 +149,15 @@ protected void configure() {
   private static final class PortConfigImpl
       extends AnnotationLiteral<PortConfig> implements PortConfig {
   }
+
+  // ConsumerMetadataTimeout in blueway
```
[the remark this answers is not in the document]
ahuang98: will end up removing the timeout altogether (once everything is async)

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -183,6 +184,13 @@
       + " Use 0 for no timeout";
   public static final String SIMPLE_CONSUMER_POOL_TIMEOUT_MS_DEFAULT = "1000";
 
+  public static final String OFFSETS_TIMEOUT_CONFIG =
+      "offsets.timeout";
+  private static final String OFFSETS_TIMEOUT_DOC =
+      "Timeout for consumer group metadata requests for lag calculations. "
+          + "Can not exceed Integer.MAX_VALUE";
```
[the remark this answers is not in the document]
ahuang98: will end up removing the timeout altogether (once everything is async)

api/v3/openapi.yaml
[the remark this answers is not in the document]
ahuang98: That's a good point, @rigelbm?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerLagManagerImpl.java
ahuang98: Making them all async!

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerGroupLagManagerImpl.java
```diff
@@ -0,0 +1,101 @@
...
+                            consumerGroupId + " could not be found.")))
+        .thenCompose(
+            cgDesc ->
+                consumerOffsetsDao.getCurrentOffsets(consumerGroupId)
+                    .thenCompose(
+                        fetchedCurrentOffsets ->
+                            consumerOffsetsDao.getLatestOffsets(isolationLevel, fetchedCurrentOffsets)
+                                .thenCompose(
+                                    latestOffsets ->
+                                        consumerOffsetsDao.getMemberIds(cgDesc)
```
rigelbm, line 69: You don't have to chain these two calls, right? You can make them in parallel, because they don't depend on each other. 

```
consumerOffsetsDao.getCurrentOffsets(consumerGroupId)
    .thenCompose(
￼        fetchedCurrentOffsets -> {
            CompletableFuture<Foo> latestOffsets =
                consumerOffsetsDao.getLatestOffsets(isolationLevel, fetchedCurrentOffsets);
            CompletableFuture<Bar> memberIds =
                consumerOffsetsDao.getMemberIds(cgDesc);

            return CompletableFuture.allOf(latestOffsets, memberIds)
                .thenApply(
                    unused ->
                        createConsumerGroupLag(
                            clusterId,
                            consumerGroupId,
                            fetchedCurrentOffsets,
                            latestOffsets.join(),
                            memberIds.join()));
        })
```

Also, while you are at it, please extract that whole creation logic into another method, as suggested above (`createConsumerGroupLag`).

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerLagManagerImpl.java
```diff
@@ -0,0 +1,114 @@
...
+                checkEntityExists(cluster, "Cluster %s could not be found.", clusterId))
+        .thenCompose(
+            cluster ->
+                consumerOffsetsDao.getConsumerGroupDescription(consumerGroupId))
+        .thenApply(
+            cgDesc ->
+                cgDesc
+                    .orElseThrow(
+                        () -> new NotFoundException("Consumer group " +
+                            consumerGroupId + " could not be found.")))
```
rigelbm, line 60: Same comment as in `ConsumerGroupLagManagerImpl.java`.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerLagManagerImpl.java
```diff
@@ -0,0 +1,114 @@
...
+        .thenCompose(
+            cgDesc ->
+                consumerOffsetsDao.getCurrentOffsets(consumerGroupId).thenCompose(
+                    fetchedCurrentOffsets ->
+                            consumerOffsetsDao.getLatestOffsets(isolationLevel, fetchedCurrentOffsets)
+                                .thenCompose(
+                                    latestOffsets ->
+                                        consumerOffsetsDao.getMemberIds(cgDesc)
```
rigelbm, line 68: Same comment as in `ConsumerGroupLagManagerImpl.java`.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerGroupLagManagerImpl.java
```diff
@@ -0,0 +1,101 @@
...
+                                                            memberId.getClientId(),
+                                                            memberId.getInstanceId(),
+                                                            tp.partition(),
+                                                            currentOffset,
+                                                            latestOffset);
+                                                      });
+                                                  return Optional.of(cgOffsets.build());
+                                                }))));
+  }
+}
```
rigelbm, line 101: nit: Missing `\n` before `EOF`.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerGroupLagManagerImpl.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.CompletableFuture;
+import javax.inject.Inject;
+import javax.ws.rs.NotFoundException;
+import org.apache.kafka.common.IsolationLevel;
+
+final class ConsumerGroupLagManagerImpl implements ConsumerGroupLagManager {
+
+  private final ConsumerOffsetsDao consumerOffsetsDao;
+  private final ClusterManager clusterManager;
+  private final IsolationLevel isolationLevel = IsolationLevel.READ_COMMITTED;
```
rigelbm, line 33: Can be static (remember to rename to ISOLATION_LEVEL).

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerLagManager.java
```diff
@@ -0,0 +1,40 @@
...
+   * {@link io.confluent.kafkarest.entities.ConsumerGroup} with the given {@code consumerGroupId}.
+   */
+  CompletableFuture<List<ConsumerLag>> listConsumerLags(
+      String clusterId, String consumerGroupId);
+
+  /**
+   * Returns the Kafka {@link ConsumerLag} with the given {@code consumerId}.
+   */
+  CompletableFuture<Optional<ConsumerLag>> getConsumerLag(
+      String clusterId, String topicName, Integer partitionId, String consumerGroupId);
```
rigelbm, line 39: This ordering of parameters does not make sense, does it?

It should be `clusterId, consumerGroupId, topicName, partitionId`. 

If you have listFoo(x, y), and getFoo, getFoo is a refinement of listFoo, and the parameters should be strictly an extension of the parameters of listFoo, e.g. x, y, z.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerAssignmentManager.java
```diff
@@ -26,7 +26,7 @@
 public interface ConsumerAssignmentManager {
 
   /**
-   * Returns the list of Kafka {@link ConsumerAssignment Consumer Asignments} belonging to the
+   * Returns the list of Kafka {@link ConsumerAssignment Consumer Assignments} belonging to the
```
rigelbm, line 29: Thank you :)

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,165 @@
...
+  private final Admin kafkaAdminClient;
+  @Inject
+  ConsumerOffsetsDaoImpl(
+      Admin kafkaAdminClient
+  ) {
+    this.kafkaAdminClient = kafkaAdminClient;
+  }
+
+  @AutoValue
+  abstract static class MemberId {
```
rigelbm, line 51: Inner classes should defined be at the end of the outer class.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,165 @@
...
+
```
rigelbm, line 52: Create a package-protected constructor.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,165 @@
...
+    abstract Optional<String> getInstanceId();
+
+    public static Builder builder() {
+      return new AutoValue_ConsumerOffsetsDaoImpl_MemberId.Builder();
+    }
+
+    @AutoValue.Builder
+    abstract static class Builder {
+
```
rigelbm, line 65: Create a package-protected constructor.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,165 @@
...
+      abstract Builder setClientId(String clientId);
+
+      abstract Builder setInstanceId(Optional<String> instanceId);
+
+      abstract MemberId build();
+    }
+  }
+
+  @Override
+  public CompletableFuture<Map<TopicPartition, MemberId>> getMemberIds(
```
rigelbm, line 77: I don't understand why does this method need to return a future. Nothing inside of it returns a future. The future you return is always complete.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,165 @@
...
+      }
+    }
+    return CompletableFuture.completedFuture(tpMemberIds);
+  }
+
+  @Override
+  public CompletableFuture<Map<TopicPartition, OffsetAndMetadata>> getCurrentOffsets(
+      String consumerGroupId
+  ) {
+    return KafkaFutures.toCompletableFuture(
```
rigelbm, line 99: This method NEEDS to return a future. Good! The one above doesn't.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,165 @@
...
+        kafkaAdminClient.listConsumerGroupOffsets(
+            consumerGroupId, new ListConsumerGroupOffsetsOptions())
+            .partitionsToOffsetAndMetadata());
+  }
+
+  @Override
+  public CompletableFuture<Map<TopicPartition, ListOffsetsResultInfo>> getLatestOffsets(
+      IsolationLevel isolationLevel,
```
rigelbm, line 107: You use the same isolation level in both ConsumerGroupLagManagerImpl and ConsumerLagManagerImpl (READ_COMMITED). Why even bother receiving this as a parameter? Why don't you define it here (in this class) instead, and remove the parameter?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,165 @@
...
+                    .filter(
+                        // When describing a consumer-group that does not exist, AdminClient returns
+                        // a dummy consumer-group with simple=true and state=DEAD.
+                        // TODO: Investigate a better way of detecting non-existent consumer-group.
+                        description -> !description.isSimpleConsumerGroup()
+                            || description.state() != ConsumerGroupState.DEAD)
+                    .findAny());
+  }
+
+
```
rigelbm, line 139: nit: remove extra line.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,165 @@
...
+                descriptions.values()
+                    .stream()
+                    .filter(
+                        // When describing a consumer-group that does not exist, AdminClient returns
+                        // a dummy consumer-group with simple=true and state=DEAD.
+                        // TODO: Investigate a better way of detecting non-existent consumer-group.
+                        description -> !description.isSimpleConsumerGroup()
+                            || description.state() != ConsumerGroupState.DEAD)
+                    .findAny());
+  }
```
rigelbm, line 137: Why do you need this method at all? Why don't you inject ConsumerGroupManager and use getConsumerGroup instead? Is there something in ConsumerGroupDescription which is not reflected in ConsumerGroup? If that is the case, would not it be better to augment ConsumerGroup instead, and ConsumerGroupManager?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerOffsetsDaoImpl.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.kafka.clients.admin.ListOffsetsOptions;
+import org.apache.kafka.clients.admin.ListOffsetsResult.ListOffsetsResultInfo;
+import org.apache.kafka.clients.admin.MemberDescription;
+import org.apache.kafka.clients.admin.OffsetSpec;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.ConsumerGroupState;
+import org.apache.kafka.common.IsolationLevel;
+import org.apache.kafka.common.TopicPartition;
+
+final class ConsumerOffsetsDaoImpl implements ConsumerOffsetsDao {
```
rigelbm, line 40: I would make this class abstract, rename it AbstractConsumerLagManager, and have ConsumerGroupLagManagerImpl and ConsumerLagManagerImpl extend from it. Just to make sure no one else uses this class for anything. And, of course, delete the interface ConsumerOffsetsDao. And make all the methods package protected.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
```diff
@@ -0,0 +1,171 @@
...
+  ConsumerGroupLag() {
+  }
+
+  public abstract Long getMaxLag();
+
+  public abstract Long getTotalLag();
+
+  public abstract String getClusterId();
+
+  public abstract String getConsumerGroupId();
```
rigelbm, line 36: Primary Key first.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
```diff
@@ -0,0 +1,171 @@
...
+
+    abstract String getTopicName();
+
+    abstract String getConsumerId();
+
+    abstract String getClientId();
+
+    abstract Optional<String> getInstanceId();
+
+    abstract int getPartitionId();
```
rigelbm, line 63: Again, ordering: consumerId, clientId, instandId, topicName, partitionId.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
```diff
@@ -0,0 +1,171 @@
...
+    @AutoValue.Builder
+    abstract static class Builder {
+
+      abstract Builder setTopicName(String topicName);
+
+      abstract Builder setConsumerId(String consumerId);
+
+      abstract Builder setClientId(String clientId);
+
+      abstract Builder setInstanceId(Optional<String> instanceId);
```
rigelbm, line 86: You can have this as `setInstanceId(@Nullable String instanceId)`, so that you don't have to wrap it in a Optional when setting. Keep the accessor as `Optional<String> getInstanceId()`.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    private Long maxLag;
+    private long totalLag = 0;
+
+    private final Set<String> consumers = new HashSet<>();
```
rigelbm, line 104: What do you use `consumers` for? You add stuff in here, but you never query. Please, remove if you are not using it for anything. Same for any other piece of information that's not used anywhere in REST Proxy.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
```diff
@@ -0,0 +1,171 @@
...
+      abstract Builder setEndOffset(long endOffset);
+
+      abstract Offset build();
+    }
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    private Long maxLag;
```
rigelbm, line 101: All lags are non negative, right? So maybe have this initialized to -1 instead.

```
private long maxLag = -1;
```

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
```diff
@@ -0,0 +1,171 @@
...
+
+    Builder() {
+    }
+
+    public void addOffset(
+        String topicName,
+        String consumerId,
+        String clientId,
+        Optional<String> instanceId,
+        int partitionId,
```
rigelbm, line 114: Ordering...

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
```diff
@@ -0,0 +1,171 @@
...
+      Offset offset =
+          Offset.builder()
+              .setTopicName(topicName)
+              .setConsumerId(consumerId)
+              .setClientId(clientId)
+              .setInstanceId(instanceId)
+              .setPartitionId(partitionId)
+              .setCurrentOffset(currentOffset)
+              .setEndOffset(endOffset)
+              .build();
```
rigelbm, line 128: What do you need Offset for? The only method you ever call from it is `getLag()`, which is literally just `endOffset - currentOffset`. You can just add here `long lag = endOffset - currentOffset` and then erase that entire class.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerLag.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public abstract String getTopicName();
+
+  public abstract int getPartitionId();
+
+  public abstract String getConsumerId();
+
+  public abstract Optional<String> getInstanceId();
+
+  public abstract String getClientId();
```
rigelbm, line 40: Ordering...

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerLag.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public abstract Builder setTopicName(String topicName);
+
+    public abstract Builder setPartitionId(int partitionId);
+
+    public abstract Builder setConsumerId(String consumerId);
+
+    public abstract Builder setInstanceId(@Nullable String instanceId);
+
+    public abstract Builder setClientId(String clientId);
```
rigelbm, line 72: Ordering...

rigelbm (commented):
Thanks! It's getting close, but still needs more work. Please, reach out if you have any questions!

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerLagManager.java
ahuang98: I took the ordering from a previous conversation - https://github.com/confluentinc/kafka-rest/pull/765#discussion_r539581651 (for 'Specific lag for given topic-partition/consumer-group'). Should I change the order of parameters everywhere?

rigelbm: It should be either (1) `clusterId, topicName, partitionId, consumerGroupId, consumerId...` or (2) `clusterId, consumerGroupId, consumerId..., topicName, partitionId`. Pick one, doesn't matter which one, and use it everywhere. My comment here is that listConsumerLags is using `clusterId, consumerGroupId`, which means you probably chose (2), so you should use (2) everywhere.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
ahuang98: I believe it's possible to have negative lag, though the only scenario that comes to mind is after an unclean leader election. 
@hachikuji, do you have any input? I think it might be confusing to initialize maxLag at -1 if negative lag is possible

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerLagManager.java
ahuang98: @rigelbm just to confirm, if I use 
`/clusters/{cluster_id}/consumer-group/{consumer_group_id}/lags` 
for listing consumer lag, 
then I should use (something along the lines of) 
`/clusters/{cluster_id}/consumer-group/{consumer_group_id}/lags/topics/{topic_name}/partitions/{partition_id}` 
or 
`/clusters/{cluster_id}/consumer-group/{consumer_group_id}/lags/{topic_name}/partitions/{partition_id}` 
for individual consumer lags?

ahuang98: `/clusters/{cluster_id}/consumer-group/{consumer_group_id}/topics/{topic_name}/partitions/{partition_id}/lag`
would also work? though I think this means I would need to split the Get and List into two separate Resource files (like how they are currently)

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
ahuang98: Hm, I guess one other thing to consider is that totalLag may also then be 'lower' if there are negative lags at play, though I don't think this is really an issue. Could be confusing if totalLag happens to drop lower than maxLag though 😬

api/v3/openapi.yaml
```diff
@@ -1183,6 +1230,95 @@ components:
...
+            - consumer_group_id
+            - max_lag
+            - total_lag
+            - max_lag_consumer_id
+            - max_lag_consumer
+            - max_lag_client_id
+            - max_lag_topic_name
+            - max_lag_partition_id
+            - max_lag_partition
```
ahuang98, line 1296: @rigelbm are you good with this ordering?

ahuang98: move lags down, relations last

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AbstractConsumerLagManager.java
```diff
@@ -0,0 +1,158 @@
...
+            .collect(
+                Collectors.toMap(Function.identity(), topicPartition -> OffsetSpec.latest()));
+    return KafkaFutures.toCompletableFuture(
+        kafkaAdminClient.listOffsets(
+            latestOffsetSpecs,
+            new ListOffsetsOptions(isolationLevel)).all());
+  }
+
+  protected final long getCurrentOffset(Map<TopicPartition, OffsetAndMetadata> map, TopicPartition topicPartition) {
+    if (map == null) {
```
rigelbm, line 97: In which situations can `map` be null? `map` should not be allowed to be null. If `map` is the result of `getCurrentOffsets` or `getLatestOffsets`, make sure those functions either can't return null, of return empty-map when they would normally return null. Same for `getOffset` below.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AbstractConsumerLagManager.java
```diff
@@ -0,0 +1,158 @@
...
+            .stream()
+            .collect(
+                Collectors.toMap(Function.identity(), topicPartition -> OffsetSpec.latest()));
+    return KafkaFutures.toCompletableFuture(
+        kafkaAdminClient.listOffsets(
+            latestOffsetSpecs,
+            new ListOffsetsOptions(isolationLevel)).all());
+  }
+
+  protected final long getCurrentOffset(Map<TopicPartition, OffsetAndMetadata> map, TopicPartition topicPartition) {
```
rigelbm, line 96: Rename `map` to something more descriptive, e.g. `offsets`, `currentOffsets`, `latestOffsets`, etc. Same for `getOffset` below.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AbstractConsumerLagManager.java
```diff
@@ -0,0 +1,158 @@
...
+    if (map == null) {
+      return -1;
```
rigelbm, line 98: Should this function return Optional.empty() instead to signal "no-offset"? Same for `getOffset` below.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AbstractConsumerLagManager.java
```diff
@@ -0,0 +1,158 @@
...
+
+    ListOffsetsResultInfo offsetInfo = map.get(topicPartition);
+    if (offsetInfo == null) {
+      return -1;
+    }
+
+    return offsetInfo.offset();
+  }
+
+  protected final <T extends Map> T checkOffsetsExist(T offsets, String message, Object... args) {
```
rigelbm, line 120: `<T extends Map>` is probably generating a warning in your IDE, which means you probably should not be doing it :). Use `<T extends Map<?, ?>>` instead.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerGroupLagManagerImpl.java
```diff
@@ -0,0 +1,118 @@
...
+                                .thenApply(
+                                    latestOffsets ->
+                                       createConsumerGroupLag(
+                                           clusterId,
+                                           consumerGroup,
+                                           fetchedCurrentOffsets,
+                                           latestOffsets))));
+  }
+
+  Optional<ConsumerGroupLag> createConsumerGroupLag(
```
rigelbm, line 75: This method never returns `Optional.empty()`, so it should probably returns `ConsumerGroupLag` instead of `Optional<ConsumerGroupLag>`. Also, it should be private and static.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerGroupLagManagerImpl.java
```diff
@@ -0,0 +1,118 @@
...
+      Map<TopicPartition, ListOffsetsResultInfo> latestOffsets) {
+    Map<TopicPartition, MemberId> tpMemberIds =
+        getMemberIds(consumerGroup);
+    ConsumerGroupLag.Builder consumerGroupLag =
+        ConsumerGroupLag.builder()
+            .setClusterId(clusterId)
+            .setConsumerGroupId(consumerGroup.getConsumerGroupId());
+    fetchedCurrentOffsets.keySet().stream().forEach(
+        topicPartition -> {
+          MemberId memberId = tpMemberIds.getOrDefault(
```
rigelbm, line 88: Instead of creating a dummy here, you should keep this optional, and use `map` below, where you use it. E.g.:


```
Optional<MemberId> memberId = Optional.ofNullable(tpMemberIds.get(topicPartition));
...
consumerGroupLag.addOffset(
    topicPartition.topic(),
    memberId.map(MemberId::getConsumerId).orElse(""),
    memberId.map(MemberId::getClientId).orElse(""),
    memberId.map(MemberId::getInstanceId).orElse(null),
    topicPartition.partition(),
    currentOffset,
    latestOffset);
```

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ConsumerGroupLagManagerImpl.java
```diff
@@ -0,0 +1,118 @@
...
+                topicPartition.partition(),
+                memberId.getConsumerId(),
+                currentOffset,
+                latestOffset);
+          } else {
+            consumerGroupLag.addOffset(
+                topicPartition.topic(),
+                memberId.getConsumerId(),
+                memberId.getClientId(),
+                memberId.getInstanceId().orElseGet(null),
```
rigelbm, line 110: This will cause a `NullPointerException`. `orElseGet` receives a `Supplier<T>`, to which the Optional will call `get()` to obtain the value to be used. Here you are passing a null `Supplier`, so `supplier.get()` will throw `NullPointerException`. What you probably want is `orElse(null)`.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AbstractConsumerLagManager.java
ahuang98: checked both can't return null

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AbstractConsumerLagManager.java
ahuang98: there was a check for -1 following the calls which corresponded to the "no offset" situation, but returning Optional.empty() is much clearer, thanks!

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
ahuang98: Actually, from addressing https://github.com/confluentinc/kafka-rest/pull/784#discussion_r564433693 
I think it might be better to let setInstanceId accept an Optional instead of @Nullable.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
ahuang98: discussed with Jason, sticking with non-negative lags

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AbstractConsumerLagManager.java
```diff
@@ -0,0 +1,158 @@
...
+      for (Partition partition : consumer.getAssignedPartitions()) {
+        MemberId memberId =
+            MemberId.builder()
+                .setConsumerId(consumer.getConsumerId())
+                .setClientId(consumer.getClientId())
+                .setInstanceId(consumer.getInstanceId().orElse(null))
+                .build();
+        TopicPartition topicPartition =
+            new TopicPartition(partition.getTopicName(), partition.getPartitionId());
+        tpMemberIds.put(topicPartition, memberId);
```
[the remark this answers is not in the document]
ahuang98: Created `Partition` objects with empty replica field

api/v3/openapi.yaml
```diff
@@ -322,6 +322,50 @@ paths:
...
+    get:
+      summary: 'Get Consumer Lag'
+      description: 'Returns the consumer lag on a partition with the given `partition_id`.'
+      tags:
+        - Partition
+      responses:
+        '200':
+          $ref: '#/components/responses/GetConsumerLagResponse'
+
+  /clusters/{cluster_id}/consumer-groups/{consumer_group_id}/lags:
```
rigelbm, line 355: You should move this endpoint before `/clusters/{cluster_id}/consumer-groups/{consumer_group_id}/lags/{topic_name}/partitions/{partition_id}` to keep them lexicographically sorted.

checkstyle/suppressions.xml
```diff
@@ -18,7 +18,7 @@
 
     <suppress
             checks="(ClassDataAbstractionCoupling)"
-            files="(AvroConverter|ConsumerManager|KafkaConsumerManager|KafkaRestApplication|SimpleConsumerManager|KafkaRestContextProvider|KafkaConsumerState|ProducerPool|TopicsResource).java"
+            files="(AvroConverter|ConfigModule|ConsumerManager|KafkaConsumerManager|KafkaRestApplication|SimpleConsumerManager|KafkaRestContextProvider|KafkaConsumerState|ProducerPool|TopicsResource).java"
```
rigelbm, line 21: I don't think you need this anymore.

checkstyle/suppressions.xml
```diff
@@ -41,6 +41,6 @@
     />
     <suppress
             checks="(ParameterNumber)"
-            files="(ClusterData).java"
+            files="(ClusterData|ConsumerGroupLagData|ConsumerLagData).java"
```
rigelbm, line 44: You can disable this check in-place in the code, using comments, e.g.:

```
// CHECKSTYLE:OFF:ParameterNumber
void myMethodWithALotOfParameters(int a, int b, int c, int d, int e, int f, int g, int h, int i, int j) {
}
// CHECKSTYLE:ON:ParameterNumber
```

It's easier to maintain that than these suppressions in this xml file.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AbstractConsumerLagManager.java
```diff
@@ -0,0 +1,100 @@
...
+
+abstract class AbstractConsumerLagManager {
+
+  private final Admin kafkaAdminClient;
+  private static final IsolationLevel ISOLATION_LEVEL = IsolationLevel.READ_COMMITTED;
+
+  AbstractConsumerLagManager(
+      Admin kafkaAdminClient
+  ) {
+    this.kafkaAdminClient = kafkaAdminClient;
```
rigelbm, line 42: requireNonNull

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroup.java
```diff
@@ -42,6 +45,16 @@
 
   public abstract ImmutableList<Consumer> getConsumers();
 
+  public final ImmutableMap<Partition, Consumer> getPartitionAssignment() {
+    Map<Partition, Consumer> partitionAssignment = new HashMap<>();
```
rigelbm, line 49: You can use `ImmutableMap.Builder` here if you want to avoid the cost of copying the map later.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerLag.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public abstract String getTopicName();
+
+  public abstract int getPartitionId();
+
+  public abstract Long getCurrentOffset();
+
+  public abstract Long getLogEndOffset();
+
+  public Long getLag() {
```
rigelbm, line 47: final

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroupLag.java
```diff
@@ -0,0 +1,102 @@
...
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    private long maxLag = -1;
+    private long totalLag = 0;
+
+    Builder() {
+    }
+
+    public void addOffset(
```
rigelbm, line 58: final

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ConsumerLagsResource.java
```diff
@@ -0,0 +1,166 @@
...
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.container.AsyncResponse;
+import javax.ws.rs.container.Suspended;
+import javax.ws.rs.core.MediaType;
+
+
+@Path("/v3/clusters/{clusterId}/consumer-groups/{consumerGroupId}/lags")
+@ResourceName("api.v3.consumer-lags.*")
+public class ConsumerLagsResource {
```
rigelbm, line 50: final

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ConsumerLagsResource.java
```diff
@@ -0,0 +1,166 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.NotFoundException;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.container.AsyncResponse;
+import javax.ws.rs.container.Suspended;
+import javax.ws.rs.core.MediaType;
+
+
```
rigelbm, line 47: nit: Remove extra  \n

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ConsumerLagsResource.java
```diff
@@ -0,0 +1,166 @@
...
+      Provider<ConsumerLagManager> consumerLagManager,
+      CrnFactory crnFactory,
+      UrlFactory urlFactory
+  ) {
+    this.consumerLagManager = requireNonNull(consumerLagManager);
+    this.crnFactory = requireNonNull(crnFactory);
+    this.urlFactory = requireNonNull(urlFactory);
+  }
+
+
```
rigelbm, line 67: nit: Remove extra \n.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/AbstractConsumerLagManagerTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import java.util.Map;
+
+import static org.easymock.EasyMock.*;
+import static org.junit.Assert.assertEquals;
+
+public class AbstractConsumerLagManagerTest {
```
rigelbm, line 27: You don't need to test the abstract class, as long as you test all subclasses.

api/v3/openapi.yaml
```diff
@@ -1070,6 +1114,7 @@ components:
             - topic_name
             - partition_id
             - partition
+            - lag
```
rigelbm, line 1117: Should we also add a link to ConsumerGroupLag in ConsumerGroup?

rigelbm (approved):
Excellent! I'm glad to LGTM this PR! Very good job!

About the PR build: The project `schema-registry-multi-tenant` in repo `confluent-cloud-plugin` is failing (seems unrelated to your PR). I would let someone (in SR) know about it. We want a green build before merging.

Thanks @ahuang98!!

PR #787, opened 2021-01-07
Multiplex all produce requests into a single Producer.

rigelbm:
Background: Currently the way V2 produce works internally is that REST Proxy keeps one producer for each serialization format (BINARY, JSON, AVRO, JSONSCHEMA, PROTOBUF). Based on the format of the incoming request, it will decide to which producer to send the request to.

Problem: This works "well" for V2 because in V2, both key and value must use the same serialization format. I say "well" in quotes because it still requires maintaining 5 producers up, but that is still manageable. In V3, unfortunately, that won't do. One of the design differences of produce V3 is that it decouples the serialization format of key and values. If we were to follow the existing design, we would need to keep 25 (5x5) producers, which is not managleable.

Solution: This PR changes the produce internals to use a single producer instead, a byte-array producer. To achieve it, we change REST Proxy to serialize the records BEFORE sending them to the producer, instead of handing the producer the unserialized record for the producer itself to serialize.

Detailed Solution:

The existing flow looks something like:

1. `schema = getSchema(request) // this is only done for AVRO, JSONSCHEMA and PROTOBUF`
2. `producer = getProducer(request)`
3. `producer.producer(schema, request)`

The new flow looks like:

1. `schema = getSchema(request) // this is only done for AVRO, JSONSCHEMA and PROTOBUF`
2. `serialized = serializeRecord(schema, request)`
3. `binaryProducer.produce(serialized)`

Step (1) is handled by `SchemaManager`, step (2) is handled by `RecordSerializerFacade`, and step (3) is handled by `ProduceController`.

Testing: Great care has been taken to achieve exact same behaviour, since we are basically reimplementing the entire V2 produce API. All the integration tests (produce is arguably the most tested feature on REST Proxy) are largely unchanged. `SchemaManager`, `RecordSerializer` and `ProduceController` are throrougly unit-tested for parity with the now deleted `ProducerPoll`, `NoSchemaRestProducer` and `SchemaRestProducer`.


kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
```diff
@@ -27,32 +33,35 @@
...
-  private ProducerPool producerPool;
   private KafkaConsumerManager kafkaConsumerManager;
+
   private Admin adminClient;
+  private Producer<byte[], byte[]> producer;
 
+  /**
+   * @deprecated Use {@link #DefaultKafkaRestContext(KafkaRestConfig)} instead.
+   */
+  @Deprecated
```
mhowlett, line 44: why is this constructor still here?

kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
```diff
@@ -27,32 +33,35 @@
...
     return config;
   }
 
   @Override
-  public synchronized ProducerPool getProducerPool() {
-    if (producerPool == null) {
-      producerPool = new ProducerPool(config);
-    }
-    return producerPool;
+  public ProducerPool getProducerPool() {
```
mhowlett, line 63: it seems like you can get rid of this too.

kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
```diff
@@ -71,16 +80,34 @@ public synchronized Admin getAdmin() {
     return adminClient;
   }
 
+  @Override
+  public synchronized Producer<byte[], byte[]> getProducer() {
+    if (producer == null) {
+      producer =
```
mhowlett, line 86: side commentary: i don't like implicit `this`, but languages fault, and this seems to be the convention in this codebase.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -332,6 +335,12 @@
       "Whether to enable REST Proxy V3 API. Default is true.";
   private static final boolean API_V3_ENABLE_DEFAULT = true;
 
+  public static final String SCHEMA_CACHE_SPEC_CONFIG = "schema.cache.spec";
+  private static final String SCHEMA_CACHE_SPEC_DOC =
+      "Guava CacheBuilderSpec string for the schema cache. Default is"
+          + "'maximumSize=10000,expireAfterAccess=10m'";
+  private static final String SCHEMA_CACHE_SPEC_DEFAULT = "maximumSize=10000,expireAfterAccess=10m";
```
mhowlett, line 342: this is kind of leaking internal details (though not in a very serious way). i'd probably have two separate properties `SCHEMA_CACHE_MAX_SIZE_BYTES` and `SCHEMA_CACHE_EXPIRE_AFTER_MINUTES`.

mhowlett: i realize that is less flexible, but i don't that matters.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestContext.java
```diff
@@ -17,16 +17,25 @@
...
+import org.apache.kafka.clients.producer.Producer;
 
 public interface KafkaRestContext {
 
   KafkaRestConfig getConfig();
 
+  /**
+   * @deprecated Use {@link #getProducer()} instead.
+   */
+  @Deprecated
```
mhowlett, line 29: really needed?

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestContext.java
```diff
@@ -17,16 +17,25 @@
...
   ProducerPool getProducerPool();
 
   KafkaConsumerManager getKafkaConsumerManager();
 
   Admin getAdmin();
 
+  default Producer<byte[], byte[]> getProducer() {
```
mhowlett, line 36: default impl. really needed?

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerPool.java
```diff
@@ -15,189 +15,25 @@
...
+
+import org.apache.kafka.clients.producer.Producer;
 
 /**
- * Shared pool of Kafka producers used to send messages. The pool manages batched sends, tracking
- * all required acks for a batch and managing timeouts. Currently this pool only contains one
- * producer per serialization format (e.g. byte[], Avro).
+ * @deprecated This class only exists to satisfy the {@link KafkaRestContext} interface. It is soon
+ *             to be deleted. Access the producer directly, either via {@link
+ *             KafkaRestContext#getProducer()} or via injection of {@code Producer<byte[], byte[]>}.
```
mhowlett, line 25: it seems to me like you can get rid of it now

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/RecordSerializerFacade.java
```diff
@@ -0,0 +1,52 @@
...
+package io.confluent.kafkarest.controllers;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.protobuf.ByteString;
+import io.confluent.kafkarest.entities.EmbeddedFormat;
+import io.confluent.kafkarest.entities.RegisteredSchema;
+import java.util.Optional;
+
+/**
+ * A facade to serializers for all supported {@link EmbeddedFormat formats}.
```
mhowlett, line 25: v. minor: "A facade covering serializers of all supported ..." ?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -0,0 +1,348 @@
...
+    private RegisteredSchema parseSchema(String topicName, String rawSchema)
+        throws RestClientException, IOException {
+      ParsedSchema schema = schemaProvider.parseSchema(rawSchema, emptyList())
+          .orElseThrow(() -> new SerializationException("Error when parsing raw schema."));
+      int schemaId = serializer.register(subjectFunction.apply(topicName), schema);
+      return RegisteredSchema.create(schemaId, schema);
+    }
+  }
+
+  @AutoValue
```
mhowlett, line 266: fancy.

is `GetSchemaParams` a good name? maybe just `SchemaParams`?

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v2/ProduceRequest.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.validation.constraints.NotEmpty;
+
+@AutoValue
+public abstract class ProduceRequest {
+
+  @NotEmpty
+  @JsonProperty("records")
+  public abstract ImmutableList<ProduceRecord> getRecords();
+
+  @JsonProperty("key_schema_id")
```
mhowlett, line 35: are we going to loose the snake_case in the v3 api?

kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
rigelbm: This is still used in confluent-security-plugins. I will have another PR there (after this one is merged) to remove these references, than I can yet another PR here to remove these deprecated elements.

kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
rigelbm: Ditto confluent-security-plugins.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestContext.java
rigelbm: Ditto confluent-security-plugins.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestContext.java
rigelbm: Ditto confluent-security-plugins.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerPool.java
rigelbm: Ditto confluent-security-plugins.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v2/ProduceRequest.java
rigelbm: Not really. The Confluent API Style Guide mandates lower underscore JSON attributes.

mhowlett (approved):

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Obsolete.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
rigelbm: Obsolete.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/RecordSerializerFacade.java
rigelbm: Done.

rigelbm:
After investigating deeper into how the Kafka(Avro|JsonSchema|Protobuf)Serializer are implemented, I realised the V2 implementation that I basically mimicked was not very good, and that I could do it much simpler by using SchemaRegistryClient directly (instead of via the serializers). So I just had to rewrite SchemaManagerImpl entirely. Sorry for the extra burden. Can you PTAL? Happy news: the cache is gone (SchemaRegistryClient already has a internal cache).

kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
mhowlett: look forward to it :-)

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
mhowlett: great. if it's obsolete though, i guess the question becomes why is it still here?

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -709,6 +729,24 @@ public Properties getOriginalProperties() {
     return properties;
   }
 
+  public final Map<String, Object> getSchemaRegistryConfigs() {
+    HashMap<String, Object> possibleConfigs = new HashMap<>();
+    possibleConfigs.putAll(originals());
+    possibleConfigs.putAll(originalsWithPrefix("schema.registry.", /* strip= */ true));
+    possibleConfigs.putAll(originalsWithPrefix("client.", /* strip= */ true));
+    possibleConfigs.putAll(originalsWithPrefix("producer.", /* strip= */ true));
+    possibleConfigs.putAll(originalsWithPrefix("consumer.", /* strip= */ true));
```
mhowlett, line 738: is it the case that previously, you could have different SR related settings for producer / consumer which are now clobbered - effectively making this a breaking change?

if so, how likely is that? (i'm assuming not very, but haven't looked into it deeply).

this way of arriving at the config is pretty sloppy, hopefully in v3, you can constrain this so SR settings must be specified with a `schema.registry.` prefix.

mhowlett: rather than silently overwrite, maybe consider failing fast and throw an exception on any conflicts?

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -709,6 +729,24 @@ public Properties getOriginalProperties() {
...
+
+    // Disable auto-registration of schemas. Schema registration is handled by SchemaManager.
+    possibleConfigs.put(AUTO_REGISTER_SCHEMAS, false);
+    // Disable latest-version fetching of schemas.
+    possibleConfigs.put(USE_LATEST_VERSION, false);
```
mhowlett, line 743: again, seems a bit sloppy - allowing the user to specify SR config, then silently ignoring it.

also, why make it so the user can't turn off auto schema registration? surely that's a feature people would want? is it a breaking v2 change?

also, why arbitrarily disable `use.latest.version` capability? is this a breaking change from current v2 capability?

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -718,11 +756,17 @@ public Properties getProducerProperties() {
     addTelemetryReporterProperties(producerProps);
     producerProps.putAll(originalsWithPrefix("client.", /* strip= */ true));
     producerProps.putAll(originalsWithPrefix("producer.", /* strip= */ true));
-    producerProps.putAll(originalsWithPrefix("schema.registry", /* strip= */ false));
+    producerProps.putAll(getSchemaRegistryConfigs());
```
mhowlett, line 759: why does the producer need SR configs? I thought you were changing things to do serialization / SR interaction before producing.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -733,7 +777,7 @@ public Properties getConsumerProperties() {
     addTelemetryReporterProperties(consumerProps);
     consumerProps.putAll(originalsWithPrefix("client.", /* strip= */ true));
     consumerProps.putAll(originalsWithPrefix("consumer.", /* strip= */ true));
-    consumerProps.putAll(originalsWithPrefix("schema.registry", /* strip= */ false));
+    consumerProps.putAll(getSchemaRegistryConfigs());
```
mhowlett, line 780: same comment as producer.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Forgot to remove this one. :)

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: That's because of the very restrictive interface in `Serializer`. The `serialize` method only takes `topicName` and the object to serialize. In order to serialize a schema-ed record, you need more information. Namely: `subject`, `schemaId`, and for protobuf whether you are serializing a key or a value. The serializer needs to acquire all this information out of band, since it can't be passed to `serialize`. So it will use schema registry for that. `subject` will be inferred using the `subject.name.strategy` schema registry config., while the `schemaId` will be acquired by calling schema registry. And the `isKey` thing will come from `Serializer.configure(..., isKey)`.

I have re-implemented `RecordSerializer` now to reuse the `SchemaRegistryClient` registed in `SchemaRegistryModule`. And I am now using `AbstractKafkaSchemaSerDe`, which exposes a method `serializeImpl` that allows to passing the subject, schema and isKey directly. So now the only call to schema registry from the serializer should be to acquire `schemaId`, and that call will likely return from the schema registry cache, since we just did the same in SchemaManager.

Unfortunately there's no method that accepts `schemaId` directly. Otherwise we would have removed all calls to `SchemaRegistryClient` from serializer. Anyways, I hope that provides enough context.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: See comment on producer.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Read https://github.com/confluentinc/kafka-rest/pull/787#discussion_r558461865 first :)

This configurations don't make much sense in REST Proxy, because the behaviour controlled by them is entirely driven by the API. The REST Proxy behaviour has as been:
* if you specify `schema_id`, use the respective schema
* otherwise, use `schema` and we will auto-register it IF it doesn't exist (in the same subject) yet.

There's has never been a way of using latest version for a subject (in V2, although there will be in V3).

There have been many tickets of people setting these configurations, and then being surprised things don't work as expected. I think it is better to explicitly disable them and update the documentation.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Ow, yeah, now I understand what you mean. Yes, the producer does not need them. The problem is that I'm the producer configs to configure the serializers (which is what the producer does). Maybe I can separate them. Let me take a look :)

rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Consumer is unchanged because it has not been rewritten. The consumer still needs the schema registry configs.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
mhowlett: sounds good (if not a regression)

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -709,31 +725,111 @@ public Properties getOriginalProperties() {
...
+    possibleConfigs.put(USE_LATEST_VERSION, false);
+
+    return Maps.filterKeys(
+        possibleConfigs, AbstractKafkaSchemaSerDeConfig.baseConfigDef().names()::contains);
+  }
+
+  public final Map<String, Object> getJsonSerializerConfigs() {
+    HashMap<String, Object> serializerConfigs = new HashMap<>();
+    serializerConfigs.putAll(originalsWithPrefix("client.", /* strip= */ true));
+    serializerConfigs.putAll(originalsWithPrefix("producer.", /* strip= */ true));
```
mhowlett, line 752: the first three lines could be factored out and reused, but whatever, marginal, if better.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -709,31 +725,111 @@ public Properties getOriginalProperties() {
     return properties;
   }
 
+  public final Map<String, Object> getSchemaRegistryConfigs() {
+    HashMap<String, Object> possibleConfigs = new HashMap<>();
+
+    // Use a default schema.registry.url, if not specified.
+    possibleConfigs.put(SCHEMA_REGISTRY_URL_CONFIG, SCHEMA_REGISTRY_URL_DEFAULT);
```
mhowlett, line 732: i don't think it's a good idea to have a default for this, unless that is replicating existing behavior. if that's the case, consider depreciating for v3.

in particular, the presence of this property is what i'd use to determine if you instantiate a schema registry instance or not, i think. i guess i'll see how you do that.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
mhowlett: so yes, two messages: make sure you don't break anything important and fail fast if possible. i won't review this in depth unless you want me to, trust you to get it right.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -709,31 +725,111 @@ public Properties getOriginalProperties() {
...
+    serializerConfigs.putAll(getSchemaRegistryConfigs());
+
+    Set<String> names =
+        Sets.union(
+            AbstractKafkaSchemaSerDeConfig.baseConfigDef().names(),
+            ImmutableSet.of(
+                KafkaJsonSchemaSerializerConfig.FAIL_INVALID_SCHEMA,
+                KafkaJsonSchemaSerializerConfig.SCHEMA_SPEC_VERSION,
+                KafkaJsonSchemaSerializerConfig.ONEOF_FOR_NULLABLES,
+                KafkaJsonSchemaSerializerConfig.JSON_INDENT_OUTPUT));
```
mhowlett, line 783: nit: you could define this set in `KafkaJsonSchemaSerializerConfig`

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -709,31 +725,111 @@ public Properties getOriginalProperties() {
...
+  public final Map<String, Object> getProtobufSerializerConfigs() {
+    HashMap<String, Object> serializerConfigs = new HashMap<>();
+    serializerConfigs.putAll(originalsWithPrefix("client.", /* strip= */ true));
+    serializerConfigs.putAll(originalsWithPrefix("producer.", /* strip= */ true));
+    serializerConfigs.putAll(getSchemaRegistryConfigs());
+
+    Set<String> names =
+        Sets.union(
+            AbstractKafkaSchemaSerDeConfig.baseConfigDef().names(),
+            singleton(KafkaProtobufSerializerConfig.REFERENCE_SUBJECT_NAME_STRATEGY_CONFIG));
```
mhowlett, line 797: again, there's only one property here, but it would be structurally better to define it in a collection in `KafkaProtobufSerializerConfig`

kafka-rest/src/main/java/io/confluent/kafkarest/backends/schemaregistry/SchemaRegistryModule.java
```diff
@@ -0,0 +1,85 @@
...
+        @SchemaRegistryConfigs Map<String, Object> configs) {
+      this.schemaRegistryUrls = requireNonNull(schemaRegistryUrls);
+      this.maxSchemasPerSubject = maxSchemasPerSubject;
+      this.requestHeaders = requireNonNull(requestHeaders);
+      this.configs = requireNonNull(configs);
+    }
+
+    @Override
+    public SchemaRegistryClient provide() {
+      return new CachedSchemaRegistryClient(
```
mhowlett, line 73: is this ever constructed if it's not needed (i don't know, haven't checked, just asking)

kafka-rest/src/main/java/io/confluent/kafkarest/config/SchemaRegistryConfig.java
```diff
@@ -0,0 +1,64 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import io.confluent.kafka.schemaregistry.ParsedSchema;
+import io.confluent.kafka.serializers.AbstractKafkaSchemaSerDeConfig;
+import io.confluent.kafka.serializers.subject.strategy.SubjectNameStrategy;
+import java.util.Map;
+import org.apache.kafka.common.config.ConfigDef;
+
+final class SchemaRegistryConfig extends AbstractKafkaSchemaSerDeConfig {
```
mhowlett, line 26: doesn't seem like an inheritance relationship is logically correct here

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -0,0 +1,90 @@
...
+import javax.inject.Inject;
+import org.apache.kafka.common.errors.SerializationException;
+
+final class SchemaManagerImpl implements SchemaManager {
+  private final SchemaRegistryClient schemaRegistryClient;
+  private final SubjectNameStrategy subjectNameStrategy;
+
+  @Inject
+  SchemaManagerImpl(
+      SchemaRegistryClient schemaRegistryClient, SubjectNameStrategy subjectNameStrategy) {
```
mhowlett, line 38: for full generality, we'd want the ability to have more than one subject name strategy for keys or values i guess. you'd need a fancy way to specify that though, and RP would be the first place this requirement has come up. I think you could tack this in later if need be, so all good.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: We actually don't need it in my new implementation, but the old implementation needed it, so now we are stuck on having this default value for `schema.registry.url` because there might be people out there relying on it.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Barking in the wrong tree. I would also love for this set to be defined in `KafkaJsonSchemaSerializerConfig`. That's something to changed in schema-registry, not here.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Ditto.

kafka-rest/src/main/java/io/confluent/kafkarest/config/SchemaRegistryConfig.java
rigelbm: That's more a fault of `AbstractKafkaSchemaSerDeConfig` which should really be named `SchemaRegistryConfig` and not be abstract. That class only defines the SR configs, and it is what is used internally in `SchemaRegistryClient`.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
rigelbm: Not as configuration though. So, my plan is that the configuration will only define the default strategy, in case you don't specify one, and let people specify the strategy per request. That will only be available in V3, of course. Currently, V2 does not even support specifying the one default strategy, so V2 is already winning here.

kafka-rest/src/main/java/io/confluent/kafkarest/backends/schemaregistry/SchemaRegistryModule.java
rigelbm: Not anymore: https://github.com/confluentinc/kafka-rest/pull/787/commits/b79b7b18caeca266822068391f4722437c337917

Done.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: You are right, you could have before the producer and the consumer using different schema registries, so yeah, that makes it breaking. I believe that has never been the intention, but rather an artefact of not checking the configs properly. I also don't think this a feature, but rather a bug. I can't tell how often this happen in the wild, but I can tell you whoever is doing it shouldn't. They will get a warning now. The `consumer.` SR will be preferred over the `producer.` for no special reason.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -709,31 +733,129 @@ public Properties getOriginalProperties() {
     return properties;
   }
 
+  public final Map<String, Object> getSchemaRegistryConfigs() {
+    Set<String> mask = AbstractKafkaSchemaSerDeConfig.baseConfigDef().names();
+    Map<String, Object> configs =
+        new HashMap<>(
+            new ConfigsBuilder(mask)
+                .addConfigs("schema.registry.", /* strip= */ false)
```
mhowlett, line 741: is this because of `schema.registry.url`? kinda sucks, i guess this is fine.

mhowlett: it might be worth commenting on the reason.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -773,4 +895,70 @@ private void addMetricsReporters(Properties props) {
...
+      Map<String, ConfigValue> toAdd =
+          Maps.filterKeys(originalsWithPrefix(prefix, strip), mask::contains)
+              .entrySet()
+              .stream()
+              .collect(
+                  Collectors.toMap(
+                      Entry::getKey,
+                      entry ->
+                          ConfigValue.create(
+                              (strip ? prefix : "") + entry.getKey(), entry.getValue())));
```
mhowlett, line 917: this is probably right, i can't be bothered thinking about it, but just want to note that when strip is true, you're adding the prefix, which seems backwards, just in case you've made a mistake.

kafka-rest/src/main/java/io/confluent/kafkarest/config/SchemaRegistryConfig.java
mhowlett: 🤦

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
mhowlett: yeah, seems a good plan 👍

mhowlett (approved):

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: The first argument, `origin`, tracks the original configuration name, as specified by the user. So if we did strip the prefix (strip = true), we have to re-append it to get the original name back. It looks a little backwards, I know, but it's doing the right thing haha.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Done. Also restricted this to only `schema.registry.url` to avoid catching things we don't want to.

PR #792, opened 2021-01-20
KREST-456 Add API for creating a topic with custom partition-to-replicas assignments

dimitarndimitrov:
This change adds the ability to create a topic with a custom, explicitly specified mapping of partition to broker replicas.

The approach taken is the path of least resistance - adding yet another parameter to the `TopicManager` topic creation API, and not that far from the approach taken in the Kafka Admin client where the `NewTopic` spec contains both a field for such mappings, and fields for number of partitions and replication factor.
- Alternatively we could have introduced a DTO/spec object acting akin to `NewTopic` in the Kafka Admin client. This would have kept the `TopicManager` API more stable, but it would have added yet another class representing some form of a topic (alongside `CreateTopicRequest`, `Topic`, `TopicData`, `NewTopic`).  Having in mind that `TopicManager` is an internal abstraction, and not an actual API, I decided against that.
- Also we could have added a method overload in the `TopicManager`, but I didn't like how this would have meant slightly more changes in `TopicsResource` (in order to determine which overload to call after processing a topic creation request), so I decided against that too.

The change includes a corresponding update to the OpenAPI spec, although I'd appreciate a specifically close look at that part of the change.
It also includes some slight refactoring related to tests and checkstyle (and not the new API).

Aside from running tests locally it was also tested with the minimal Docker setup + simple cURL calls - e.g. 
```
curl -X POST -H "Content-Type: application/json" -d '{"topic_name":"ddimitrov","replicas_assignments":{"0":[1,2],"1":[2,3],"2":[3,1]}}' http://localhost:9391/v3/clusters/WSk_4zf7S0a-tLCF8QFCuQ/topics
```
- One interesting finding there is that apparently the new API does require the same number of partitions for all replicas, contrary to what the Kafka Admin client Javadoc claims.

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -1501,16 +1501,16 @@ components:
               value:
                 topic_name: 'topic-Y'
                 replicas_assignments:
-                  - partition: 0
-                    replicas:
+                  - partition_id: 0
+                    broker_ids:
                       - 1
                       - 2
-                  - partition: 1
-                    replicas:
+                  - partition_id: 1
+                    broker_ids:
                       - 2
                       - 3
-                  - partition: 2
-                    replicas:
+                  - partition_id: 2
+                    broker_ids:
                       - 3
                       - 1
                 configs:
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -1011,8 +1011,6 @@ components:
       type: object
       required:
         - topic_name
-        - partitions_count
-        - replication_factor
       properties:
         topic_name:
           type: string
@@ -1020,6 +1018,20 @@ components:
           type: integer
         replication_factor:
           type: integer
+        replicas_assignments:
+          type: array
+          items:
+            type: object
+            required:
+              - partition
+              - replicas
+            properties:
+              partition:
+                type: integer
+              replicas:
+                type: array
+                items:
+                  type: integer
         configs:
           type: array
           items:
@@ -1474,15 +1486,38 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/CreateTopicRequestData'
-          example:
-            topic_name: 'topic-X'
-            partitions_count: 64
-            replication_factor: 3
-            configs:
-              - name: 'cleanup.policy'
-                value: 'compact'
-              - name: 'compression.type'
-                value: 'gzip'
+          examples:
+            uniform_replication:
+              value:
+                topic_name: 'topic-X'
+                partitions_count: 64
+                replication_factor: 3
+                configs:
+                  - name: 'cleanup.policy'
+                    value: 'compact'
+                  - name: 'compression.type'
+                    value: 'gzip'
+            explicit_replicas_assignments:
+              value:
+                topic_name: 'topic-Y'
+                replicas_assignments:
+                  - partition: 0
+                    replicas:
+                      - 1
+                      - 2
+                  - partition: 1
+                    replicas:
+                      - 2
+                      - 3
+                  - partition: 2
+                    replicas:
+                      - 3
+                      - 1
+                configs:
+                  - name: 'cleanup.policy'
+                    value: 'compact'
+                  - name: 'compression.type'
+                    value: 'gzip'
 
     UpdateBrokerConfigRequest:
       description: 'The broker configuration parameter update request.'
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -1023,12 +1023,12 @@ components:
           items:
             type: object
             required:
-              - partition
-              - replicas
+              - partition_id
+              - broker_ids
             properties:
-              partition:
+              partition_id:
                 type: integer
-              replicas:
+              broker_ids:
                 type: array
                 items:
                   type: integer
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import javax.annotation.Nullable;
 
 /**
  * A service to manage Kafka {@link Topic Topics}.
@@ -51,13 +52,15 @@ public interface TopicManager {
   CompletableFuture<Optional<Topic>> getLocalTopic(String topicName);
 
   /**
-   * Creates a new Kafka {@link Topic}.
+   * Creates a new Kafka {@link Topic} with either partitions count and replication factor
+   * or explicitly specified partition-to-replicas assignments.
    */
   CompletableFuture<Void> createTopic(
       String clusterId,
       String topicName,
       Optional<Integer> partitionsCount,
       Optional<Short> replicationFactor,
+      @Nullable Map<Integer, List<Integer>> replicasAssignments,
       Map<String, Optional<String>> configs);
 
   /**
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
@@ -21,7 +21,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import javax.annotation.Nullable;
 
 /**
  * A service to manage Kafka {@link Topic Topics}.
@@ -60,7 +59,7 @@ public interface TopicManager {
       String topicName,
       Optional<Integer> partitionsCount,
       Optional<Short> replicationFactor,
-      @Nullable Map<Integer, List<Integer>> replicasAssignments,
+      Map<Integer, List<Integer>> replicasAssignments,
       Map<String, Optional<String>> configs);
 
   /**
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/CreateTopicRequest.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/CreateTopicRequest.java
@@ -23,7 +23,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 @AutoValue
@@ -83,8 +82,7 @@ public abstract class CreateTopicRequest {
 
     public abstract Builder setReplicationFactor(@Nullable Short replicationFactor);
 
-    public abstract Builder setReplicasAssignments(
-        @Nonnull Map<Integer, List<Integer>> replicasAssignments);
+    public abstract Builder setReplicasAssignments(Map<Integer, List<Integer>> replicasAssignments);
 
     public abstract Builder setConfigs(List<ConfigEntry> configs);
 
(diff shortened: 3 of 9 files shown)


dimitarndimitrov:
[clabot:check]

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
```diff
@@ -51,13 +52,15 @@
...
-   * Creates a new Kafka {@link Topic}.
+   * Creates a new Kafka {@link Topic} with either partitions count and replication factor
+   * or explicitly specified partition-to-replicas assignments.
    */
   CompletableFuture<Void> createTopic(
       String clusterId,
       String topicName,
       Optional<Integer> partitionsCount,
       Optional<Short> replicationFactor,
+      @Nullable Map<Integer, List<Integer>> replicasAssignments,
```
rigelbm, line 63: Not nullable. Use empty map for "no replica assignment".

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/CreateTopicRequest.java
```diff
@@ -38,6 +39,10 @@
   @JsonProperty("replication_factor")
   public abstract Optional<Short> getReplicationFactor();
 
+  @JsonProperty("replicas_assignments")
+  @Nullable
+  public abstract Map<Integer, List<Integer>> getReplicasAssignments();
```
rigelbm, line 44: Not nullable. Have it empty if not specified.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/CreateTopicRequest.java
```diff
@@ -48,14 +53,17 @@ public static Builder builder() {
   @JsonCreator
   static CreateTopicRequest fromJson(
       @JsonProperty("topic_name") String topicName,
-      @JsonProperty("partitions_count") @Nullable  Integer partitionsCount,
+      @JsonProperty("partitions_count") @Nullable Integer partitionsCount,
       @JsonProperty("replication_factor") @Nullable Short replicationFactor,
+      @SuppressWarnings("checkstyle:LineLength")
```
rigelbm, line 58: Don't suppress the warning. Break the line. I think breaking it anywhere is valid, either between the annotations, after the annotations or after the type.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/CreateTopicRequest.java
```diff
@@ -72,6 +80,9 @@ static CreateTopicRequest fromJson(
 
     public abstract Builder setReplicationFactor(@Nullable Short replicationFactor);
 
+    @SuppressWarnings("checkstyle:LineLength")
+    public abstract Builder setReplicasAssignments(@Nullable Map<Integer, List<Integer>> replicasAssignments);
```
rigelbm, line 84: Break the line on params start:

```
public abstract Builder setReplicasAssignments(
    @Nullable Map<Integer, List<Integer>> replicasAssignments);
```

We have a strict 100-cols policy in REST Proxy.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
```diff
@@ -137,24 +138,37 @@ public void createTopic(
...
             .stream()
             .collect(Collectors.toMap(ConfigEntry::getName, ConfigEntry::getValue));
 
+    // We have no way of knowing the default replication factor in the Kafka broker.
+    short rf = replicationFactor.orElse((short) 0);
+    if (replicasAssignments != null) {
+      // In case of explicitly specified partition-to-replicas assignments, all partitions should
+      // have the same number of replicas.
+      rf = (short) replicasAssignments.values().iterator().next().size();
+    }
```
rigelbm, line 153: You can have rf being non mutable if you chain it:

```
replicationFactor.orElse(replicasAssignments.values().stream().map(List::size).findAny().orElse(0))
```

We avoid mutables whenever we can. Also, don't use abbreviations. Prefer full names.

api/v3/openapi.yaml
```diff
@@ -1011,15 +1011,27 @@ components:
...
-        - partitions_count
-        - replication_factor
       properties:
         topic_name:
           type: string
         partitions_count:
           type: integer
         replication_factor:
           type: integer
+        replicas_assignments:
```
rigelbm, line 1021: nit: replica_assignments

api/v3/openapi.yaml
```diff
@@ -1011,15 +1011,27 @@ components:
...
+          type: array
+          items:
+            type: object
+            required:
+              - partition
+              - replicas
+            properties:
+              partition:
```
rigelbm, line 1029: Whenever we are talking about an identifier in the REST Proxy API, we are explicit about it. So here you should use `partition_id` instead of `partition`. If you look at the API you will notice that whenever `partition` is used, that is referring to the `Partition` resource, i.e., it will be a link to the partition.

api/v3/openapi.yaml
```diff
@@ -1011,15 +1011,27 @@ components:
...
+                type: integer
+              replicas:
```
rigelbm, line 1031: Same here. Since these won't be links to the replicas, then you should use the identifier name instead. For replicas, we use `broker_id`, so this here should be `broker_ids`.

rigelbm (commented):
Very good! Just some minor comments. Thanks!

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
dimitarndimitrov: Yep, that was one of the things I was considering, but decided to go with a nullable, because that's what the Kafka Admin client does internally (they null-check to determine whether to apply assignments).

api/v3/openapi.yaml
dimitarndimitrov: Note that I've been using this somewhat cumbersome wording everywhere - that's because that's the wording `NewTopic` uses. I'd prefer to keep it as-is - I did notice that such a convention is not present or enforced (e.g. you're using `partitions_count`/`partitionsCount` and not `numPartitions`), but for a hard to name field like this it's the easy to pick choice.

api/v3/openapi.yaml
dimitarndimitrov: Then maybe also change `replicas` to `broker_id_list` or something similar?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
dimitarndimitrov: Unfortunately this expression goes through `Optional<Integer>` and in order to get back from `Integer` to `short` I need to explicitly unbox then do a narrowing conversion. The result is a mouthful, and I'm not sure the much worse readability is worth the `final` here. I replaced it with a ternary conditional, although it can also be wrestled to work with your suggestion. E.g.
```
final short assumedReplicationFactor = replicationFactor.orElse(
    replicasAssignments.isEmpty() ? 0 : (short) replicasAssignments.values().iterator().next().size());
```
vs
```
final short assumedReplicationFactor = replicationFactor.orElse(
        (short) ((int) (
            replicasAssignments.values().stream().map(List::size).findAny().orElse(0))));
```

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/CreateTopicRequest.java
```diff
@@ -38,24 +41,32 @@
...
+  @JsonProperty("replicas_assignments")
+  public abstract Map<Integer, List<Integer>> getReplicasAssignments();
+
   @JsonProperty("configs")
   public abstract ImmutableList<ConfigEntry> getConfigs();
 
   public static Builder builder() {
-    return new AutoValue_CreateTopicRequest.Builder();
+    return new AutoValue_CreateTopicRequest.Builder()
+        .setReplicasAssignments(Collections.emptyMap());
```
rigelbm, line 52: Do we need to default?

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/CreateTopicRequest.java
```diff
@@ -72,6 +83,9 @@ static CreateTopicRequest fromJson(
 
     public abstract Builder setReplicationFactor(@Nullable Short replicationFactor);
 
+    public abstract Builder setReplicasAssignments(
+        @Nonnull Map<Integer, List<Integer>> replicasAssignments);
```
rigelbm, line 87: We don't use `@Nonnull`. Everything that is not explicitly marked as `@Nullable` in the REST Proxy repo is to be assumed to be non-null.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
```diff
@@ -137,24 +138,38 @@ public void createTopic(
...
+    Map<Integer, List<Integer>> replicasAssignments = request.getReplicasAssignments();
     Map<String, Optional<String>> configs =
         request.getConfigs()
             .stream()
             .collect(Collectors.toMap(ConfigEntry::getName, ConfigEntry::getValue));
 
+    // We have no way of knowing the default replication factor in the Kafka broker. Also in case
+    // of explicitly specified partition-to-replicas assignments, all partitions should have the
+    // same number of replicas.
+    final short assumedReplicationFactor = replicationFactor.orElse(
```
rigelbm, line 150: We don't use `final` for local variables.

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/TopicsResourceTest.java
```diff
@@ -626,27 +542,50 @@ public void createTopic_nonExistingTopic_defaultReplicationFactor_createsTopic()
...
+        response,
+        TOPIC_1.getClusterId(),
+        CreateTopicRequest.builder()
+            .setTopicName(TOPIC_1.getName())
+            .setReplicasAssignments(replicasAssignments)
+            .setConfigs(
+                singletonList(CreateTopicRequest.ConfigEntry.create("cleanup.policy", "compact")))
+            .build());
+
+    short expectedRf = (short) (TOPIC_1.getReplicationFactor() - 1);
```
rigelbm, line 586: nit: avoid abbreviations.

rigelbm (approved):
Awesome! Thanks! Just some minor comments. LGTM.

api/v3/openapi.yaml
```diff
@@ -1474,15 +1486,38 @@ components:
...
+                configs:
+                  - name: 'cleanup.policy'
+                    value: 'compact'
+                  - name: 'compression.type'
+                    value: 'gzip'
+            explicit_replicas_assignments:
+              value:
+                topic_name: 'topic-Y'
+                replicas_assignments:
+                  - partition: 0
```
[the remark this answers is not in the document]
dimitarndimitrov: Good catch, fixed.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/CreateTopicRequest.java
dimitarndimitrov: Fixed.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
dimitarndimitrov: Fixed.

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/TopicsResourceTest.java
dimitarndimitrov: Fixed.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/CreateTopicRequest.java
dimitarndimitrov: Yes, unfortunately - otherwise the `AutoValue` validation complains that a field that's neither `@Nullable`, nor `Optional` has _not_ been initialized.

PR #794, opened 2021-01-22
Allow setting the request contentType in testProduceToAuthorizationError

dimitarndimitrov:
This enables correctly running some tests in `confluent-security-plugins/confluent-kafka-rest-security-plugin` (PR to follow soon).

_Nota bene_: I'd like to avoid keeping the possibly redundant `(String, TopicRequestT, Map<String, String>)` overload, so hopefully the PR CI will check and notify me if a downstream dependency breaks due to that overload's removal. If CI is green, I'll add reviewers for this PR.


rigelbm (approved):
LGTM

PR #795, opened 2021-01-26
KREST-491 Test with exponential backoff retries racy checks causing flaky failures in TopicsResourceIntegrationTest

dimitarndimitrov:
Adds a simple test retry utility that should help with inherently racy test procedures that nevertheless are expected to pass if given a little bit of slack.

One example of a procedure like that is topic creation/deletion immediately followed by checking for the topic's presence/absence in the list of available topics.
- To me this sounded like a pretty bad case of not reading your own writes, but apparently this is a known issue related to reading from different data sources (broker vs ZK).

Before this change, when running locally on my machine, the flaky tests causing [KREST-491](https://confluentinc.atlassian.net/browse/KREST-491) and [KREST-496](https://confluentinc.atlassian.net/browse/KREST-496) were failing ~30 times out of 200 total test runs. After this change, they were failing 0 times out of 200 total test runs.
- Tested by keeping just the corresponding 2 tests in the test file and running
```
rm ~/Desktop/KREST-491.log && for i in {1..100}; do mvn -f kafka-rest/pom.xml -Dit.test=TopicsResourceIntegrationTest test-compile failsafe:integration-test >> ~/Desktop/KREST-491.log; done
```

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
@@ -53,7 +53,7 @@ public class TestUtils {
   private static final ObjectMapper jsonParser = new ObjectMapper();
 
   private static final int DEFAULT_EXP_BACKOFF_RETRIES = 3;
-  private static final long DEFAULT_INITIAL_BACKOFF_MS = 200;
+  private static final Duration DEFAULT_INITIAL_BACKOFF = Duration.ofMillis(200L);
 
   /**
    * Try to read the entity. If parsing fails, errors are rethrown, but the raw entity is also logged for debugging.
@@ -251,30 +251,28 @@ public class TestUtils {
         testCondition,
         errorMessage,
         DEFAULT_EXP_BACKOFF_RETRIES,
-        DEFAULT_INITIAL_BACKOFF_MS,
-        TimeUnit.MILLISECONDS);
+        DEFAULT_INITIAL_BACKOFF);
   }
 
   public static <T> void testWithRetry(
       Supplier<Boolean> testCondition,
       String errorMessage,
       int numRetries,
-      long initialBackoff,
-      TimeUnit backoffTimeUnit) {
-    long backoffInNs = backoffTimeUnit.toNanos(initialBackoff);
+      Duration initialBackoff) {
+    Duration backoff = initialBackoff;
     for (int i = 0; i <= numRetries; i++) {
       if (testCondition.get()) {
         return;
       }
-      LockSupport.parkNanos(backoffInNs);
-      backoffInNs *= 2;
+      LockSupport.parkNanos(backoff.toNanos());
+      backoff = backoff.multipliedBy(2L);
     }
     Assert.fail(
         String.format(
             "%s. Failed after %d exponential backoff retries with %d ms initial backoff",
             errorMessage,
             numRetries,
-            backoffTimeUnit.toMillis(initialBackoff)));
+            initialBackoff.toMillis()));
   }
 
   private static <V, K> Map<Object, Integer> topicCounts(final KafkaConsumer<K, V> consumer,
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
@@ -31,7 +31,10 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.LockSupport;
+import java.util.function.Supplier;
 import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.Response;
 import org.apache.avro.generic.IndexedRecord;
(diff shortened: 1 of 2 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
```diff
@@ -241,6 +246,37 @@ public static void assertPartitionOffsetsEqual(List<PartitionOffset> a, List<Par
...
+        DEFAULT_INITIAL_BACKOFF_MS,
+        TimeUnit.MILLISECONDS);
+  }
+
+  public static <T> void testWithRetry(
+      Supplier<Boolean> testCondition,
+      String errorMessage,
+      int numRetries,
+      long initialBackoff,
+      TimeUnit backoffTimeUnit) {
```
rigelbm, line 263: Can you use Duration instead?

kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
```diff
@@ -48,6 +52,9 @@
   private static final Logger log = LoggerFactory.getLogger(TestUtils.class);
   private static final ObjectMapper jsonParser = new ObjectMapper();
 
+  private static final int DEFAULT_EXP_BACKOFF_RETRIES = 3;
+  private static final long DEFAULT_INITIAL_BACKOFF_MS = 200;
```
rigelbm, line 56: Can you use Duration instead?

kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
```diff
@@ -241,6 +246,37 @@ public static void assertPartitionOffsetsEqual(List<PartitionOffset> a, List<Par
...
+        testCondition,
+        errorMessage,
+        DEFAULT_EXP_BACKOFF_RETRIES,
+        DEFAULT_INITIAL_BACKOFF_MS,
+        TimeUnit.MILLISECONDS);
+  }
+
+  public static <T> void testWithRetry(
+      Supplier<Boolean> testCondition,
+      String errorMessage,
```
rigelbm, line 260: Make errorMessage the last arg, and also accept error message args, so that the `String.format` is done inside this function: e.g.

```
testWithRetry(true, "my test is %s", "awesome");
```

rigelbm (approved):
LGTM! Very good!

kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
dimitarndimitrov: I'd prefer to avoid that for a couple of reasons.

First, varargs in multi-argument functions are something that almost never reads well - the only exceptions I can think of being SLF4J-style loggers and even they are practically always limited to less than 2 unrelated arguments.
Second, in order not to force users to pass `null` more often than they want, vararg methods require overloads, which I don't think we need here.
Third, if we want to get everything working with a single `String.format` call we will need to change the wording (leaving the vararg-ed template, which is the important part of the message, last), I think either escape the vararg template or concatenate it to the main template, and probably re-wrap the varargs anyway. Alternatively we can use two `String.format` calls, but I don't think that's any sort of win then.

So unless you're feeling strongly about this, can we avoid the varargs?

kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
dimitarndimitrov: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
dimitarndimitrov: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
rigelbm: I still think that the API would be cleaner as in:

```
  public static <T> void testWithRetry(
￼      Supplier<Boolean> testCondition,
      int numRetries,
      long initialBackoff,
      TimeUnit backoffTimeUnit,
      String errorMessage,
      Object... errorMessageArgs) {
    // ...
    Assert.fail(
        String.format(
            "%s. Failed after %d exponential backoff retries with %d ms initial backoff",
            String.format(errorMessage, errorMessageArgs),
            numRetries,
            backoffTimeUnit.toMillis(initialBackoff)));
```

(errorMessage, errorMessageArgs) is a common pattern, and it makes usage easier. But I'm not strong about it. Acknowledged.

dimitarndimitrov: I'd still prefer to take advantage of your ack, because for the purpose of this method I can easily see how the error message might not have arguments (my initial error messages were simple strings), I hate either passing `null` in such cases or providing an overload that hides the `null`, and I'm afraid of call-sites being
```
testWithRetry(
    () -> {
        <epic lambda body here>
    },
    /* numRetries= */ 5,
    /* initialBackoff= */ Duration.ofMillis(500),
    "This %s was not that %s but a third %s",
    thisObject,
    thatObject,
    thirdObject);
```

Thank you for the enjoyable discussion :)

rigelbm: > I hate either passing null in such cases or providing an overload that hides the null
Not important anymore, but be aware in the future that 0 arguments is a valid input a var-arg parameter.

`foo(String... bar)` can be called as just `foo()` (an empty array will be created and passed as argument.

PR #796, opened 2021-01-27
KREST-524 Fix a test failure caused by a HashSet used concurrently.

dimitarndimitrov:
Tested locally in IntelliJ IDEA and via the command line.
With both approaches the results were fairly consistent, and out of 100 runs of `DefaultKafkaRestContextTest` (so a total of 300 tests):
- **before** the change, there were consistently between 5 and 10 failures.
- **after** the change, there were consistently no failures.


rigelbm (approved):
LGTM

rnpridgeon (approved):
Thanks!

PR #797, opened 2021-02-05
Produce V3 API

rigelbm:
This PR introduces the Produce V3 API!!

In a nutshell, the API looks like this:

Requests:
```
{
    "partition_id": 1,                                                             // optional
    "headers": [                                                                    // optional
        { "name": "Header-1", "value": "SGVhZGVyLTE=" }, // value is optional, base64 encoded
        { "name": "Header-2", "value": "SGVhZGVyLTI=" }
    ],
    "key": {
        "type": "BINARY",       // optional
        "data": "Zm9vYmFy", // optional, base64 encoded
    },
    "value": {
        "type": "AVRO",
        "subject": "topic-1-key",                   // optional, mutual exclusive with subject_name_strategy
        "schema": "{\"type\":\"string\"}",    // optional, mutual exclusive with schema_id and schema_version
        "data": "foobar"                               // data can be anything, as long as it matches the schema
    },
    "timestamp": "2021-02-05T19:14:42Z", // optional, ISO 8601
}
// multiple messages can be concatenated together in a single stream of JSON objects
{
    "key": {
        "subject_name_strategy": "TOPIC",
        "schema_id": 1,                                     // no need for "type", we ask Schema Registry for it
        "data": 1000
    },
    "value": {
        // subject_name_strategy defaults to key.subject.name.strategy and value.subject.name.strategy configs
        "schema_version": 1,
        "data": { "foo": "bar" }
    }
}
{
    "key": {
        "schema_id": 1
        // data can be omitted
    },
    "value": {
        "schema_version": 1,
        "data": null
    }
}
{
    "key": {
        // if no schema, schema_id or schema_version is given, the latest schema for the subject is assumed
        "data": 1000
    },
    // both key and value are entirely optional
}
```

Responses:
```
{
    "cluster_id": "cluster-1",
    "topic_name": "topic-1",
    "partition_id": 1,
    "offset": 0,
    "timestamp": "2021-02-05T19:14:42Z", // ISO 8601
    "key": {
        "type": "BINARY",
        "size": 7
    },
    "value": {
        "type": "AVRO",
        "subject": "topic-1-value",
        "schema_id": 1,
        "schema_version": 1,
        "size": 7
    }
}
// Responses are streamed as concatenated JSON in the same order as the requests
{
    "cluster_id": "cluster-1",
    "topic_name": "topic-1",
    "partition_id": 1,
    "offset": 1,
    "timestamp": "2021-02-05T19:14:43Z",
    "key": {
        "type": "JSONSCHEMA",
        "subject": "topic-1-key",
        "schema_id": 1,
        "schema_version": 1,
        "size": 7
    },
    "value": {
        "type": "AVRO",
        "subject": "topic-1-value",
        "schema_id": 1,
        "schema_version": 1,
        "size": 7
    }
}
{
    "cluster_id": "cluster-1",
    "topic_name": "topic-1",
    "partition_id": 1,
    "offset": 0,
    "timestamp": "2021-02-05T19:14:42Z", // ISO 8601
    "key": {
        "type": "JSONSCHEMA",
        "subject": "topic-1-key",
        "schema_id": 1,
        "schema_version": 1,
        "size": 7
    }
}
{
    "cluster_id": "cluster-1",
    "topic_name": "topic-1",
    "partition_id": 1,
    "offset": 0,
    "timestamp": "2021-02-05T19:14:42Z", // ISO 8601
    "key": {
        "type": "JSONSCHEMA",
        "subject": "topic-1-key",
        "schema_id": 1,
        "schema_version": 1,
        "size": 7
    }
}
```

The API will accept a single request, multiple requests concatenated, or a stream of requests, on the same endpoint. To send a stream of requests, you can use "Transfer-Encoding: chunked". The response will always use "Transfer-Encoding: chunked". It will stream back the responses in the order they appear on the requests, as soon as the response is available.

You can test streaming with curl by doing:

```
curl -X POST -H "Transfer-Encoding: chunked" -H "Content-Type: application/json" http://${hostname}/v3/clusters/${clusterId}/topics/${topicName}/records -T-
```

Beware that you might have to send extra empty chunks in order to force curl to the responses. Just press enter if you don't see your response immediately after sending a request.

Missing in this PR is the OpenAPI spec. I will update the PR next week with the spec for the API.


kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
```diff
@@ -0,0 +1,305 @@
...
+  public abstract static class ProduceRequestHeader {
+
+    ProduceRequestHeader() {
+    }
+
+    @JsonProperty("name")
+    public abstract String getName();
+
+    @JsonIgnore
+    public abstract Optional<ByteString> getValue();
```
mhowlett, line 116: using a protobuf type here seems a bit out of place to me (i could be missing a good reason for it)

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
```diff
@@ -0,0 +1,305 @@
...
+
+    @JsonProperty("value")
+    @JsonInclude(Include.NON_ABSENT)
+    final Optional<BinaryNode> getSerializedValue() {
+      return getValue().map(value -> BinaryNode.valueOf(value.toByteArray()));
```
mhowlett, line 121: header values are byte arrays, but I think a lot of the time people will use strings, and it would be convenient to allow for this with a type property. i guess fine to leave as possible future enhancement, though maybe it affects the types you use in this class. not that they couldn't be changed in the future as well.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
```diff
@@ -0,0 +1,305 @@
...
+
+@AutoValue
+public abstract class ProduceRequest {
+
+  ProduceRequest() {
+  }
+
+  @JsonProperty("partition_id")
+  @JsonInclude(Include.NON_ABSENT)
+  public abstract Optional<Integer> getPartitionId();
```
mhowlett, line 52: my instinct is this would be better specified in the url. i.e.:

`/v3/clusters/{clusterId}/topics/{topicName}/partitions/{partitionId}/records`

and not be specifiable in the JSON when producing at the topic level.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ProduceControllerImpl.java
```diff
@@ -53,7 +56,13 @@
             timestamp.toEpochMilli(),
             key.map(ByteString::toByteArray).orElse(null),
             value.map(ByteString::toByteArray).orElse(null),
-            /* headers= */ emptyList()),
+            headers.entries().stream()
+                .map(
+                    header ->
+                        new RecordHeader(
+                            header.getKey(),
+                            header.getValue().map(ByteString::toByteArray).orElse(null)))
```
mhowlett, line 64: noted elsewhere too - don't understand prevalent use of ByteString over byte array.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
rigelbm: `ByteString` is one of the best `immutable byte-array` types out there. (They can't make it part of Guava for legacy reasons). byte-array is not a class to start with, so it does not implement equals/hash-code properly. It also lacks reasonable constructors, like `copyFromUt8`. And, most importantly, it's mutable, which I limit as much as possible everywhere in Kafka REST.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
```diff
@@ -26,12 +27,42 @@
...
-  RegisteredSchema getSchemaById(EmbeddedFormat format, int schemaId);
+  RegisteredSchema getSchemaById(String subject, int schemaId);
 
   /**
-   * Parses and returns the potentially newly registered {@link RegisteredSchema schema}
-   * corresponding to {@code rawSchema}.
+   * Returns the {@link RegisteredSchema schema} registered with the given {@code schemaId}.
+   */
+  RegisteredSchema getSchemaById(
+      SubjectNameStrategy subjectNameStrategy, String topicName, boolean isKey, int schemaId);
```
mhowlett, line 36: why the subjectNameStrategy, topicName and isKey parameters on a method names this?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
```diff
@@ -26,12 +27,42 @@
...
+
+  /**
+   * Returns the {@link RegisteredSchema schema} registered with the given {@code schemaVersion}.
+   */
+  RegisteredSchema getSchemaByVersion(String subject, int schemaVersion);
+
+  /**
+   * Parses and returns {@code rawSchema}.
+   *
+   * <p>If the subject does not contain a registered schema equals to the parsed {@code rawSchema},
```
mhowlett, line 46: "equal to"

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
```diff
@@ -26,12 +27,42 @@
...
+   * one will be registered.
```
mhowlett, line 47: one will be registered if compatibility rules apply

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
```diff
@@ -26,12 +27,42 @@
...
+   */
+  RegisteredSchema parseSchema(EmbeddedFormat format, String subject, String rawSchema);
+
+  /**
+   * Parses and returns {@code rawSchema}.
+   *
+   * <p>If the subject does not contain a registered schema equals to the parsed {@code rawSchema},
```
mhowlett, line 54: equal to

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
```diff
@@ -26,12 +27,42 @@
...
+   * one will be registered.
```
mhowlett, line 55: if compatibility rules allow

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
```diff
@@ -26,12 +27,42 @@
...
    */
   RegisteredSchema parseSchema(
-      EmbeddedFormat format, String topicName, String rawSchema, boolean isKey);
+      EmbeddedFormat format,
+      SubjectNameStrategy subjectNameStrategy,
```
mhowlett, line 59: topic's the only valid value right? might be better to change the name of the method to be topic specific and remove this parameter. minor point though, internal detail.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
```diff
@@ -26,12 +27,42 @@
   /**
    * Returns the {@link RegisteredSchema schema} registered with the given {@code schemaId}.
    */
-  RegisteredSchema getSchemaById(EmbeddedFormat format, int schemaId);
+  RegisteredSchema getSchemaById(String subject, int schemaId);
```
mhowlett, line 30: schema:id is 1:1 - why do you need a subject? this looks wrong

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
```diff
@@ -26,12 +27,42 @@
...
+   */
+  RegisteredSchema getSchemaByVersion(String subject, int schemaVersion);
+
+  /**
+   * Parses and returns {@code rawSchema}.
+   *
+   * <p>If the subject does not contain a registered schema equals to the parsed {@code rawSchema},
+   * one will be registered.
+   */
+  RegisteredSchema parseSchema(EmbeddedFormat format, String subject, String rawSchema);
```
mhowlett, line 49: not into the name `parseSchema`, what about `registerOrLookupSchema`?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
mhowlett: hmm.. naming is awkward. maybe good as is, though maybe `registerOrLookupSchema()`

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -31,47 +33,101 @@
...
-    this.subjectNameStrategy = requireNonNull(subjectNameStrategy);
   }
 
   @Override
-  public RegisteredSchema getSchemaById(EmbeddedFormat format, int schemaId) {
-    checkArgument(format.requiresSchema(), "%s does not support schemas.", format);
+  public RegisteredSchema getSchemaById(String subject, int schemaId) {
+    ParsedSchema schema;
+    try {
+      schema = schemaRegistryClient.getSchemaBySubjectAndId(subject, schemaId);
```
mhowlett, line 46: whoa, i didn't know about this method, and don't understand why it exists. maybe change the name of your method to match `getSchemaBySubjectandIs`.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
```diff
@@ -26,12 +27,42 @@
...
+
+  /**
+   * Parses and returns {@code rawSchema}.
+   *
+   * <p>If the subject does not contain a registered schema equals to the parsed {@code rawSchema},
+   * one will be registered.
    */
   RegisteredSchema parseSchema(
```
mhowlett, line 57: `regsiterOrLookupSchema`?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -31,47 +33,101 @@
...
+    } catch (IOException | RestClientException e) {
+      throw new SerializationException("Error when fetching schema version.", e);
+    }
+
+    return RegisteredSchema.create(subject, schemaId, schemaVersion, schema);
+  }
 
+  @Override
+  public RegisteredSchema getSchemaById(
+      SubjectNameStrategy subjectNameStrategy, String topicName, boolean isKey, int schemaId) {
```
mhowlett, line 63: if all the parameters really are needed, maybe rename to just `getSchema`.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ProduceResult.java
```diff
@@ -28,11 +31,32 @@
...
+  public abstract Optional<Instant> getTimestamp();
+
+  public abstract int getSerializedKeySize();
+
+  public abstract int getSerializedValueSize();
+
+  public static ProduceResult create(
+      int partitionId,
+      long offset,
+      @Nullable Instant timestamp,
```
mhowlett, line 43: why nullable?

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
mhowlett: ok, those reasons sound sufficient ..

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceResponse.java
```diff
@@ -0,0 +1,183 @@
...
+
+  @JsonProperty("partition_id")
+  public abstract int getPartitionId();
+
+  @JsonProperty("offset")
+  public abstract long getOffset();
+
+  @JsonProperty("timestamp")
+  @JsonInclude(Include.NON_ABSENT)
+  public abstract Optional<Instant> getTimestamp();
```
mhowlett, line 48: wondering why this is optional

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -0,0 +1,331 @@
...
+    // allOf completion handles. For example, let's say we push 3 futures into the queue:
+    //
+    // 1. tail = (null) // empty queue
+    // 2. tail = then(allOf((null), f_1), write(f_1)) // -> f_1
+    // 3. tail = then(allOf(then(allOf((null), f_1), write(f_1)), f_2), write(f_2)) // -> f_1 -> f_2
+    //
+    // The chaining of the futures guarantees that the write(f) callbacks are called in the order
+    // that the futures got pushed into the queue. Also, due to the way allOf is implemented, it
+    // makes sure that, as soon as f_0..i is completed, the whole
+    // then(allOf((f_i-1), f_i), write(f_i)) monad is made available for garbage collection.
```
mhowlett, line 184: fancy. i haven't completely checked what's in this file for correctness, but it's at least approximately set up nicely and you're thinking about what needs to be thought about.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceResponse.java
dimitarndimitrov: My guess would be it's because this value more often than not comes from `RecordMetadata` [which is not always guaranteed to have a valid timestamp](https://github.com/apache/kafka/blob/a4c292173618e22b3f99ab8615c6b585c7688a40/clients/src/main/java/org/apache/kafka/clients/producer/RecordMetadata.java#L75-L81).

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/BadRequestException.java
```diff
@@ -0,0 +1,53 @@
...
+    this("Bad Request", detail, cause);
+  }
+
+  public BadRequestException(Integer code, String detail) {
+    this(code, "Bad Request", detail);
+  }
+
+  public BadRequestException(int code, String detail, Throwable cause) {
+    this(code, "Bad Request", detail, cause);
+  }
```
dimitarndimitrov, line 36: nit: Canonically _Bad Request_ is HTTP 400, so I wonder in which case would we need any of these ctors?

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/v2/ErrorResponse.java
```diff
@@ -15,58 +15,33 @@
...
+import com.google.auto.value.AutoValue;
 
 /**
  * Response returned when an exception happens in a V2 API.
  */
-public final class ErrorResponse {
-
-  private final String errorCode;
+@AutoValue
+public abstract class ErrorResponse {
```
dimitarndimitrov, line 26: I don't think I understand - why do we need to keep both of the identical v2 and v3 `ErrorResponse` classes? We're far from fully isolating the v2 and the v3 implementations and even interfaces, so why the duplication?

dimitarndimitrov (commented):
Partial review of all the source changes except `ProduceAction` and `StreamingResponse`.

The changes look pretty good so far, but I still haven't "ingested" the change as a whole - next I'll review the aforementioned 2 classes with any additional comments that might pop up, and after that I'll go through the tests - I'm planning to do that as soon as meaningful (often a better alternative to as soon as possible).

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -0,0 +1,331 @@
...
+              .thenApply(
+                  unused -> {
+                    try {
+                      sink.write(result.join());
+                    } catch (IOException e) {
+                      log.error("Error when writing streaming result to response channel.", e);
+                    }
+                    return null;
+                  });
+    }
```
dimitarndimitrov, line 208: Have you tested this manually, and can you please write a unit test for it?

Correct me if I'm wrong, but I'm concerned about the following:
1. A single request from the middle of a multi-request chunk getting an error during the produce handling on the broker side and therefore having its future (`result`) complete exceptionally.
2. The `CompletableFuture#allOf` tail wrapper corresponding to that `result` completing exceptionally too, and with that making any of the subsequent `CompletableFuture#allOf` tail wrappers to complete exceptionally too, even if the other half of what they wrap (their `result`s, the actual produce futures here) completes normally. 
3. The `thenApply` calls for these tail wrappers not being called (as `thenApply` is called for normal completions), despite their `result` containing a valid result that could be written back to the client.

In addition I don't think we're logging anything for the exceptions causing exceptional completion.

Maybe we should use `whenComplete` here, and in its body, specifically catching and logging for `CompletionException`s from `result.join()`?

dimitarndimitrov (commented):
Review of `ProduceAction` and `StreamingResponse`.
I think it will be great if we split the schema computation logic out of `ProduceAction` and if we add a unit test for `StreamingResponse`.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
dimitarndimitrov: I just noticed that the `handleNext` result is what is passed down as the `result` future here, so the above should not be a concern.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
rigelbm: I'm using bytes here to allow everything that the Java API allows, but ideally this would be ASCII encoded string, which is what is used for HTTP headers. We can extend/add an option later for string header values. I will skip it for now.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
rigelbm: I'm receiving partitionId in the body to allow production to multiple partitions using a single streaming connection. That would be the behaviour of the Java API. I do agree that we should add an specialization of the API that receives the partitionId in the URL. I will defer it to a future PR.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ProduceControllerImpl.java
rigelbm: Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
rigelbm: Although schema ID uniquely identifies the actual schema, we also return the schema version in `RegisteredSchema`. To acquire schema version, we need schema ID and subject. This also guarantees the subject has a version with this ID. Otherwise serialization would fail later.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
rigelbm: This guarantees the subject has a version with this ID. Otherwise serialization would fail later.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceResponse.java
rigelbm: Exactly. In practice we will always have a timestamp, since we always timestamp when producing. But since the Java API gives us a optional timestamp back, I have to propagate it, as there's no guarantees.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -0,0 +1,363 @@
...
+        produceController.get().produce(
+            clusterId,
+            topicName,
+            request.getPartitionId(),
+            request.getHeaders().stream().collect(PRODUCE_REQUEST_HEADER_COLLECTOR),
+            serializedKey,
+            serializedValue,
+            request.getTimestamp().orElse(Instant.now()));
+
+    return produceResult.thenApply(
```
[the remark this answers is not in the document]
rigelbm: 1. The ProduceController.produce API is non-blocking.
2. Preserving the order of input/output is the whole point of `StreamingResult`. 

Initially the plan was to force idempotent producer in REST Proxy to guarantee this ordering (idempotent producer will be the default for the Java API in Kafka 3 anyways). But I didn't want to force that, since it would make Kafka REST configuration more complex that what it already is. So I implemented the ordering outside the Producer, in `StreamingResult`.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
rigelbm: This is the bit that guarantees the ordering. i.e. `write` callback will be called in the order the futures are appended in the queue.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/NoSchemaRecordSerializer.java
```diff
@@ -58,11 +58,15 @@
   }
 
   private static ByteString serializeBinary(JsonNode data) {
+    if (!data.isTextual()) {
+      throw new BadRequestException(String.format("data=%s is not a base64 string.", data));
+    }
```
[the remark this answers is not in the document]
rigelbm: If using `JsonNode`, jackson will never map to `BinaryNode`, since it can't distinguish between a base64 encoded byte-array and a normal string. We use `JsonNode` because the request object is the same for all kinds of requests. We use `BinaryNode` on the places that accept only binary, e.g. `ProduceRequestHeader`.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/EmbeddedFormat.java
```diff
@@ -100,4 +100,16 @@ public SchemaProvider getSchemaProvider() {
...
+    if (schemaType.equals(AVRO.getSchemaProvider().schemaType())) {
+      return AVRO;
+    } else if (schemaType.equals(JSONSCHEMA.getSchemaProvider().schemaType())) {
+      return JSONSCHEMA;
+    } else if (schemaType.equals(PROTOBUF.getSchemaProvider().schemaType())) {
+      return PROTOBUF;
+    } else {
+      throw new IllegalArgumentException(String.format("Illegal schema type: %s", schemaType));
+    }
+  }
```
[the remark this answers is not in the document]
rigelbm: We want a `NullPointerException`. `schemaType` is not supposed to be `null`. If it is, we want to know.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
```diff
@@ -0,0 +1,305 @@
...
+        if (request.getFormat().isPresent()) {
+          if (request.getFormat().get().requiresSchema()) {
+            checkState(
+                !request.getSchemaId().isPresent(),
+                "'schema_id=%s' cannot be used with 'serializer'.",
+                request.getSchemaId().orElse(null));
+            checkState(
+                !request.getSchemaVersion().isPresent(),
+                "'schema_version=%s' cannot be used with 'serializer'.",
+                request.getSchemaVersion().orElse(null));
```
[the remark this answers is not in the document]
rigelbm: Except that this expression (`request.getSchemaVersion().orElse(null)` or `request.getSchemaVersion().get()`) is executed regardless of the value of the condition. So, if the condition is true, this would actually throw, even though the message would not be printed anyway.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
```diff
@@ -0,0 +1,305 @@
...
+                "'serializer=%s' cannot be used with 'schema_id'.",
+                request.getFormat().orElse(null));
+            checkState(
+                !request.getSchemaVersion().isPresent(),
+                "'serializer=%s' cannot be used with 'schema_version'.",
+                request.getFormat().orElse(null));
+            checkState(
+                !request.getRawSchema().isPresent(),
+                "'serializer=%s' cannot be used with 'schema'.",
+                request.getFormat().orElse(null));
```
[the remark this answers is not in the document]
rigelbm: As noted in the comment above, this is not `else` condition. For Java, this is just a normal function call. All expressions will be evaluated before the function is called.

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/BadRequestException.java
rigelbm: This code is not the HTTP code, but rather an application error code that we return in the response body. Users can later search for it in our documentation to understand the error better.

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/v2/ErrorResponse.java
rigelbm: Oh, but we are. The V2 and V3 API are completely independent, even if they use the same implementation behind the curtains. This is so that we can evolve the V3 error response format independently of V2. It still needs to be backwards compatible to the V2 error response format, but we can add fields in V3 that won't be returned in V3 if want (and we will, for example, for the Cluster Mirror API).

kafka-rest/src/main/java/io/confluent/kafkarest/entities/EmbeddedFormat.java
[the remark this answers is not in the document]
dimitarndimitrov: I think it would be nicer to handle it explicitly or have it fall-through to the `IllegalArgumentException` (which will have message _Illegal schema type: null_), instead of getting a generic `NullPointerException`, but I'm OK with this too.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -0,0 +1,363 @@
...
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.container.AsyncResponse;
+import javax.ws.rs.container.Suspended;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response.Status;
+
+@Path("/v3/clusters/{clusterId}/topics/{topicName}/records")
```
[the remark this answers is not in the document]
rigelbm: All V3 APIs use the same prefix `/v3/clusters/{clusterId}`, so we are tied to it. It would be highly inconsistent to diverge here. The `/topics/{topicName}` seems indispensable too, since we need to know the topic name. Up for discussion really is what comes after it, `/records`. I used only `/records` exactly to make the API shorter. But, if we were to be super consistent with everything else we have done for V3, this API should really be:

`POST /v3/clusters/{clusterId}/topics/{topicName}/partitions/-/records`

Why? Because I also want to add:

* `POST /v3/clusters/{clusterId}/topics/{topicName}/partitions/{partitionId}/records` to produce to a single partition (i.e. no partition_id field in the request body).
* `GET /v3/clusters/{clusterId}/topics/{topicName}/partitions/{partitionId}/records` to list records in a partition.
* `GET /v3/clusters/{clusterId}/topics/{topicName}/partitions/{partitionId}/records/{offset}` to get a single record from a partition.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -0,0 +1,363 @@
...
+      ImmutableMultimap<String, Optional<ByteString>>> PRODUCE_REQUEST_HEADER_COLLECTOR =
+      Collector.of(
+          ImmutableMultimap::builder,
+          (builder, header) -> builder.put(header.getName(), header.getValue()),
+          (left, right) -> left.putAll(right.build()),
+          ImmutableMultimap.Builder::build);
+
+  private final Provider<SchemaManager> schemaManager;
+  private final Provider<RecordSerializer> recordSerializer;
+  private final Provider<ProduceController> produceController;
```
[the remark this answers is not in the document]
rigelbm: If we could, we would also be able to just inject the instance directly instead of injection the provider :). TL;DR, the request handling pipeline currently looks something like:

URL-matching -> ctor called -> auth -> method called.

We can only call the provider after `auth`. We will fix this when we re-implement `auth` using dependency injection. We will only do that after we rewrite the consume API, just because it will be much easier then.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -0,0 +1,363 @@
...
+  public ProduceAction(
+      Provider<SchemaManager> schemaManager,
+      Provider<RecordSerializer> recordSerializer,
+      Provider<ProduceController> produceController,
+      SubjectNameStrategy defaultSubjectNameStrategy) {
+    this.schemaManager = requireNonNull(schemaManager);
+    this.recordSerializer = requireNonNull(recordSerializer);
+    this.produceController = requireNonNull(produceController);
+    this.defaultSubjectNameStrategy = requireNonNull(defaultSubjectNameStrategy);
+  }
```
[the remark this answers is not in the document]
rigelbm: Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/v2/ErrorResponse.java
dimitarndimitrov: If we do strive for complete independency I think we are on very thin ice with the copious amounts of shared code (and things like the `ServerProperties.OUTBOUND_CONTENT_LENGTH_BUFFER` flag mentioned above).
This looks like a separate discussion though.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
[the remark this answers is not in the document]
dimitarndimitrov: I don't think this is equivalent to the case above. The evaluation of these is safe, just as the `request.getFormat().get()` in the `if` condition before them is safe - they are all in another `if` that checks `if (request.getFormat().isPresent())`.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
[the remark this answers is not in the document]
dimitarndimitrov: I would have said that we should just add a lambda-using `checkState` here, but I checked and it turns out that `checkState` is a Guava method :) Given that this is a problem just here I won't bother you with asking for a wrapper.

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/BadRequestException.java
dimitarndimitrov: Two comments then:
1. Do we need these now though, if we are not using them anywhere?
2. If you think we need them, could you please add Javadoc for them explaining the application error bit?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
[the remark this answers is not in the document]
dimitarndimitrov: Thanks for unwrapping this for me (no pun intended)!
Can you please add a comment explaining that?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
rigelbm: Completely rewrote this bit. Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
rigelbm: Completely rewrote this bit. Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
rigelbm: Completely rewrote this bit. Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
rigelbm: Completely rewrote this bit. Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
rigelbm: Completely rewrote this bit. Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
rigelbm: Completely rewrote this bit. Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManager.java
rigelbm: Completely rewrote this bit. Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
rigelbm: Completely rewrote this bit. Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/ProduceResult.java
rigelbm: As pointed out somewhere else, timestamp is optionally returned by Kafka.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -31,47 +33,101 @@
...
 
   @Override
-  public RegisteredSchema getSchemaById(EmbeddedFormat format, int schemaId) {
-    checkArgument(format.requiresSchema(), "%s does not support schemas.", format);
+  public RegisteredSchema getSchemaById(String subject, int schemaId) {
+    ParsedSchema schema;
+    try {
+      schema = schemaRegistryClient.getSchemaBySubjectAndId(subject, schemaId);
+    } catch (IOException | RestClientException e) {
+      throw new SerializationException("Error when fetching schema by id.", e);
```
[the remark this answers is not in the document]
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -31,47 +33,101 @@
...
+    }
+
+    int schemaVersion;
+    try {
+      schemaVersion = schemaRegistryClient.getVersion(subject, schema);
+    } catch (IOException | RestClientException e) {
+      throw new SerializationException("Error when fetching schema version.", e);
```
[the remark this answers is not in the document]
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/EmbeddedFormat.java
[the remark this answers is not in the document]
rigelbm: Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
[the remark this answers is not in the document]
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceRequest.java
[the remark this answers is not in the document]
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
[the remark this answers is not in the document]
rigelbm: Not adding the comment since the is the usual/expected usage of Provider for injection.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -0,0 +1,363 @@
...
+            serializedValue,
+            request.getTimestamp().orElse(Instant.now()));
+
+    return produceResult.thenApply(
+        result ->
+            toProduceResponse(
+                clusterId, topicName, keyFormat, keySchema, valueFormat, valueSchema, result));
+  }
+
+  private Optional<RegisteredSchema> getSchema(
```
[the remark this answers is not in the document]
rigelbm: Moved to SchemaManager/SchemaManagerImp. Done.

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/BadRequestException.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -99,6 +100,7 @@ public void setupResources(Configurable<?> config, KafkaRestConfig appConfig) {
             contextInvocationHandler
         );
 
+    config.property(ServerProperties.OUTBOUND_CONTENT_LENGTH_BUFFER, 0);
```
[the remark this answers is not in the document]
rigelbm: This is for streaming (via chunked encoding) to work. Without this, the server won't send chunks back to users until this buffer is filled. I don't expect it to affect other APIs, but it's hard to tell. We will be able to tell better when we have the benchmarks.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  public static <T> StreamingResponse<T> from(MappingIterator<T> input) {
+    return new InputStreamingResponse<>(input);
+  }
+
+  public final <O> StreamingResponse<O> compose(
+      Function<? super T, ? extends CompletableFuture<O>> transform) {
+    return new ComposingStreamingResponse<>(this, transform);
+  }
```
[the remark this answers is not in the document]
rigelbm: > in mind that we don't need them separately
I don't see a way of combining them, actually. They are separate because we control the calls to `next()` which are synchronous, and call `transform`, which is async, on each result individually. Having the callers call `next()` on `MappingIterator` defeats some of the purpose of this class, which is to orchestrate the input and output of data. Callers should only care about transforming input to output. We could have a `from` that takes both the `input` and the `transform`, but we would not be gaining anything, really, since internally it would be the same. I think this way reads more naturally on the calling code, as it resembles what the Java `Stream` API offers.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  @POST
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Produces(MediaType.APPLICATION_JSON)
+  public void produce(
+      @Suspended AsyncResponse asyncResponse,
+      @PathParam("clusterId") String clusterId,
+      @PathParam("topicName") String topicName,
+      MappingIterator<ProduceRequest> requests) throws Exception {
```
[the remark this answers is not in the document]
rigelbm: Yes. Every time we can `next()`, it will try to unblock as quickly as possible, or so is my understanding.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
rigelbm: I have not written a test for StreamingResponse because it ultimately terminates into a ChunkedOutput sent to AsyncResponse. Although it's possible to mock the AsyncResponse, we can't get anything out of the ChunkedOutput (it's not a very good interface). I will try to improve the testability of this class later, but I prefer to delay it to a later PR.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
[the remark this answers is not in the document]
dimitarndimitrov: My guess would be that it will affect other APIs. Hopefully it won't be too bad.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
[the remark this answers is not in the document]
dimitarndimitrov: I reserve my right to noodle on this once we get some breathing room :D

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
dimitarndimitrov: Agreed.

dimitarndimitrov (approved):
+1.
Great work, Rigel!
I have mostly skimmed through the new testing util commit, but I think I went fairly thoroughly through the other changes.
Only one high-level comment right now - we can address any other things at a later point:
- Let's create a ticket (I wasn't able to find an existing one) for the openapi updates for this change.

rigelbm:
> * Let's create a ticket (I wasn't able to find an existing one) for the openapi updates for this change.

https://confluentinc.atlassian.net/browse/KREST-728

Veryfirefly:
Excuse me. Why the value of header needs to be encoded as Base64?

rigelbm:
> Excuse me. Why the value of header needs to be encoded as Base64?

Kafka headers can transport arbitrary byte strings, not just UTF-8 encoded strings.

Veryfirefly:
> > Excuse me. Why the value of header needs to be encoded as Base64?
> 
> Kafka headers can transport arbitrary byte strings, not just UTF-8 encoded strings.

True, but the header in the interface provided by kafka-rest to send messages to kafka needs to be Base64 processed. Why?

And can we not rely on schema-registry when sending messages? Because our business needs to send native json data instead of protobuf or Avro, and we need headers to route messages, we cannot use the v2 version.

The document [https://docs.confluent.io/platform/current/kafka-rest/api.html#records](https://docs.confluent.io/platform/current/kafka-rest/api.html#records)

leonardotorresaltez:
Hi All, in the documentation this feature is in early access.  Do you know when can we use it in production ?

ehumber:
Hi @leonardotorresaltez , we don't have a date for production usage yet.  The documentation will be updated once this is supported.

PR #799, opened 2021-02-08
KREST-558 Backport the fix for KREST-524 (fixing a concurrent test failure) to 5.5.x and 6.0.x.

dimitarndimitrov:
As the changes causing the [KREST-524](https://confluentinc.atlassian.net/browse/KREST-524)/[KREST-558](https://confluentinc.atlassian.net/browse/KREST-558) failures originally landed in `5.5.x`, the corresponding fix should have been merged starting from there.

This change cherry-picks it in `5.5.x` - it also needs to be merged up from there to `6.0.x`.
- From `6.0.x` though it's probably best to just `-s ours` merge up to `6.1.x` and `master`.

Tested locally in IntelliJ IDEA, by running 100 runs of `DefaultKafkaRestContextTest` (so a total of 400 tests):
- **before** the change, there were consistently between 5 and 10 failures.
- **after** the change, there were consistently no failures.

PR #800, opened 2021-02-08
KREST-558 6.0.x

dimitarndimitrov:
See #799 - this is a manual merge of the `5.5.x` fix from there into a `6.0.x`-based branch.
Created explicitly (and not `pint merge`-d) because the simple merge hits conflicts.

Tested locally in IntelliJ IDEA, by running 100 runs of `DefaultKafkaRestContextTest` (so a total of 300 tests):
- **before** the change, there were consistently between 5 and 10 failures.
- **after** the change, there were consistently no failures.


rigelbm (approved):
LGTM

PR #804, opened 2021-02-16
Return Bad Request on JSON parsing errors.

rigelbm:


dimitarndimitrov (approved):
+1.
I don't remember nearly enough from the last time I touched JAX-RS or Jackson and still haven't refreshed to my satisfaction, but the change makes sense.
My only comment here is - this seems very testable, can you please add a unit test for it?

rigelbm:
> My only comment here is - this seems very testable, can you please add a unit test for it?

Done

PR #806, opened 2021-02-18
KREST-638 Some small deployment improvements.

dimitarndimitrov:
Fixes the `kafka-rest-run-class` script to correctly work with the `development` assembly target, moves the `standalone` assembly target into the main profile, and beefs up a bit the Deployment section of the README.

Tested by ensuring that after a `mvn clean package`:
- the `kafka-rest-start` script now can be used to deploy the local changes.
    - We still need to pass valid configuration. I didn't default to the `./config/kafka-rest.properties` ones to avoid side effects to the `package` use-cases.
- ` java -jar ./kafka-rest/target/kafka-rest-6.2.0.0-standalone.jar ./config/kafka-rest.properties` correctly starts a server with the local changes if `kafka-rest.properties` is valid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,21 @@ To install from source, follow the instructions in the Development section below
 Deployment
 ----------
 
-The REST proxy includes a built-in Jetty server. The wrapper scripts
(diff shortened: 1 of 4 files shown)


rigelbm (approved):
LGTM

PR #809, opened 2021-02-22
MMA-9932 add dry run topic create and all topic configs apis

thomaskwscott:
This PR adds:

1. The ability to run create topics requests as "validate_only" so that the topic will not be created but the user can check they have the correct configs/perms.
2. A new endpoint: "/v3/clusters/{clusterId}/topics-configs" that returns a list of all topic configs from all topics in the cluster.

This is covered with unit and integration tests.


rigelbm (commented):
Can you split this PR in two: 1. the dry_run, and 2. the topic configs, so that it's easier to review? Thanks!

thomaskwscott:
Closing to be replaced with separate PRs

PR #811, opened 2021-02-24
MMA-9932 added default configs resource 

thomaskwscott:
New resource at:

/v3/clusters/{clusterId}/topics/{topicName}/default-configs

that fetches default configs that would be applied should topicName be created.

Note: this applies to new topics only, requests where topicName is pre-existing fail with 400 error.

This is covered with unit and integration tests


thomaskwscott:
@srpanwar-confluent @dimitarndimitrov bump for review.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/DefaultTopicConfigManagerImpl.java
```diff
@@ -0,0 +1,153 @@
...
+        .thenApply(cluster -> checkEntityExists(cluster, "Cluster %s cannot be found.", clusterId))
+        .thenCompose(
+            cluster ->
+                KafkaFutures.toCompletableFuture(
+                    adminClient.createTopics(
+                        singletonList(new NewTopic(topicName, Optional.empty(), Optional.empty())),
+                        new CreateTopicsOptions().validateOnly(true)
+                    ).config(topicName)))
+        .handle((result, ex) -> {
+              // InvalidRequestException can happen in the case of multi-tenancy, where passing
```
srpanwar-confluent, line 76: `multi-tenancy` is not available in AK. Maybe we should remove its reference from the comment.

srpanwar-confluent:
PR looks good to me!. Few things 
1. Should we move this `ce-kafka-rest-extension`? There is nothing proprietary with the API per se, so adding it here should be fine.
2. Reason I am thinking of 1) is that there is an edge case in ccloud, where, if the cluster is close to hitting the partition limit this call will fail. This is because #defaultpartition is > # partition remaining. This is not fixed in the original implementation as well, so I don't we should try to fix it in this PR.  It may require a bit more work to pipe the remaining partition cloud. Something to keep a note of though.

srpanwar-confluent (approved):

thomaskwscott:
closed in favour of https://github.com/confluentinc/ce-kafka-rest/pull/149

PR #815, opened 2021-03-04
KREST-729 Set the endpoint blocking filter's priority to be higher than all the predefined filter priorities.

dimitarndimitrov:
Before this change, a higher-priority (lower numerical priority) authentication filter could leak the existence of a blocked endpoint by returning HTTP 401 for it instead of HTTP 404.

In addition this change makes the endpoint blocking filter mimic the way Jersey throws HTTP 404/HTTP 405 by default.
There are still differences by which one might recognize that something's being hidden - Jersey will return an `Allow` header with the available methods when it throws HTTP 405, and will throw it only if some methods are indeed allowed for that endpoint.
- So if you POST to a non-existing endpoint, you'll get HTTP 404, not HTTP 405.

I considered whether it will be better to just always return HTTP 404 here, but in the end decided against that.
- Even if we did that, one might still have been able to recognize that something's being hidden (if that something has a non-hidden method on the same endpoint) by noticing that they should have got HTTP 405, not HTTP 404.

Tested by wiring a dummy authentication filter with `Priority.AUTHENTICATION` throwing HTTP 401 for every resource, then ensuring that:
- For non-existing endpoints, we get HTTP 404.
- For blocked endpoints, we get HTTP 404/HTTP 405.
- For non-blocked endpoints, we get HTTP 401 (thanks to the dummy filter).


kafka-rest/src/main/java/io/confluent/kafkarest/extension/ResourceBlocklistFeature.java
```diff
@@ -88,14 +92,15 @@ public void configure(ResourceInfo resourceInfo, FeatureContext context) {
     String value();
   }
 
+  @Priority(Priorities.AUTHENTICATION - 1)
```
rigelbm, line 95: Can you create a class `io.confluent.kafkarest.extension.MorePriorities` (to avoid clashing with `javax.ws.rs.Priorities`), and define a constant PRE_AUTHENTICATION (possibly equal to AUTHENTICATION - 500)?

rigelbm (commented):
> Tested by wiring a dummy authentication filter with Priority.AUTHENTICATION throwing HTTP 401 for every resource, then ensuring that:
>
> - For non-existing endpoints, we get HTTP 404.
> - For blocked endpoints, we get HTTP 404/HTTP 405.
> - For non-blocked endpoints, we get HTTP 401 (thanks to the dummy filter).

Can this be transformed into a unit test?

dimitarndimitrov:
> Can this be transformed into a unit test?

Done.

> Can you create a class `io.confluent.kafkarest.extension.MorePriorities` (to avoid clashing with `javax.ws.rs.Priorities`), and define a constant PRE_AUTHENTICATION (possibly equal to AUTHENTICATION - 500)?

Done.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ResourceDiscoverabilityTest.java
```diff
@@ -0,0 +1,82 @@
...
+import javax.ws.rs.Priorities;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.core.Configurable;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+import org.junit.Test;
+
+public class ResourceDiscoverabilityTest extends ClusterTestHarness {
```
rigelbm, line 34: ResourceBlocklistFeatureTest would make this test more discoverable (pun intended).

rigelbm (approved):
LGTM

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ResourceDiscoverabilityTest.java
dimitarndimitrov: Do you mean merging the two tests?

I was thinking about making this test a part of `ResourceBlocklistFeatureTest`, by tweaking the fake auth filter to 401 a specific endpoint only (that's not used in the other tests).
In the end I decided against that, as it makes the single resulting test harder to understand as the fake auth filter applies to just one of the test methods.
Let me know if you still prefer having me merge the two tests - if you don't, I'll keep these as they are.

rigelbm: Ack.

PR #817, opened 2021-03-09
Fix mtls test env run script.

rigelbm:
Missing packaging commands.

### DIFF
--- a/testing/environments/mtls/run.sh
+++ b/testing/environments/mtls/run.sh
@@ -42,6 +42,9 @@ docker pull 368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc/cp-zookeep
 # Download latest cp-server image.
 docker pull 368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc/cp-server:master-latest
 
+# Make sure kafka-rest is packaged.
+mvn -f "$base_dir"/kafka-rest/pom.xml -Dmaven.test.skip=true clean package
+
 # Generate certificates
 "$env_dir"/../../secrets/configure.sh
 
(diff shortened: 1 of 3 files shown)


testing/environments/mtls/run.sh
```diff
@@ -42,6 +42,9 @@ docker pull 368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc/cp-zookeep
 # Download latest cp-server image.
 docker pull 368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc/cp-server:master-latest
 
+# Make sure kafka-rest is packaged.
+mvn -f "$base_dir"/kafka-rest/pom.xml -Dmaven.test.skip=true clean package
```
dimitarndimitrov, line 46: nit: Do we want to skip compiling tests?
`-DskipTests` will skip running the tests while still compiling them, and will do that only negligibly slower than `-Dmaven.test.skip=true`.

dimitarndimitrov (approved):
+1.

P.S. Are the ports for the Kafka brokers correct (see below). The HTTP ones seem to be wrong.
```
  * 3 Kafka brokers:
    1. `kafka-1:9291`, `http://kafka-1:9193`
    2. `kafka-2:9292`, `http://kafka-2:9293`
    3. `kafka-3:9293`, `http://kafka-3:9393`
```

mtodzo (approved):
thanks again!

testing/environments/mtls/run.sh
rigelbm: Done.

dimitarndimitrov:
@rigelbm FWIW, the `mtls` and `sasl_plain` environments still use `-Dmaven.test.skip=true` (not that I mind too much).

PR #822, opened 2021-03-16
KREST-770 Annotate the v3 Produce API

dimitarndimitrov:
With:
- `@ResourceName` (so that it can be blocked if needed)
- `@PerformanceMetric` (so that metrics for it are exported)

Tested locally, by making sure that the API can be blocked with `api.endpoints.blocklist`, and that metrics for it are exported as attributes (e.g. `v3.records.create.request-error-rate`) to the `kafka-rest:type=jersey-metrics` MBean.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
@@ -53,7 +53,7 @@ import javax.ws.rs.container.Suspended;
 import javax.ws.rs.core.MediaType;
 
 @Path("/v3/clusters/{clusterId}/topics/{topicName}/records")
-@ResourceName("api.v3.records.create.*")
+@ResourceName("api.v3.produce.*")
 public final class ProduceAction {
 
   private static final Collector<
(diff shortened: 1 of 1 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -80,6 +83,8 @@ public ProduceAction(
   @POST
   @Consumes(MediaType.APPLICATION_JSON)
   @Produces(MediaType.APPLICATION_JSON)
+  @PerformanceMetric("v3.records.create")
```
rigelbm, line 86: I would call this `v3.records.produce-to-topic` instead. We will have `produce-to-partition` later.

rigelbm (approved):
LGTM

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
dimitarndimitrov: Fixed.
I also changed the resource names, so that they are a bit more hierarchical and allowing differentiation from the future v3 Consume and v3 Produce-to-partition APIs.

dimitarndimitrov: And again, after shortening the identifiers by not trying to fit them in a RESTful paradigm, per your suggestion.

rigelbm (approved):
LGTM

PR #823, opened 2021-03-17
KREST-771 Implement an allowlist for API endpoints.

dimitarndimitrov:
When specified, an allowlist disables all endpoints but the ones enumerated in it. An allowlist and a blocklist can be specified together in which case the blocklist can further constrain the allowlist, including blocking specific methods under a resource class that is present in the allowlist.

The allowlist functionality is implemented in the same feature (now renamed from `ResourceBlocklistFeature` to `ResourceAccesslistFeature`), and tested with integration tests in the spirit of the former `ResourceBlocklistFeatureTest`.
- I moved away from testing ACL APIs because in order to successfully call these we need the unnecessarily convoluted step of configuring a Kafka broker authorizer.

Note that the change in the feature name will break all users of `@ResourceName` (I know only of `ce-kafka-rest`, for which I'll be providing a PR soon).


rigelbm (approved):
Pure genius! LGTM! Thanks!

PR #828, opened 2021-03-31
MMA-10143 added support for isolation levels to consumer offset resources

thomaskwscott:
This PR adds the ability to produce isolation levels to resources that fetch latest offsets. These are passed as options queryparms and if unspecified will default to the previously hardcoded read_committed. 

Tested with new integration tests.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ConsumerLagsResource.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ConsumerLagsResource.java
@@ -30,16 +30,20 @@ import io.confluent.kafkarest.resources.AsyncResponses;
 import io.confluent.kafkarest.response.CrnFactory;
 import io.confluent.kafkarest.response.UrlFactory;
 import io.confluent.rest.annotations.PerformanceMetric;
+import org.apache.kafka.common.IsolationLevel;
+
 import java.util.Comparator;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Provider;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.container.AsyncResponse;
 import javax.ws.rs.container.Suspended;
 import javax.ws.rs.core.MediaType;
@@ -47,6 +51,8 @@ import javax.ws.rs.core.MediaType;
 @Path("/v3/clusters/{clusterId}/consumer-groups/{consumerGroupId}/lags")
 @ResourceName("api.v3.consumer-lags.*")
 public final class ConsumerLagsResource {
+
+  public static final String DEFAULT_ISOLATION_LEVEL = "read_committed";
 
   private final Provider<ConsumerLagManager> consumerLagManager;
   private final CrnFactory crnFactory;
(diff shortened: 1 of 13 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ConsumerLagsResource.java
```diff
@@ -48,6 +52,8 @@
 @ResourceName("api.v3.consumer-lags.*")
 public final class ConsumerLagsResource {
 
+  public static final String DEFAULT_ISOLATION_LEVEL = "read_committed";
```
srpanwar-confluent, line 55: IsolationLevel.READ_COMMITTED?

srpanwar-confluent: Ok, I see we are setting this as defaultvalue

rigelbm:
@thomaskwscott @srpanwar-confluent Does this API makes sense, though? Is not the previously hard-coded READ_COMMITTED is good enough? If not, would not be better to just change it to whatever is best, instead of exposing this option? I think this option is too low level to expose through the REST API. cc @ahuang98, what do you think?

ahuang98:
@thomaskwscott do you have more background/context for these changes? I believe kafka apis (blueway) basically hard coded usage of IsolationLevel.READ_COMMITTED

thomaskwscott:
> @thomaskwscott do you have more background/context for these changes? I believe kafka apis (blueway) basically hard coded usage of IsolationLevel.READ_COMMITTED

@ahuang98 I don't know if it's hard coded in the UI but on the back end the isolation level can be passed as a query param:

https://github.com/confluentinc/blueway/blob/009ce580a5d9dcf3495a33af4fad60e3bd2832db/control-center/src/main/java/io/confluent/controlcenter/rest/ConsumerOffsetsResource.java#L55

This PR brings parity in Rest Proxy so that we can retire that API.

rigelbm:
> This PR brings parity in Rest Proxy so that we can retire that API.

It does not matter that the C3 backend accepts this as a query parameter. This is too low level to expose in the API, in my opinion. 

It does not make sense to port over this operation unless we are sure it is being used. Otherwise, we are just porting over tech debt from C3 backend.  We need to know how the C3 frontend uses this API, and if it makes usage of this query param at all. If it doesn't, I vote against adding it to Kafka REST.

dimitarndimitrov:
> It does not make sense to port over this operation unless we are sure it is being used. Otherwise, we are just porting over tech debt from C3 backend. We need to know how the C3 frontend uses this API, and if it makes usage of this query param at all. If it doesn't, I vote against adding it to Kafka REST.

I agree with that, and I would even expand on the "porting tech debt" part.

Porting code the way we've been porting code here is an inherently volatile operation and not a "lift and shift" for code, so it disrupts any Lindy effect that the old code might have been benefiting from.
What this means is that it's OK to rethink what's the best way to achieve the required final functionality _in C3_, as if the old code doesn't really exist.

Aside from the obvious case, where the best way forward is to just port the backend functionality to Kafka REST, or the less obvious case here, where we might be able to avoid adding a redundant API, I would like us to also consider a third option, which might otherwise just fall into the first, obvious case.

That third option would pertain to APIs that ideally should be implemented differently, but which then would necessitate changes on the C3 frontend.
If it's not possible to do the necessary work on the C3 frontend in the appropriate timeframe, I think it's best to add such API not to this repo, but to our internal add-on. I also think we should immediately mark such APIs as `@Deprecated`, in order to annotate that we're actually porting tech debt, and that we need to prioritize the corresponding clean-up in the near future.

thomaskwscott:
@dimitarndimitrov @rigelbm these are all good points and I'm trying to confirm the UI side now. If we can be certain there is no case for a different isolation level upstream then we can close this PR.

Isolation Level is a strange case though and I'm surprised this hasn't actually caused more support cases. The default level for consumers and adminclients is read_uncommitted so with base configurations REST Proxy can produce different values for list offset requests to the other clients with no explanation. I figure REST Proxy should broadly follow the behaviour of the other clients so maybe we consider changing the default here or in the other clients?

srpanwar-confluent:
Agree with the points expressed above, but I don't fully agree with exposing Isolation level or similar functionality in a different way unless are adding a new vocabulary to describe such operations. Said so, I'd leave it folks here to decide the best way forward. 
Wrt to usage in FE, looks like we are not passing isolation level to backend call, so not adding it here is fine.

Wrt to `porting tech debt`, it is a balance we want to strive for at the end of the day. For APIs that exist in C3 and naturally fit in the rest API model we would like to add them to the public repo. Said so, not all of them fit the model and I agree we can add them as extension APIs and that is what we have been doing in other PRs. Another point I'd like to emphasize is changing the frontend is not trivial and carries its own see of risk of regression.

thomaskwscott:
Closing this one out as it is not necessary to satisfy the front end requirements.

PR #830, opened 2021-04-05
KREST-826, KREST-837 Reduce coupling to kafka-clients

dimitarndimitrov:
Reduce coupling to `kafka-clients` by providing our own `waitForCondition` test utilities based on `testWithRetry`.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
@@ -31,6 +31,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.LockSupport;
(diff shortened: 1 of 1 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
```diff
@@ -246,28 +238,127 @@ public static void assertPartitionOffsetsEqual(List<PartitionOffset> a, List<Par
...
+      boolean isExponentialBackoff) {
+    long sleepNanos = initialRetryInterval.toNanos();
+    long start = System.nanoTime();
+    long elapsed;
+    long maxElapsed = timeout != null ? timeout.toNanos() : Long.MAX_VALUE;
     for (int i = 0;; i++) {
       try {
         assertions.run();
         return;
+      } catch (NoRetryException e) {
```
rigelbm, line 265: Do we use this `NoRetryException` anywhere? If we don't, I would get rid of it. One less dependency.

rigelbm (approved):
LGTM

kafka-rest/src/test/java/io/confluent/kafkarest/TestUtils.java
dimitarndimitrov: We do not, and it's not part of the [Kafka 2.7.0 API](https://kafka.apache.org/27/javadoc/index.html?overview-summary.html), but I was afraid that some components we query might throw it.

Still, what you're suggesting is exactly what I wanted to do, so consider it done.

dimitarndimitrov:
There is one test failure - `ConsumerLagsResourceIntegrationTest.getConsumerLag_returnsConsumerLag`. It seems to be entirely unrelated to the changes here though, as it doesn't use `TestUtils`.

PR #831, opened 2021-04-08
KREST-865 Remove usage of the deprecated SimpleAclAuthorizer.

dimitarndimitrov:
Starting from `5.4.x`, this will fix the deprecated API usage in `master` that is affected by https://github.com/confluentinc/kafka/pull/544.
- There are other deprecated API usages across the different branches, which should be addressed as soon as possible.

Tested locally with `mvn -f kafka-rest/pom.xml -Dit.test=AuthorizationErrorTest test-compile failsafe:integration-test`.


rigelbm (approved):
LGTM

PR #833, opened 2021-04-09
修复因编译器不一致导致，识别不了Java新特性的代码

GZ315200:


rigelbm:
@GZ315200, what compiler exactly is having problems with this code?

PR #834, opened 2021-04-09
KREST-865 Remove usage of a deprecated Admin Client ConfigEntry constructor

dimitarndimitrov:
That constructor was actually used a lot throughout the configuration management tests.
Because there's no good alternative for that constructor, a subtype mimicking the old constructor was added.

As the deprecated constructor is actually being removed in the most recent AK version, I'll go ahead and merge this once CI looks good. The PR will be available for post-merge reviews and for posterity.

Tested locally, by ensuring that the affected tests still pass, and that IntelliJ doesn't find any other usages of the deprecated constructor.

P.S. Moving away from this constructor also means it's better to change our `AbstractConfig` concept, so that it relies on the config source to determine whether the related config is default or not. That change however affects many classes, so it should come in a separate PR.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConfigSource.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConfigSource.java
@@ -74,4 +74,12 @@ public enum ConfigSource {
       return UNKNOWN;
     }
   }
+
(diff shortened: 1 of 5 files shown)


dimitarndimitrov:
The CI run failed because of a downstream dependency failure unrelated to these changes.
Grepping the enormous (~100 MB) log shows that there were no test failures in `kafka-rest` itself, so I will go ahead and merge this.

PR #836, opened 2021-04-12
fix: resolve ProduceController early/outside of lambda

srpanwar-confluent:
There is a subtle bug in `produce` logic where KafkaRestCleanupFilter is called before the lambda function (passed to StreamingResponse) is executed. This has got to do with how jersey handles(or lack of thereof) `ChunkedOutput` during a request's lifecycle. Basically, once `AsyncResponse` is resumed the subsequent Filters are executed. This leads to `KafkaRestContextProvider` getting reset before the `produceController` can be resolved. As a result, the `KafkaRestContext` is not the one that is set by the client but instead is `DefaultKafkaRestContext`.

The fix in this PR is to resolve `produceController` outside of the lambda function.  A better fix, however, would be to 
1. Call `KafkaRestContextProvider.clearCurrentContext()` inside one of the lifecycle event listeners. 
2. If 1. is not possible then create a custom ListenableChunkedOuput class, register a close listener in `KafkaRestCleanupFilter`, and clear context once chunked response has finished.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
@@ -90,13 +90,14 @@ public final class ProduceAction {
       @PathParam("clusterId") String clusterId,
       @PathParam("topicName") String topicName,
       MappingIterator<ProduceRequest> requests) throws Exception {
+    ProduceController controller = produceController.get();
     StreamingResponse.from(requests)
-        .compose(request -> produce(clusterId, topicName, request))
+        .compose(request -> produce(clusterId, topicName, request, controller))
         .resume(asyncResponse);
   }
 
   private CompletableFuture<ProduceResponse> produce(
-      String clusterId, String topicName, ProduceRequest request) {
+      String clusterId, String topicName, ProduceRequest request, ProduceController controller) {
     Optional<RegisteredSchema> keySchema =
         request.getKey().flatMap(key -> getSchema(topicName, /* isKey= */ true, key));
     Optional<EmbeddedFormat> keyFormat =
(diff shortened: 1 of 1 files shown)


rigelbm (dismissed):
Hi @srpanwar-confluent. Thanks for catching this and sorry if you had to spend time investigating it. To add further context, this error happens not because of the way Jersey handles the ChunkedOutput, but rather because we store our KafkaRestContext in a inheritable thread-local. It just so happens that the thread that executes the lambda (and thus finishes the request) is not a descendant of (or same as) the thread that starts the request. We have this exact same problem in all other resources, and your solution is exactly what we do elsewhere, i.e. inject ProduceController in the starting thread. We currently don't have a better solution for now. When we switch over to use Dependency Inject everywhere in the server, and thus get rid of the annoying KafkaRestContext, we can explore better ways of solving it (e.g. don't use inheritable thread-local). Thanks for the fix!!!

rigelbm (commented):
I removed my Approval since I don't understand your last commit. What are trying to do, and why what we had is not enough?

srpanwar-confluent:
> why what we had is not enough?
@rigelbm, regarding the above, the 1st commit is sufficient. Happy to revert back the 2nd commit. The issue I see with 1st commit is that it forces developers to follow a particular pattern and if not careful can happen again in the future for any new API we add. The 2nd commit makes it a bit more robust by moving the fix at the common catch-all/filter layer while letting lazy resolution.
Said so, 2nd commit is not fool-proof either if one ends up using ChunkedOuput over ListenableChunkedOutput.

For your earlier comment, looks like we are already thinking of a better solution to this. In that case, I agree 2nd commit is adding one more pattern that will become redundant.

srpanwar-confluent:
Reverted 2nd

rigelbm (approved):
Thanks @srpanwar-confluent!! LGTM!

PR #838, opened 2021-04-13
KREST-600: Fix KafkaConsumerManagerTest.testBackoffMsControlsPollCalls flaky test.

rigelbm:
Current way it is being tested, it relies on Thread.sleep to try to peek into the KafkaConsumerManager at right time. Problem is, 1. it is using the wrong time (it assumes the backoff is 250ms, when actually it is 50ms), and 2. it is peeking the wrong thing. I just looks if there's a delayed task, but that gives no guarantees the task is being retried at the given backoff. So I rewrote it to actually the backoff.

It was previously consistently failing 5% of time (out of 1000 runs). Now it fails 0%. :)

Example failure: https://jenkins.confluent.io/job/confluentinc/job/kafka-rest/job/6.0.x/480/testReport/junit/io.confluent.kafkarest.v2/KafkaConsumerManagerTest/testBackoffMsControlsPollCalls/

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
@@ -427,7 +427,8 @@ public class KafkaConsumerManagerTest {
       long actualNanos = pollTimestamps.get(i) - pollTimestamps.get(i-1);
       assertTrue(
           String.format(
-              "Expected time between poll calls to be at least 50ms, but was %sms.",
+              "Expected time between poll calls to be at least %dms, but was %sms.",
+              backoffMillis,
               TimeUnit.NANOSECONDS.toMillis(actualNanos)),
           actualNanos >= TimeUnit.MILLISECONDS.toNanos(backoffMillis));
     }
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
@@ -32,8 +32,6 @@ import io.confluent.kafkarest.entities.EmbeddedFormat;
 import io.confluent.kafkarest.entities.TopicPartitionOffset;
 import io.confluent.kafkarest.entities.v2.ConsumerOffsetCommitRequest;
 import io.confluent.kafkarest.entities.v2.ConsumerSubscriptionRecord;
-import java.time.Clock;
-import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -91,7 +89,7 @@ public class KafkaConsumerManagerTest {
 
     @Before
     public void setUp() {
-      setUpConsumer(setUpProperties());
+        setUpConsumer(setUpProperties());
     }
 
     private void setUpConsumer(Properties properties) {
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
@@ -394,12 +394,12 @@ public class KafkaConsumerManagerTest {
     consumerManager = new KafkaConsumerManager(config, consumerFactory);
     consumer = new MockConsumer<>(OffsetResetStrategy.EARLIEST, groupName);
     bootstrapConsumer(consumer);
-    ArrayList<Long> pollTimestamps = new ArrayList<>();
+    ArrayList<Long> pollTimestampsMillis = new ArrayList<>();
     consumer.schedulePollTask(
         new Runnable() {
           @Override
           public void run() {
-            pollTimestamps.add(System.nanoTime());
+            pollTimestampsMillis.add(TimeUnit.NANOSECONDS.toMillis(System.nanoTime()));
             consumer.schedulePollTask(this);
           }
         });
@@ -419,18 +419,20 @@ public class KafkaConsumerManagerTest {
     //   1. second initial poll() returns empty
     //   2..N-1. wait backoffMillis, poll() returns empty, repeat
     //   N. wait max(backoffMillis, remainder of timeoutMillis), poll() returns empty
-    assertEquals(2 + (timeoutMillis / backoffMillis), pollTimestamps.size());
+    assertEquals(2 + (timeoutMillis / backoffMillis), pollTimestampsMillis.size());
 
     // We need to verify that the interval between poll() calls in 1..N-1 above is at least
     // backoffMillis.
-    for (int i = 2; i < pollTimestamps.size() - 1 ; i++) {
-      long actualNanos = pollTimestamps.get(i) - pollTimestamps.get(i-1);
+    for (int i = 2; i < pollTimestampsMillis.size() - 1 ; i++) {
+      long actualMillis = pollTimestampsMillis.get(i) - pollTimestampsMillis.get(i-1);
       assertTrue(
           String.format(
-              "Expected time between poll calls to be at least %dms, but was %sms.",
+              "Expected time between poll calls %d and %d to be at least %dms, but was %sms.",
+              i,
+              i-1,
               backoffMillis,
-              TimeUnit.NANOSECONDS.toMillis(actualNanos)),
-          actualNanos >= TimeUnit.MILLISECONDS.toNanos(backoffMillis));
+              actualMillis),
+          actualMillis >= backoffMillis);
     }
   }
 
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
@@ -33,7 +33,6 @@ import io.confluent.kafkarest.entities.TopicPartitionOffset;
 import io.confluent.kafkarest.entities.v2.ConsumerOffsetCommitRequest;
 import io.confluent.kafkarest.entities.v2.ConsumerSubscriptionRecord;
 import java.time.Instant;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
@@ -32,7 +32,9 @@ import io.confluent.kafkarest.entities.EmbeddedFormat;
 import io.confluent.kafkarest.entities.TopicPartitionOffset;
 import io.confluent.kafkarest.entities.v2.ConsumerOffsetCommitRequest;
 import io.confluent.kafkarest.entities.v2.ConsumerSubscriptionRecord;
+import java.time.Duration;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -381,27 +385,47 @@ public void onCompletion(List<TopicPartitionOffset> offsets, Exception e) {
...
-        assertEquals(1, consumerManager.delayedReadTasks.size());
+  @Test
+  public void testBackoffMsControlsPollCalls() throws Exception {
+    bootstrapConsumer(consumer);
+    ArrayList<Instant> pollTimestamps = new ArrayList<>();
+    consumer.schedulePollTask(
+        new Runnable() {
+          @Override
+          public void run() {
+            pollTimestamps.add(Instant.now());
```
dimitarndimitrov, line 396: By default `Instant.now()` uses the system clock which is non-monotonic, and below we expect it to be monotonic when we compare differences between consecutive poll timestamps. I'd use `System.nanoTime()` instead.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -381,27 +385,47 @@ public void onCompletion(List<TopicPartitionOffset> offsets, Exception e) {
...
+        BinaryKafkaConsumerState.class,
+        -1,
+        Long.MAX_VALUE,
+        (records, e) -> sawCallback = true);
+
+    TestUtils.waitForCondition(() -> sawCallback, "Consumer task finished.");
+    assertEquals(
+        "Expected (2 initial polls + (1s timeout / 50ms backoff) =) 22 consumer polls.",
+        22,
+        pollTimestamps.size());
```
dimitarndimitrov, line 412: Honest question: Do you think it makes sense to bump up the backoff? I have no idea whether it's OK to always expect such a precise time-slicing from the consumer.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -520,38 +544,48 @@ public void onCompletion(List<ConsumerRecord<ByteString, ByteString>> records, E
...
-                consumer.groupName, ConsumerInstanceConfig.create(EmbeddedFormat.BINARY));
-
-        consumer.cid(cid);
-        consumerManager.subscribe(consumer.groupName, cid, new ConsumerSubscriptionRecord(Collections.singletonList(topicName), null));
-        consumer.rebalance(Collections.singletonList(new TopicPartition(topicName, 0)));
-        consumer.updateBeginningOffsets(singletonMap(new TopicPartition(topicName, 0), 0L));
-
-        return referenceRecords;
+  /**
+   * Subscribes a consumer to a topic and schedules a poll task
```
dimitarndimitrov, line 548: nit: This no longer calls `schedulePollTask`, can you either update the Javadoc, or add an inline comment when adding the records how is this equivalent?

dimitarndimitrov (commented):
That test doesn't seem to be in great shape, I'm a bit surprised that only one test in it flakes out. Some suggestions below.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -75,7 +78,7 @@
     private static final String topicName = "testtopic";
 
     // Setup holding vars for results from callback
-    private boolean sawCallback = false;
+    private boolean sawCallback;
     private static Exception actualException = null;
     private static List<ConsumerRecord<ByteString, ByteString>> actualRecords = null;
     private static List<TopicPartitionOffset> actualOffsets = null;
```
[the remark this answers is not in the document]
dimitarndimitrov: > We can also have another field using a resettable latch (i.e. a `Phaser` or a 2-party `CyclicBarrier` if the `KafkaConsumerThreadPoolExecutor` thread can be blocked for a bit) that tests await in order to avoid arbitrary sleeps.

It actually might be simpler to just have the latch mimic the lifecycle of the state holder - add a `volatile CountDownLatch` field, recreate the latch in `setUp()`, countdown the latch in the `ConsumerReadCallback` callback.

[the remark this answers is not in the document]
rigelbm: I actually undid the changes in `sawCallback` and `setUp`, and used a CountDownLatch in the test. Changing anything here would force me to changes pretty much all the tests, which I would rather do separately, preferably in `master`.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -381,27 +385,47 @@ public void onCompletion(List<TopicPartitionOffset> offsets, Exception e) {
...
+        });
+    consumerManager.readRecords(
+        groupName,
+        consumer.cid(),
+        BinaryKafkaConsumerState.class,
+        -1,
+        Long.MAX_VALUE,
+        (records, e) -> sawCallback = true);
+
+    TestUtils.waitForCondition(() -> sawCallback, "Consumer task finished.");
```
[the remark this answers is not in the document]
rigelbm: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
rigelbm: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
rigelbm: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
rigelbm: Removed comment, as it is redundant.

rigelbm:
Thanks for the review @dimitarndimitrov. PTAL

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -381,27 +383,56 @@ public void onCompletion(List<TopicPartitionOffset> offsets, Exception e) {
...
-        // backoff is 250
-        Thread.sleep(100);
-        // backoff should be in place right now. the read task should be delayed and re-ran until the max.bytes or timeout is hit
-        assertEquals(1, consumerManager.delayedReadTasks.size());
-        Thread.sleep(100);
-        assertEquals(1, consumerManager.delayedReadTasks.size());
+  @Test
+  public void testBackoffMsControlsPollCalls() throws Exception {
+    long timeoutMillis = 1000;
+    long backoffMillis = 100;
```
dimitarndimitrov, line 389: nit: Use `long` literals (add `L` suffix).

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -381,27 +383,56 @@ public void onCompletion(List<TopicPartitionOffset> offsets, Exception e) {
...
+    Properties props = setUpProperties();
+    props.put(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG, String.valueOf(timeoutMillis));
+    props.put(KafkaRestConfig.CONSUMER_ITERATOR_BACKOFF_MS_CONFIG, String.valueOf(backoffMillis));
+    config = new KafkaRestConfig(props, new SystemTime());
+    consumerManager = new KafkaConsumerManager(config, consumerFactory);
+    consumer = new MockConsumer<>(OffsetResetStrategy.EARLIEST, groupName);
```
dimitarndimitrov, line 395: nit: I think you can replace all of this with `setUpConsumer(props)`.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -381,27 +383,56 @@ public void onCompletion(List<TopicPartitionOffset> offsets, Exception e) {
...
+    bootstrapConsumer(consumer);
+    ArrayList<Long> pollTimestamps = new ArrayList<>();
```
dimitarndimitrov, line 397: nit: I'm not sure on which thread is this mutated, but just to be sure, let's use `CopyOnWriteArrayList`.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -520,38 +551,45 @@ public void onCompletion(List<ConsumerRecord<ByteString, ByteString>> records, E
...
+    consumer.rebalance(Collections.singletonList(new TopicPartition(topicName, 0)));
+    consumer.updateBeginningOffsets(singletonMap(new TopicPartition(topicName, 0), 0L));
+    for (ConsumerRecord<ByteString, ByteString> record : referenceRecords) {
+      consumer.addRecord(
+          new org.apache.kafka.clients.consumer.ConsumerRecord<>(
+              record.getTopic(),
+              record.getPartition(),
+              record.getOffset(),
+              record.getKey().toByteArray(),
+              record.getValue().toByteArray()));
```
dimitarndimitrov, line 587: Nice!

dimitarndimitrov (approved):
+1.
Nice work, some nits below.

P.S. I guess if we see anything here failing again, we can go for the bigger changes starting from `6.2.x` (so that we don't have to deal with merge conflicts).

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
rigelbm: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
rigelbm: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
rigelbm: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
rigelbm: Ack.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -33,10 +33,13 @@
 import io.confluent.kafkarest.entities.v2.ConsumerOffsetCommitRequest;
 import io.confluent.kafkarest.entities.v2.ConsumerSubscriptionRecord;
 import java.time.Instant;
+import java.util.ArrayList;
```
dimitarndimitrov, line 36: nit: This import seems redundant now.

rigelbm: Done.

PR #839, opened 2021-04-13
MMA-10463: list all broker configs

srpanwar-confluent:
Adds a new endpoint - /v3/clusters/{clusterId}/brokers/-/configs

**What**
Fetches configuration for all brokers in the cluster.


thomaskwscott (approved):
LGTM

srpanwar-confluent:
This change will also bring the broker API on par with topic API, as `list all configs` is present only for topics today.

PR #841, opened 2021-04-16
Cherry-pick of "fix: resolve ProduceController early/outside of lambda (#836)"

rigelbm:


dimitarndimitrov (approved):
+1.
It would be nice to create a JIRA about removing the thread-local approach for context propagation.

PR #843, opened 2021-04-26
MMA-10482: include authorized operations in topic response

srpanwar-confluent:
Fetch `authorizedOperations` for topics

**Why**
Currently, C3 depends on authorized operations to determine the actions a user can take. For example - can a given topic be deleted, and if yes, then show the appropriate elements in the UI.
The existing Kafka-API  (/2.0/kafka/<cluster>/topics) includes the authorized operations in the response, making this possible. 

**Change**
As the title says, the proposal here is to add an equivalent API in Kafka-rest to expose authorized operations for the given resources. 

Since operations are not resources but metadata on the topic, the change here Includes it in the response of the topic (/kafka/v3/clusters/<cluster>/topics) itself. 
```
"authorizedOperations": [
      "WRITE",
      "ALTER_CONFIGS",
      "DESCRIBE_CONFIGS",
      "DELETE",
      "DESCRIBE",
      "READ",
      "CREATE",
      "ALTER"
    ]
```
```API: /kafka/v3/clusters/<cluster>/topics?include_authorized_operations=true```

### CODE REVIEW

File: kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java [hunk 1]
```diff
@@ -83,10 +85,14 @@ public TopicsResource(
   @PerformanceMetric("v3.topics.list")
   @ResourceName("api.v3.topics.list")
   public void listTopics(
-      @Suspended AsyncResponse asyncResponse, @PathParam("clusterId") String clusterId) {
+      @Suspended AsyncResponse asyncResponse,
+      @PathParam("clusterId") String clusterId,
+      @QueryParam("include_authorized_operations") @DefaultValue("false")
```
rigelbm, line 90: I don't think we need the option. Just always include it.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/Topic.java
```diff
@@ -35,13 +38,32 @@
 
   public abstract boolean isInternal();
 
+  @Nullable
+  public abstract Set<AclOperation> authorizedOperations();
```
rigelbm, line 42: We have our own `AclOperation` type in `kafkarest.entities`. Use that instead.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/TopicData.java
```diff
@@ -47,6 +50,11 @@
   @JsonProperty("partition_reassignments")
   public abstract Relationship getPartitionReassignments();
 
+  @Nullable
+  @JsonProperty("authorized_operations")
+  public abstract Set<AclOperation> getAuthorizedOperations();
```
rigelbm, line 55: Ditto.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/TopicData.java
```diff
@@ -69,7 +78,8 @@ static TopicData fromJson(
       @JsonProperty("replication_factor") int replicationFactor,
       @JsonProperty("partitions") Relationship partitions,
       @JsonProperty("configs") Relationship configs,
-      @JsonProperty("partition_reassignments") Relationship partitionReassignments
+      @JsonProperty("partition_reassignments") Relationship partitionReassignments,
+      @JsonProperty("authorized_operations") Set<AclOperation> authorizedOperations
```
rigelbm, line 82: Move up. Relationships should be in the end. Same for everywhere else.

rigelbm (commented):
Just a few comments.

api/v3/openapi.yaml
```diff
@@ -1561,6 +1561,7 @@ components:
             - is_internal
             - replication_factor
             - partitions_count
+            - authorized_operations
```
rigelbm, line 1564: It's not required, right? If you don't set `include_authorized_operations=true`, you won't get this property.

api/v3/openapi.yaml
```diff
@@ -1574,7 +1575,11 @@ components:
             replication_factor:
               type: integer
             partitions_count:
-                type: integer
+              type: integer
```
rigelbm, line 1578: You should also add the query parameter in the spec of the respective API.

checkstyle/suppressions.xml
```diff
@@ -43,7 +43,7 @@
 
     <suppress
             checks="(ParameterNumber)"
-            files="(ClusterData).java"
+            files="(ClusterData|TopicData).java"
```
rigelbm, line 46: I don't add any suppressions here in this file. Instead we add them (if necessary) in the place we want to supress, using comments:

```
// CHECKSTYLE:OFF:ParameterNumber
// My code that needs supression.
// CHECKSTYLE:ON:ParameterNumber
```

kafka-rest/src/main/java/io/confluent/kafkarest/entities/Topic.java
```diff
@@ -35,13 +38,32 @@
 
   public abstract boolean isInternal();
 
+  @Nullable
```
rigelbm, line 41: This should not be nullable.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/Topic.java
```diff
@@ -35,13 +38,32 @@
...
+        clusterId, name, ImmutableList.copyOf(partitions), replicationFactor, isInternal, null);
+  }
+
+  public static Topic create(
+      String clusterId,
+      String name,
+      List<Partition> partitions,
+      short replicationFactor,
+      boolean isInternal,
+      Set<Operation> authorizedOperations) {
```
rigelbm, line 60: Mark `authorizedOperations` as `@Nullable`.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/Topic.java
```diff
@@ -35,13 +38,32 @@
...
+    return new AutoValue_Topic(
+        clusterId,
+        name,
+        ImmutableList.copyOf(partitions),
+        replicationFactor,
+        isInternal,
+        authorizedOperations);
```
rigelbm, line 67: `authorizedOperations` == null ? emptySet() : `authorizedOperations`

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/TopicData.java
```diff
@@ -41,6 +44,10 @@
   @JsonProperty("partitions_count")
   public abstract int getPartitionsCount();
 
+  @Nullable
```
rigelbm, line 47: Ditto.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManagerImpl.java
```diff
@@ -130,8 +132,15 @@
...
-  private CompletableFuture<List<Topic>> describeTopics(String clusterId, List<String> topicNames) {
-    return KafkaFutures.toCompletableFuture(adminClient.describeTopics(topicNames).all())
+  private CompletableFuture<List<Topic>> describeTopics(
+      String clusterId, List<String> topicNames) {
+    return KafkaFutures.toCompletableFuture(
+            adminClient
+                .describeTopics(
+                    topicNames,
+                    new DescribeTopicsOptions()
+                        .includeAuthorizedOperations(true))
```
rigelbm, line 142: We should only set it if user is asking for it. I assume it is optional because there's a cost involved in asking for it :). That means you should pipe include_authorized_operations from the resource to the manager.

rigelbm (commented):
Where is the query parameter? I don't see it anywhere (spec or code). Did we give up on it and decided to always include it? If so, we should probably update the PR description. I have not strong feelings either way.

PR #844, opened 2021-04-27
MMA-10490: include count of partitions in topic(s) response

srpanwar-confluent:
**What**
Get useful meta info without making an additional call to fetch the full partition data. 

**Why**
For the web clients that show topic data in a single view (e.g. table), making an additional call to get and display partition count leads to an inefficient fetch pattern.

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -1560,6 +1560,7 @@ components:
             - topic_name
             - is_internal
             - replication_factor
+            - partitions_count
             - partitions
             - configs
             - partition_reassignments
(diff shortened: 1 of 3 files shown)


dimitarndimitrov (changes requested):
The change makes sense to me and I don't think it violates Kafka REST's entity relationship principles.
You need to also update the OpenAPI YAML spec (both the `TopicData` component and the examples where it has been used) and the `@JsonCreator` factory method with the new field though.

dimitarndimitrov (approved):
+1.

Can you please also update the examples for `ListTopicsResponse` in the OpenAPI spec?

dimitarndimitrov:
@srpanwar-confluent this change broke `TopicsResourceIntegrationTest`: https://jenkins.confluent.io/job/confluentinc/job/kafka-rest/job/master/4269/

Could you please fix that? My guess is that you'll need to update the expectations to include the new partitions count field.

PR #845, opened 2021-04-28
fix integration test

srpanwar-confluent:
**What**
Fix the integration test failure. The change here though includes more than just the fix for the test. `TopicManager.createTopic` has been changed to return `CreateTopicsResult` instead of void. Few advantages of doing this
1. Caller can decide how it wants to use `CreateTopicsResult` ([example](https://github.com/confluentinc/kafka-rest/pull/845/files#diff-0811689ae333a0789fa5a6c9911332eeb8fb48ebfac0f0401ff329f4d9b6375fR681)). On the flip side, it is the responsibility of the caller now to make sure create has succeeded. This is not a publicly exposed API so no impact on backward compatibility.
2. Allows us to return the correct partition count vs passing back arbitrary value.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/TopicsResourceIntegrationTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/TopicsResourceIntegrationTest.java
@@ -96,6 +96,7 @@ public class TopicsResourceIntegrationTest extends ClusterTestHarness {
                             .setTopicName(TOPIC_1)
                             .setInternal(false)
                             .setReplicationFactor(1)
+                            .setPartitionsCount(1)
                             .setPartitions(
                                 Resource.Relationship.create(
                                     baseUrl
@@ -129,6 +130,7 @@ public class TopicsResourceIntegrationTest extends ClusterTestHarness {
                             .setTopicName(TOPIC_2)
                             .setInternal(false)
                             .setReplicationFactor(1)
+                            .setPartitionsCount(1)
                             .setPartitions(
                                 Resource.Relationship.create(
                                     baseUrl
@@ -162,6 +164,7 @@ public class TopicsResourceIntegrationTest extends ClusterTestHarness {
                             .setTopicName(TOPIC_3)
                             .setInternal(false)
                             .setReplicationFactor(1)
+                            .setPartitionsCount(1)
                             .setPartitions(
                                 Resource.Relationship.create(
                                     baseUrl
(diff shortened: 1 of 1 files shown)


dimitarndimitrov (changes requested):
I would rather avoid the `CreateTopicsResult` change at that moment, because of the following:
- We have generally avoided so far returning the created entities for the C of our CRUD APIs. There's an argument that the current model should be good enough because:
    - You already have the specifications that you have explicitly passed for the topic creation and you can assert that they have been honored.
    - Any other topic config depends on the default config setup, which is its own thing that should be configured and checked separately.
- We have generally avoided so far using `kafka-clients` entities when possible - almost all of the manager interfaces are free of such dependencies, with very small exceptions. In almost all cases the serde allows us to avoid them.
- Having in mind that the current API allows the creation of just a single topic, I don't see much utility in the added flexibility of that asynchronous result container.
- This change is separate from the test fix, so we should address it separately, after a discussion, and if we agree to go with it, in its own commit.

So I think we should just fix the tests for now and open up a discussion about returning the created entity as a result in our creation APIs.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
```diff
@@ -53,8 +54,9 @@
   /**
    * Creates a new Kafka {@link Topic} with either partitions count and replication factor
    * or explicitly specified partition-to-replicas assignments.
+   * @return
    */
-  CompletableFuture<Void> createTopic(
+  CompletableFuture<CreateTopicsResult> createTopic(
```
rigelbm, line 59: I agree with @dimitarndimitrov that we should not expose a `kafka.clients` type here in `kafkarest.controllers`. In particular, `CompletableFuture<CreateTopicsResult>` does not make sense, since `CreateTopicsResult` itself has another future inside. We flat-map all such cases in our interfaces. That all being said, it seems you do need to propagate some information about the topic creation up. I suggestion you creating a new auto-value in `kafkarest.entities` for this, e.g. `CreateTopicResult`, that only contain the information you need.

rigelbm (commented):
Thanks for the fix @srpanwar-confluent!! Added some further comments to what @dimitarndimitrov said.

srpanwar-confluent:
@dimitarndimitrov @rigelbm makes sense. Scoping the change to just the test fix.

dimitarndimitrov (approved):
+1.

Thank you for addressing our comments super-quick!
The sole test failure is unrelated, so I think we're safe to go.

PR #850, opened 2021-05-13
KREST-1187: Propagate client.sasl.jaas.config to Schema Registry.

rigelbm:
Schema Registry client needs `sasl.jaas.config` when `basic.auth.credentials.source=SASL_INHERIT`.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/SchemaRegistrySaslInheritTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/SchemaRegistrySaslInheritTest.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.integration;
+
+import static org.junit.Assert.assertEquals;
+
+import com.fasterxml.jackson.databind.node.TextNode;
+import io.confluent.kafkarest.entities.EmbeddedFormat;
+import io.confluent.kafkarest.entities.v3.ProduceRequest;
+import io.confluent.kafkarest.entities.v3.ProduceRequest.ProduceRequestData;
+import io.confluent.kafkarest.entities.v3.ProduceResponse;
+import io.confluent.kafkarest.exceptions.v3.ErrorResponse;
+import io.confluent.kafkarest.testing.*;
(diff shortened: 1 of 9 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/integration/SchemaRegistrySaslInheritTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class SchemaRegistrySaslInheritTest {
```
dimitarndimitrov, line 42: I think it's important to comment the hell out of this test.

AFAIU, these are some of the big pieces of this daisy-chained craziness:
1. The new `PropertyFileLoginModule` fixture uses `JaasUtils.JAVA_LOGIN_CONFIG_PARAM` to register a JAAS config pointing to a property file with auth credentials for when JAAS is later used for authentication. This is the first step to be performed only because it requires a system property to be set. It's not used until later.
2. The Kafka fixture enables SASL and creates 2 super users with the credentials from the JAAS config.
3. Schema Registry and Kafka REST do some config magic which seems almost understandable, but I think we should ideally comment these extensively. Some questions around these:
    - What exactly happens when we set `SASL_INHERIT` in the SR basic auth configs in Kafka REST and `USER_INFO` in the SR basic auth configs in SR?
    - Where is the  `sasl.jaas.config` property populated, in order to be inherited down to Schema Registry? Does Kafka when it finds the JAAS system property?
    - Why is `producer.max.block.ms` needed in the Kafka REST config?
    - Does Kafka REST actually use JAAS, or just logs into Kafka with its configured superuser?

And at some point I think we need a tech talk or a very extensive wiki on these.

dimitarndimitrov (approved):
+1, because I very much believe that this works.

See below for a comment about how far I am from what I would have really wanted - to understand this in full detail.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/SchemaRegistrySaslInheritTest.java
rigelbm: 2. The credentials come from the `addUser` calls. The Kafka fixture does not use the JAAS file at all.
3.1. The USER_INFO is set as a schema registry *client* config. It means the `SchemaRegistryClient` returned by `getClient()` will use the `basic.auth.user.info` config to obtain SR credentials (also set as a *client* config).
3.2. SASL_INHERIT means the Kafka REST server will use `client.sasl.jaas.config` to obtain SR credentials. That config is set automagically in the kafkaRest fixture based on the user set in `setKafkaUser()` (if the kafka fixture is using SASL auth).
3.3. The test uses a produce call, If there's any problem with the broker, `producer.max.block.ms` makes sure the producer fails fast. The default deadline is 60s.
3.4. Kafka REST logs into Kafka with its configured superuser *using* JAAS (as mentioned in 3.2).

The meaning of a lot of these configurations is explained in our external docs. I added some documentation to the fixtures to hopefully clear a bit of the magic.

PR #855, opened 2021-05-26
ESCALATION-5224: URL encode query param values in returned URLs.

rigelbm:


kafka-rest/src/test/java/io/confluent/kafkarest/response/UrlFactoryImplTest.java
```diff
@@ -256,4 +256,20 @@ public void testCreateHostAdvertisedListenerAndListenerReturnsRelativeToAdvertis
...
+    expect(requestUriInfo.getAbsolutePath())
+        .andStubReturn(URI.create("http://1.2.3.4:1000/xxx/yyy"));
+    expect(requestUriInfo.getBaseUri()).andReturn(URI.create("http://1.2.3.4:1000/"));
+    replay(requestUriInfo);
+
+    UrlFactory urlFactory =
+        new UrlFactoryImpl("hostname", 2000, emptyList(), emptyList(), requestUriInfo);
+    UrlBuilder urlBuilder = urlFactory.newUrlBuilder();
+
+    String url = urlBuilder.appendPathSegment("foobar").putQueryParameter("foo", "b a r").build();
```
dimitarndimitrov, line 271: nit: Maybe include multiple query params, to exercise the appending branch too.

kafka-rest/src/main/java/io/confluent/kafkarest/response/UrlBuilder.java
```diff
@@ -65,6 +65,14 @@ public String build() {
 
     public abstract String getValue();
 
+    public final String encode() {
+      try {
+        return getKey() + "=" + URLEncoder.encode(getValue(), "UTF-8");
+      } catch (UnsupportedEncodingException  e) {
```
dimitarndimitrov, line 71: super-nit: Redundant space before `e`.

kafka-rest/src/main/java/io/confluent/kafkarest/response/UrlBuilder.java
```diff
@@ -65,6 +65,14 @@ public String build() {
 
     public abstract String getValue();
 
+    public final String encode() {
```
dimitarndimitrov, line 68: nit: Maybe rename to `asStringWithEncodedValue`?

kafka-rest/src/main/java/io/confluent/kafkarest/response/UrlBuilder.java
```diff
@@ -65,6 +65,14 @@ public String build() {
...
+      try {
+        return getKey() + "=" + URLEncoder.encode(getValue(), "UTF-8");
```
dimitarndimitrov, line 70: If it was possible to get a `null` value, this will now throw (where it wouldn't before).

AFAICS `AutoValue` is an earlier stop for potential `null` fields, and parsing the URL also probably defaults to an empty `String` instead of a `null`, but I thought I would point it out, in case I'm missing something.

dimitarndimitrov (approved):
+1.

kafka-rest/src/main/java/io/confluent/kafkarest/response/UrlBuilder.java
rigelbm: How come it was possible to have `null` value before? `AutoValue_UrlBuilder_QueryParameter(key, value)` will throw NPE for `null` key or value, because neither `getKey()` or `getValue()` are marked as `@Nullable`.

kafka-rest/src/main/java/io/confluent/kafkarest/response/UrlBuilder.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/response/UrlBuilder.java
rigelbm: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/response/UrlFactoryImplTest.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/response/UrlBuilder.java
dimitarndimitrov: > If it was possible...

and

> AFAICS AutoValue is an earlier stop for potential null fields

are saying "I don't think it's actually possible" :)

That's also why I ended with

> ...but I thought I would point it out, in case I'm missing something.

rigelbm: For some reason my brain skipped the "If" :)

PR #858, opened 2021-06-09
KREST-728: Add OpenAPI spec for Produce V3.

rigelbm:


api/v3/openapi.yaml
```diff
@@ -627,6 +627,45 @@ paths:
...
+      summary: 'Produce records to the given topic.'
+      description: '<p>Produce records to the given topic, returning delivery reports for each
+                    record produced.</p>
+                    <p>This is a streaming API. To use it in streaming mode, you need to specify the
+                    "Transfer-Encoding: chunked" header. For as long as the connection is kept open,
+                    the server will keep accepting records. For each record sent to the server, the
+                    server will asynchronously send back a delivery report, in the same order.
+                    Records are streamed to and from the server as Concatenated JSON.</p>
+                    <p>Errors are reported per record. The HTTP status code will be HTTP 200 OK as
+                    long the connection is successfully established.</p>'
```
dimitarndimitrov, line 646: nit: as long -> as long as

api/v3/openapi.yaml
```diff
@@ -627,6 +627,45 @@ paths:
...
+    parameters:
+      - $ref: '#/components/parameters/ClusterId'
+      - $ref: '#/components/parameters/TopicName'
+      - $ref: '#/components/parameters/PartitionId'
+
+    post:
+      summary: 'Produce records to the given topic.'
+      description: '<p>Produce records to the given topic, returning delivery reports for each
+                    record produced.</p>
+                    <p>This is a streaming API. To use it in streaming mode, you need to specify the
```
dimitarndimitrov, line 640: nit: _This is a streaming API_ followed by _To use it in streaming mode_ sounds a bit confusing to me. If this is a streaming API, is there any other mode but streaming mode?

api/v3/openapi.yaml
```diff
@@ -627,6 +627,45 @@ paths:
...
+      - $ref: '#/components/parameters/ClusterId'
+
+    get:
+      summary: 'Get All Topic Configs'
+      description: 'Returns all topic configurations for topics hosted by the specified cluster.'
+      tags:
+        - Configs
+      responses:
+        '200':
+          $ref: '#/components/responses/ListTopicConfigsResponse'
```
dimitarndimitrov, line 667: nit: This is just fixing an unrelated omission, right?

api/v3/openapi.yaml
```diff
@@ -627,6 +627,45 @@ paths:
         '200':
           $ref: '#/components/responses/GetReplicaResponse'
 
+  /clusters/{cluster_id}/topics/{topic_name}/records:
+    parameters:
+      - $ref: '#/components/parameters/ClusterId'
+      - $ref: '#/components/parameters/TopicName'
+      - $ref: '#/components/parameters/PartitionId'
```
dimitarndimitrov, line 634: I'm not sure why is the `PartitionId` param referred here - isn't it encapsulated in the `ProduceRequest`?

api/v3/openapi.yaml
```diff
@@ -1658,6 +1815,76 @@ components:
...
+                    {"partition_id":1}{"partition_id":2}. Delivery reports will be concatenated on
+                    the same order as the records are sent. See examples for the options available.'
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/ProduceRequest'
+          examples:
+            binary_and_avro_with_subject_and_raw_schema:
+              description: 'If using type: "BINARY" or type: "JSON", or using schema, type is
+                            required.'
```
dimitarndimitrov, line 1830: This doesn't read very well to me.
AFAIU if the key or value content uses one of the types where the serde procedure is known beforehand - binary or JSON - then we need `type` to designate that we don't need a schema lookup to deserialize.
If the key or value content uses one of the schema-serialized types, like Avro or Protobuf, I'm not sure why the type is needed too - isn't the schema reference enough?
Also _or using schema_ seems to be a bit at odds with _If not setting type, the record is assumed to use a schema_ from the next example.
There's probably a very good explanation behind all of this, but I don't think it's very obvious from the descriptions here and below.

api/v3/openapi.yaml
```diff
@@ -2725,6 +2952,28 @@ components:
...
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/ProduceResponse'
+          example:
+            cluster_id: 'cluster-1'
+            topic_name: 'topic-1'
+            partition_id: 1
+            offset: 0
+            timestamp: '2021-02-05T19:14:42Z'
```
dimitarndimitrov, line 2966: nit: Is the timestamp format always this? AFAIR a server-prodived `Instant` is the source of the timestamp, and it gets formatted to a human-readable string somewhere in Kafka REST.

londoncalling (approved):
LGTM

api/v3/openapi.yaml
rigelbm: By "schema", I mean "the schema field". You when you the raw schema (so that Kafka REST register the schema for you), you need to say what type that schema is. We can't guess from the raw schema string (we probably could, but we don't).

api/v3/openapi.yaml
axisc: I'm assuming that this can be used in non-streaming mode too...

if yes, perhaps we can word it as "This API can be used in streaming mode by setting "Transfer-Encoding: chunked" on the header".

api/v3/openapi.yaml
rigelbm: Done.

api/v3/openapi.yaml
rigelbm: Done.

api/v3/openapi.yaml
rigelbm: That slipped through. :|

api/v3/openapi.yaml
rigelbm: Changed from `schema` to `the schema field`.

api/v3/openapi.yaml
rigelbm: See https://github.com/confluentinc/kafka-rest/blob/master/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java#L129.

api/v3/openapi.yaml
rigelbm: This was deleted on another PR, and came back when merging. I missed it, sorry.

api/v3/openapi.yaml
axisc: 🔥

PR #862, opened 2021-07-01
KREST-1458: Always allow OPTIONS requests.

rigelbm:
Because OPTIONS requests are handled by the default jersey
OptionsMethodProcessor, the allowlist feature does not play nice with
it, as it will block anything that is not annotated with something
that is in the allowlist, and the jersey processor is clearly not. We
also have the problem that this processor handles OPTIONS requests for
all endpoints, and we don't have a good way on intercepting it making
it work nice with the allowlist.

The current stop gap solution is to allow OPTIONS for all
endpoints. We can investigate better solutions afterwards.


dimitarndimitrov (approved):
+1.

Can you please file a JIRA for the follow-up?

PR #875, opened 2021-08-10
Pass KafkaRestContext to HK2 via ContainerRequestContext.

rigelbm:
Currently the mechanism used by the various request
processing stages (request/response filters/interceptors and request
handlers) to communicate with each other is
KafkaRestContextProvider. This class owns a thread-local, which is
supposed to keep state per-request-handling-thread (although that's
not 100% true in the presence of async requests). If this thread-local
is not set, than Kafka REST fallsback to a default KafkaRestContext,
which is kept in the KafkaRestContextProvider in a static variable, to
preserve singleton property. This has worked fine so far because up to
this point there has always been a single KafkaRestApplication running
per JVM, but that assumption is not valid anymore. Since the static
cannot isolation the singleton instances accross multiple
KafkaRestApplication instances in the same JVM, another approach is
taken here.

What this PR does is to use instead the ContainerRequestContext to
keep the state per-request state. This approach has the nice benefit
of working, regardless of the request being suspended or not. The
default singleton KafkaRestContext is owned by HK2 now, and will be
isolated per injector (which will be one per application).

A follow-up PR will be made to delete KafkaRestContextProvider, once
all usages have been removed.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
@@ -17,10 +17,14 @@ package io.confluent.kafkarest.backends.kafka;
 
 import static java.util.Objects.requireNonNull;
 
+import io.confluent.kafkarest.DefaultKafkaRestContext;
+import io.confluent.kafkarest.KafkaRestConfig;
 import io.confluent.kafkarest.KafkaRestContext;
-import io.confluent.kafkarest.extension.KafkaRestContextProvider;
 import javax.inject.Inject;
 import javax.inject.Provider;
+import javax.inject.Singleton;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.core.Context;
 import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.producer.Producer;
 import org.glassfish.hk2.api.Factory;
@@ -33,7 +37,7 @@ import org.glassfish.jersey.process.internal.RequestScoped;
  *
  * <p>Right now this module does little but delegate to {@link KafkaRestContext}, since access to
  * Kafka is currently being configured there. It's the author's intention to move such logic here,
- * and eliminate {@code KafkaRestContext}, once dependence injection is properly used elsewhere.</p>
+ * and eliminate {@code KafkaRestContext}, once dependence injection is properly used elsewhere.
  */
 public final class KafkaModule extends AbstractBinder {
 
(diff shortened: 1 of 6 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
```diff
@@ -33,35 +37,48 @@
  *
  * <p>Right now this module does little but delegate to {@link KafkaRestContext}, since access to
  * Kafka is currently being configured there. It's the author's intention to move such logic here,
- * and eliminate {@code KafkaRestContext}, once dependence injection is properly used elsewhere.</p>
+ * and eliminate {@code KafkaRestContext}, once dependence injection is properly used elsewhere.
```
mhowlett, line 40: the style checker was not unhappy about this?

rigelbm: Google Java Style forbids the closing `</p>`. This has been auto-formatted. Check-style is fine with it, since it's optional.

mhowlett (approved):
i have not reviewed this super carefully, but have confidence it's well founded and happy for it to be merged / lgtm if you don't feel you need a more careful review (which i suspect is the case).

kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
```diff
@@ -33,35 +37,48 @@
...
+      this.requestContext = requireNonNull(requestContext);
+      this.defaultContext = new DefaultKafkaRestContext(config);
+    }
+
     @Override
     public KafkaRestContext provide() {
-      return KafkaRestContextProvider.getCurrentContext();
+      KafkaRestContext context =
+          (KafkaRestContext) requestContext.get().getProperty(KAFKA_REST_CONTEXT_PROPERTY_NAME);
+      return context != null ? context : defaultContext;
```
dimitarndimitrov, line 75: So just for my own understanding, would we be hitting always the same factory singleton instance (and therefore the same provider and default context instance), but `requestContext.get()` will be returning request-scoped instance (different `ContainerRequestContext` for different requests)?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/KafkaRestContextIntegrationTest.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.kafka.clients.admin.DescribeClusterResult;
+import org.apache.kafka.clients.producer.Producer;
+import org.junit.Test;
+
+public class KafkaRestContextIntegrationTest extends ClusterTestHarness {
+
+  /** HTTP 418 I'm a teapot */
+  private static final int I_M_A_TEAPOT_STATUS_CODE = 418;
+
+  private DescribeClusterResult describeClusterResult;
```
dimitarndimitrov, line 51: nit: I think this is redundant, hidden by a local variable.

dimitarndimitrov (commented):
I am almost sure `KafkaRestContextProvider` could and should be removed immediately. Is there a specific reason not to want to do that here (e.g. to minimize the impact and risk of each separate change)?
I think now the only piece remaining is the cleanup filter, as getting the context is also fully removed from the provider, so it's especially weird to leave just the cleanup part.

rigelbm:
> I am almost sure `KafkaRestContextProvider` could and should be removed immediately. Is there a specific reason not to want to do that here (e.g. to minimize the impact and risk of each separate change)?
> I think now the only piece remaining is the cleanup filter, as getting the context is also fully removed from the provider, so it's especially weird to leave just the cleanup part.

I want to avoid breaking the build. That's why we use both ways in downstream dependencies. This PR switches what is used. Then we can have follow up PRs to remove usages of old, and finally a PR to delete old.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/KafkaRestContextIntegrationTest.java
rigelbm: Done.

dimitarndimitrov (approved):
+1.

Let's take the context provider discussion offline. Please check whether the CI failures are benign, as I don't have my computer in front of me right now, so browsing Jenkins is a pain.

kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
```diff
@@ -33,35 +37,48 @@
  *
  * <p>Right now this module does little but delegate to {@link KafkaRestContext}, since access to
  * Kafka is currently being configured there. It's the author's intention to move such logic here,
- * and eliminate {@code KafkaRestContext}, once dependence injection is properly used elsewhere.</p>
+ * and eliminate {@code KafkaRestContext}, once dependence injection is properly used elsewhere.
  */
 public final class KafkaModule extends AbstractBinder {
 
+  public static final String KAFKA_REST_CONTEXT_PROPERTY_NAME = "io.confluent.kafkarest.context";
```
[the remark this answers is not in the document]
rigelbm: Done.

rigelbm:
One test failure, our good friend ` io.confluent.kafkarest.v2.KafkaConsumerManagerTest.testBackoffMsControlsPollCalls`.

PR #877, opened 2021-08-11
KREST-1574 Mark consumer lag as preview in openapi docs

ehumber:
Mark the three consumer lag endpoints as preview, ready for kafka-rest cloud documentation update.

There will be a follow on PR for the CP changes, as these will need to be different as there is no x-lifecycle tag available at present,  and they will also need to go into the 700 branch.

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -344,7 +344,7 @@ paths:
           $ref: '#/components/responses/ListConsumersResponse'
 
   /clusters/{cluster_id}/consumer-groups/{consumer_group_id}/lag-summary:
-    x-audience: component-internal
+    x-lifecycle-stage: Preview
     parameters:
       - $ref: '#/components/parameters/ClusterId'
       - $ref: '#/components/parameters/ConsumerGroupId'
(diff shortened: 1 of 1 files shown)


dimitarndimitrov:
Hey @ehumber - should this start from `7.0.0-post` or it's intentionally starting from `master` (maybe because of the tag thing)?

ehumber:
> Hey @ehumber - should this start from `7.0.0-post` or it's intentionally starting from `master` (maybe because of the tag thing)?

I intentionally started it from master because the change is for the cloud docs only, and my understanding is that they are taken from master.  

The CP doc update would ned to come from 7.0.0-post, but the change for CP will need to be different as the docs don't understand `x-lifecycle`.  I'll do the CP changes later, as the urgency atm is to get the cloud docs sorted.

The  understanding of the copy from master (rather than a branch) for cloud docs comes from line 23 of this PR https://github.com/confluentinc/ce-kafka-rest/pull/211/files

ehumber:
For reference, the PR that added in the x-lifecycle-stage tag functionality to confluentinc/api was https://github.com/confluentinc/api/pull/62

rigelbm (approved):
LGTM

PR #878, opened 2021-08-11
KREST-574 Hide unsupported api endpoints in docs

ehumber:


api/v3/openapi.yaml
```diff
@@ -395,6 +406,7 @@ paths:
           $ref: '#/components/responses/GetConsumerResponse'
 
   /clusters/{cluster_id}/consumer-groups/{consumer_group_id}/consumers/{consumer_id}/assignments:
+    x-audience: component-internal
```
rigelbm, line 409: I think this should be in.

api/v3/openapi.yaml
```diff
@@ -410,6 +422,7 @@ paths:
           $ref: '#/components/responses/ListConsumerAssignmentsResponse'
 
   /clusters/{cluster_id}/consumer-groups/{consumer_group_id}/consumers/{consumer_id}/assignments/{topic_name}/partitions/{partition_id}:
```
rigelbm, line 424: It think this should be in.

rigelbm (approved):
LGTM

PR #879, opened 2021-08-12
Remove KafkaRestContextProvider.

rigelbm:
KafkaRestContextProvider is not used anymore. ContainerRequestContext is used instead.


rigelbm:
io.confluent.kafkarest.v2.KafkaConsumerManagerTest.testBackoffMsControlsPollCalls failed. Ignoring.

dimitarndimitrov (approved):
+1.

I have to say, it's really confusing that clearing the TL state and cleaning up the context were so intertwined before. Can you please add some Javadoc to `KafkaRestContext` explaining how the latter is an important responsibility of subclasses now, and maybe to the `DefaultKafkaRestContext` explaining how this won't be cleaned up during the instance lifetime?

rigelbm:
> +1.
> 
> I have to say, it's really confusing that clearing the TL state and cleaning up the context were so intertwined before. Can you please add some Javadoc to `KafkaRestContext` explaining how the latter is an important responsibility of subclasses now, and maybe to the `DefaultKafkaRestContext` explaining how this won't be cleaned up during the instance lifetime?

We have this already: https://github.com/confluentinc/kafka-rest/blob/master/kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java#L80. Is it good enough?

PR #880, opened 2021-08-13
KREST-1574 Add icons to descriptions to indicate preview

ehumber:
KREST-1574 Add icons to descriptions to indicate preview

Also fix a linter fail for a missing size parameter


ehumber:
image

PR #881, opened 2021-08-18
Fix tests to support ProxyConnectionFactory

bob-barrett:


pom.xml
```diff
@@ -7,7 +7,8 @@
     <parent>
         <groupId>io.confluent</groupId>
         <artifactId>rest-utils-parent</artifactId>
-        <version>[7.1.0-0, 7.1.1-0)</version>
+        <version>7.1.0-0</version>
+        <!-- <version>[7.1.0-0, 7.1.1-0)</version> -->
```
dimitarndimitrov, line 11: Do we need to submit that temporarily? If yes, can you please add a comment explaining why? If not, is that a leftover from some testing?

kafka-rest/src/test/java/io/confluent/kafkarest/testing/KafkaRestFixture.java
```diff
@@ -76,6 +76,11 @@ public void before() throws Exception {
     server = application.createServer();
     server.start();
     baseUri = server.getURI();
+    if (certificates != null && baseUri.getScheme().equals("http")) {
+      // If PROXY protocol is enabled, the base URI will come back as http because the first connection
+      // factory is not the SSL factory. If we want SSL, explicitly set the scheme here
+      baseUri = new URI(baseUri.toString().replace("http", "https"));
```
dimitarndimitrov, line 82: nit: We have loaded the JAX-RS classes for Kafka REST, so it might be a tad nicer to use `javax.ws.rs.core.UriBuilder` here:
```
baseUri = UriBuilder.fromUri(baseUri).scheme("https").build();
```

dimitarndimitrov (commented):
Hey @bob-barrett, is it that because of some specificity only these tests will be affected by adding a `ProxyConnectionFactory` to our connection factory pipeline? I.e. I'm not sure I understand why is this a problem just for some tests.

pom.xml
bob-barrett: Yeah this is just so that it will consume the `rest-utils` local changes when running tests, I'll get rid of it before I convert this to a real PR.

dimitarndimitrov (commented):
@bob-barrett Is this still relevant? I assume it's not and we can close it now.

PR #882, opened 2021-08-18
KREST-574 Make openapi.yaml match what is needed for cloud docs

ehumber:
Title needs to change

Contact details are required

x-api/x-audience-x-tags are required

Include both cloud and platform servers, tag the platform ones as x-audience: component-internal so that they are excluded from the cloud docs, and the could one as x-audience: business-unit-internal (arbitrary choice from the options) so it can be excluded from the platform one

Add /kafka/v3 at the front of each path to match the cloud requirements

Change the tags (which make the titles) to all include (v3) to match the cloud requirements

Add `operationId ` which is just a unique string per endpoint, as required by cloud

Add extra `x-lifecycle-stage: Preview` tags at the `get` level for the lag endpoints, and link to the icons indicating preview status in the description

Added in linebreaks to stop the linter complaining

Changed `enum` to `x-extensible-enum` where it was possible to do this (the fields need to be SNAKE_CASE) to allow the change, as otherwise the linter broke.  Note that there are a couple of enums that I think could be extensible_enums that weren't changed over, but for now I am just making our docs match what has been published.  We can address any missing ones as part of https://confluentinc.atlassian.net/wiki/spaces/~70581479/pages/2575204659/Kafka+Admin+REST+v3+API+Linting+Errors

Added a missing, required, size parameter to ProduceResponseData as the linter spotted this and failed

Updated example urls to not be localhost.  eg `localhost:PORTNUMBER/v3/` has changed to be `example.us-central1.gcp.confluent.cloud/kafka/v3`

`CreateAclRequest` had a couple of bugs.  The resource name has to be `kafka-cluster`, not the arbitrary `cluster-1`.  And the principal requires a : in the middle, and is expecting to be of the `format principalType:principalName`

Added tags with the section descriptions as required for the cloud docs.

Added in public usage icon

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -1,11 +1,20 @@
 openapi: 3.0.0
 
 info:
-  title: Kafka HTTP API
+  title: REST Admin API
   version: 3.0.0
+  contact:
+    name: Kafka REST Team
+    url: https://confluent.slack.com/app_redirect?channel=kafka-rest-eng
+    email: kafka-clients-proxy-team@confluent.io
+  x-api-id: 499e3476-71e0-4d6f-b4f9-6776cec6df27
+  x-api-group: v3
+  x-audience: external-public
+  x-tag-group: Cluster Admin for Kafka (%s)
 
 servers:
   - url: "{protocol}://{address}:{port}/v3"
+    x-audience: component-internal
     description: Kafka REST Proxy. See https://docs.confluent.io/current/kafka-rest/index.html.
     variables:
       address:
@@ -20,7 +29,9 @@ servers:
         enum:
           - http
           - https
+
   - url: "{protocol}://{address}:{port}/kafka/v3"
+    x-audience: component-internal
     description: Kafka REST Proxy as part of Confluent Server.
     variables:
       protocol:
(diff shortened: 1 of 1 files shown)


ehumber:
Updates:

/kafka/v3 removed from the paths as agreed with Rigel (these will be added back later based on the section after the hostname:port in the server url (which is now https://pkc-00000.region.provider.confluent.cloud/kafka/v3 for the CP endpoint)

Changed the examples to be https://pkc-00000.region.provider.confluent.cloud

The API team added in the "Generally Available" icon to all the APIs, as well as some descriptions so I've added these to our docs too

The lag endpoints have been changed from preview to early access, and include a mailto that should set up an email to ccloud-api-access v2-early-access@confluent.io  that says 

image says: Request Access To Cluster Admin For Kafka v3
mailto is: ccloud-api-access consumerlag-early-access@confluent.io
subject is: Request to join v3 API Early Access
email body:  I’d like to join the Confluent Cluster Admin For Kafka Early Access for v3 to provide early feedback on consumer lag apis! My Cloud Organization ID is <retrieve from https://confluent.cloud/settings/billing/payment>.)

rigelbm (approved):
LGTM

PR #888, opened 2021-08-31
Add Google-Java-Format to Maven configuration.

rigelbm:
This PR does two things:

1. Adds the spotless plugin to maven and configures it to run google-java-format in the project. It will run by default in the verify phase. You can manually run it using `mvn spotless:check` to look for problems, and `mvn spotless:apply` to format.
2. Run `google-java-format` in the entire project.

I also had to add a suppression for single-line-javadoc in the check-style configuration, since check-style doesn't like single line javadocs, and Google Java Format does.

It's a big change, but hopefully one that we will need to do only once. Jenkins should make sure the code will always be formatted correctly from now on (+ maven won't let you build if it isn't).


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -388,336 +356,281 @@
...
-        Type.BOOLEAN,
-        API_V3_ENABLE_DEFAULT,
-        Importance.LOW,
-        API_V3_ENABLE_DOC);
+            KAFKAREST_PORT_DEFAULT,
+            KAFKAREST_LISTENERS_DEFAULT,
+            String.join(",", Versions.PREFERRED_RESPONSE_TYPES),
+            MediaType.APPLICATION_JSON,
+            METRICS_JMX_PREFIX_DEFAULT_OVERRIDE)
+        .define(ID_CONFIG, Type.STRING, ID_DEFAULT, Importance.HIGH, ID_CONFIG_DOC)
```
ehumber, line 364: Can the formatting be overriden here so all the defines match?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerConfigManagerImpl.java
```diff
@@ -48,28 +47,35 @@
...
-                    config.isSensitive(),
-                    config.getSource(),
-                    config.getSynonyms()
-              )).collect(Collectors.toList())
-            )
-        ));
+            clusterId,
+            brokerResources,
+            BrokerConfig.builder().setClusterId(clusterId).setBrokerId(-1))
+        .thenApply(
```
ehumber, line 58: This one looks easier to read before the changes.

dimitarndimitrov (commented):
SGTM overall, but while we're at it:
- Does it make sense to also enable Spotbugs (there's an internal ticket about this)?
- Should we also enable this on the similar internal projects downstream, where there are also multiple contributors from multiple teams?
- Can you please review the suppressed Checkstyle rules? I think at least a couple of them have become outdated by now.

rigelbm:
> * Can you please review the suppressed Checkstyle rules? I think at least a couple of them have become outdated by now.

Done

rigelbm:
> * Does it make sense to also enable Spotbugs (there's an internal ticket about this)?
I took a look and it seems non-trivial, so I will leave it as a follow-up change.

> * Should we also enable this on the similar internal projects downstream, where there are also multiple contributors from multiple teams?
After this PR merges, we can reference the configuration here in `ce-kafka-rest`. We will need a separate PR for `ce-kafka-http-server`, since it does not have `kafka-rest` as parent.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Google Java Format is opinionated in the fact that you can't override or suppress anything.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/BrokerConfigManagerImpl.java
rigelbm: I think here the new formatting actually exacerbates the underlying problem, which is that this lambda->lambda->lambda is too nested. Ugly things should look ugly :)

ehumber (approved):

PR #890, opened 2021-09-06
KREST-1673 Add error responses to the OpenAPI spec.

dimitarndimitrov:
This adds some common error responses to all endpoints in our OpenAPI spec.
Unfortunately OpenAPI does not provide a nice way to handle these in a single place, so there's a lot of duplication.
- See https://github.com/OAI/OpenAPI-Specification/issues/398, https://github.com/OAI/OpenAPI-Specification/issues/521, https://github.com/OAI/OpenAPI-Specification/issues/566.

The error responses also include some examples that will hopefully be useful for users and developers.
However the examples are by no means exhaustive.

Tested locally with:
```
% swagger-cli validate ~/work/dev/kafka-rest/api/v3/openapi.yaml -d
{
  "command": "validate",
  "file": "/Users/ddimitrov/work/dev/kafka-rest/api/v3/openapi.yaml",
  "options": {
    "schema": true,
    "spec": true,
    "format": 2,
    "type": "json",
    "wrap": null,
    "debug": true
  }
}
/Users/ddimitrov/work/dev/kafka-rest/api/v3/openapi.yaml is valid
```

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -49,7 +49,7 @@ servers:
 
   - url: https://pkc-00000.region.provider.confluent.cloud/kafka/v3
     x-audience: business-unit-internal
-    description: Confluent Cloud REST Endpoint. For example https://pkc-00000.region.provider.confluent.cloud 
+    description: Confluent Cloud REST Endpoint. For example https://pkc-00000.region.provider.confluent.cloud
 
 paths:
   /clusters:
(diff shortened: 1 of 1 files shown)


api/v3/openapi.yaml
```diff
@@ -894,6 +1282,17 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/ProduceResponse'
+        '400':
+          $ref: '#/components/responses/BadRequestErrorResponse'
+        '401':
+          $ref: '#/components/responses/UnauthorizedErrorResponse'
+        # Apparently trying to produce to non-existing topic with auto-create enabled doesn't
```
ehumber, line 1289: Should this be a follow on jira rather than a comment?

ehumber (approved):

ehumber:
LGTM

PR #892, opened 2021-09-07
KREST-1654 JMX metrics for produce v3

thomaskwscott:
This PR adds global JMX metrics to Rest Proxy for the purpose of monitoring in cloud during the dedicated preview. For more information see the 1 pager here: 

https://confluentinc.atlassian.net/wiki/spaces/MMA/pages/2497381076/Engineeering+1+pager+-+REST+Proxy+produce+GA#Engineeering1pager-RESTProxyproduceGA-Metrics%3A

Tested with unit tests and manually using trogdor workloads/jconsole.


ehumber:
LGTM but I think you will also need a review from Rigel or Dimitar.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,396 @@
...
+        .withCumulativeCount()
+        .build();
+
+    new SensorBuilder(metricMBean)
+        .withName(ProducerMetricsRegistry.RECORD_ERROR_RATE)
+        .withDoc(ProducerMetricsRegistry.RECORD_ERROR_RATE_DOC)
+        .withRate()
+        .build();
+
+    new SensorBuilder(metricMBean)
```
ehumber, line 147: I added a windowed counter via https://github.com/confluentinc/rest-utils/pull/284/files so that we can alert on absolute values for errors, rather than aggregated rates.  Might be worth doing the same here.

thomaskwscott: @ehumber sorry for the delay, I've added the windowed metrics now.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,414 @@
...
+
+    // Clear out any previous metrics and sensors that might have been present
+    if (metricMBean != null) {
+      metricMBean.close();
+    }
+
+    // request metrics
+    new SensorBuilder(metricMBean, ProducerMetricsRegistry.REQUEST_SENSOR)
+        .addRate(ProducerMetricsRegistry.REQUEST_RATE, ProducerMetricsRegistry.REQUEST_RATE_DOC)
+        .addCumulativeCount(
```
ehumber, line 95: We removed the cumulative counts from rest-utils in the end, and just kept the windowed ones.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,414 @@
...
+            ProducerMetricsRegistry.REQUEST_TOTAL_WINDOWED_DOC)
+        .build();
+
+    new SensorBuilder(metricMBean, ProducerMetricsRegistry.REQUEST_SIZE_SENSOR)
+        .addAvg(
+            ProducerMetricsRegistry.REQUEST_SIZE_AVG, ProducerMetricsRegistry.REQUEST_SIZE_AVG_DOC)
+        .addCumulativeSum(
+            ProducerMetricsRegistry.REQUEST_SIZE_TOTAL,
+            ProducerMetricsRegistry.REQUEST_SIZE_TOTAL_DOC)
+        .addWindowedSum(
```
ehumber, line 108: Should this be addWindowedCount rather than the WindowedSum (I'm assuming a Windowed sum is a running total of all the separate windowedCounts so far)?  

The changes in rest-utils didn't add any counter based metrics to the sizes (just the count of requests and count of errors), so I guess it depends on what we think adding up all the sizes of the requests so far will tell us, and if that's useful.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
thomaskwscott: removed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
thomaskwscott: these were mirroring the topic level bytes metrics from the other clients but I can't think of a major use for them for cloud monitoring so have removed.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+        }
+      }
+    }
+  }
+
+  // I'm not going to lie this exists only to defeat checkstyle ;-)
+  public static class SensorBuilder {
+
+    private ProduceMetricMBean bean;
+    private Sensor sensor;
```
dimitarndimitrov, line 254: nit: Shouldn't these be `final`?

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+      Sensor sensor = metrics.sensor(name);
+      return sensor;
+    }
+
+    /** Remove all sensors and metrics associated with this group. */
+    @Override
+    public synchronized void close() {
+      for (MetricName metricName : new HashSet<>(metrics.metrics().keySet())) {
+        if (beanCoordinate.beanName.equals(metricName.group())
+            && beanCoordinate.tags.equals(metricName.tags())) {
```
dimitarndimitrov, line 243: While `BeanCoordinate` itself can work with `null` fields, this usage will actually throw NPE. Let's either add `requireNonNull` in the `BeanCoordinate` ctor, or if the `metricName` getters are guaranteed to not return `null`, reverse the relation.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+     * @return The sensor
+     */
+    synchronized Sensor sensor(String name) {
+      Sensor sensor = metrics.sensor(name);
+      return sensor;
+    }
+
+    /** Remove all sensors and metrics associated with this group. */
+    @Override
+    public synchronized void close() {
```
dimitarndimitrov, line 240: I'm not sure what are we trying to achieve with the mutual exclusion here. In case it's mostly avoiding an internal problem similar to `ConcurrentModificationException` I guess it's OK, but if not, I think we should clarify the lifecycle of this and enforce it with a simple state machine.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+  // request
+  public static final String REQUEST_RATE = "request-send-rate";
+  public static final String REQUEST_RATE_DOC = "The average number of requests sent per second.";
+
+  public static final String REQUEST_SIZE_AVG = "request-size-avg";
+  public static final String REQUEST_SIZE_AVG_DOC = "The average request size in bytes.";
+
+  public static final String REQUEST_TOTAL_WINDOWED = "request-total-windowed";
+  public static final String REQUEST_TOTAL_WINDOWED_DOC =
+      "The total number of requests sent within the given window.";
```
dimitarndimitrov, line 39: nit: I'd replace _total_ with _count_ here - the former is mostly used as an adjective to a metric type identifier, while the latter is mostly used as a metric type identifier. Same applies for `RESPONSE_TOTAL_WINDOWED` and `RECORD_ERROR_TOTAL_WINDOWED`.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+public class ProducerMetricsRegistry {
+
+  public static final String GROUP_NAME = "produce-api-metrics";
+
+  // sensor names
+  public static final String REQUEST_SENSOR = "request-sensor";
+  public static final String REQUEST_SIZE_SENSOR = "request-size-sensor";
+  public static final String RESPONSE_SENSOR = "response-sensor";
+  public static final String RESPONSE_SIZE_SENSOR = "response-size-sensor";
+  public static final String RECORD_ERROR_SENSOR = "record-error-sensor";
```
dimitarndimitrov, line 27: QQ: What does _record_ indicate here? That it's an aggregation of all errors related to produce, which might be split to request errors and response errors?

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+  public static final String REQUEST_LATENCY_SENSOR = "request-latency-sensor";
+
+  // request
+  public static final String REQUEST_RATE = "request-send-rate";
```
dimitarndimitrov, line 31: nit: I think _-send-_ is redundant. Same applies for `RESPONSE_RATE`.

kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
```diff
@@ -121,4 +127,20 @@ public void dispose(Producer<byte[], byte[]> producer) {
...
+
+  private static final class ProducerMetricsFactory implements Factory<ProducerMetrics> {
+
+    private ProducerMetrics producerMetrics;
+
+    @Override
+    public ProducerMetrics provide() {
+      producerMetrics = new ProducerMetrics(Time.SYSTEM);
+      return producerMetrics;
+    }
```
dimitarndimitrov, line 139: Given that this is used for DI purposes in a mostly concurrent matter, it's an interesting question how safe is this publishing. On first glance, `ProducerMetrics` is kinda safely-initialized, as it uses `final` fields, but the objects referred by the fields are actually mutable and are being mutated after the fields are initialized. We should ponder a bit more whether there might be a problem around that, although I think just not treating this as safely-initialized, and publishing via `volatile` would be good enough.

dimitarndimitrov (commented):
First pass, reviewing the changes up to `ProduceControllerImpl`.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+        mbean(ProducerMetricsRegistry.GROUP_NAME, Collections.emptyMap());
+
+    // Clear out any previous metrics and sensors that might have been present
+    if (metricMBean != null) {
+      metricMBean.close();
+    }
+
+    // request metrics
+    new SensorBuilder(metricMBean, ProducerMetricsRegistry.REQUEST_SENSOR)
+        .addRate(ProducerMetricsRegistry.REQUEST_RATE, ProducerMetricsRegistry.REQUEST_RATE_DOC)
```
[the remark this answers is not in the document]
ehumber: In counter to what Dimitar has said, why not use all the calculation logic that is built in to the metrics, to save us having to do it ourselves?  We export request rate for the other rest calls, I would propose that produce is no different.  

Is there a harm in adding it here?  (Especially if we don't export in dd-agent to start with, but the info is there if we need it).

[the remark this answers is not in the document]
thomaskwscott: @dimitarndimitrov the rate metrics are consistent with the other clients (and I believe we even kept them alongside windowedcounts upstream). If we were to remove any of them it should be the windowedcount metrics imo.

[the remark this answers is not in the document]
dimitarndimitrov: The reason I don't like the built-in `Rate` metric is that it's not very flexible and it's hard to interpret because it naturally results in double aggregation. It's a sliding window rate (I personally prefer exponential weighting) which in its most simple form just uses some default window and you have to look into `ce-kafka` to find what that window is.

I personally think that metric-wise we need very little more than histogrammed counts and latencies, if we can use a dynamic range histogram, with raw values as a close second (or first, if we can't use good histograms). Everything that can be inferred from that at a later point seems like a redundancy that we want to avoid at runtime.
- The same applies for the rates on the Admin APIs - I actually think we should calculate these from the request counts. My main concern with cumulative vs windowed counts was the potential loss of precision/fidelity if we were to use these for rate calculation, but that didn't turn out to be a problem.

I am very open to changing my mind on this, but if I happen to be pig-headed without even realising that, please do let me know that :D

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+import org.apache.kafka.common.metrics.stats.Max;
+import org.apache.kafka.common.metrics.stats.Rate;
+import org.apache.kafka.common.metrics.stats.WindowedCount;
+import org.apache.kafka.common.metrics.stats.WindowedSum;
+import org.apache.kafka.common.utils.Time;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProducerMetrics {
+  public static final String JMX_PREFIX = "kafka.rest";
```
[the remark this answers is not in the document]
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+            ProducerMetricsRegistry.REQUEST_LATENCY_MAX_DOC)
+        .addAvg(
+            ProducerMetricsRegistry.REQUEST_LATENCY_AVG,
+            ProducerMetricsRegistry.REQUEST_LATENCY_AVG_DOC)
+        .build();
+  }
+
+  private class BeanCoordinate {
+    private final String beanName;
+    private final Map<String, String> tags;
```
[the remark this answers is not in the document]
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
```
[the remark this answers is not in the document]
thomaskwscott: makes sense, autogen issues...

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+    BeanCoordinate beanCoordinate = new BeanCoordinate(groupName, tags);
+    beansByCoordinate.putIfAbsent(beanCoordinate, new ProduceMetricMBean(beanCoordinate));
+    return beansByCoordinate.get(beanCoordinate);
+  }
+
+  /** Stop and unregister the metrics from any reporters. */
+  public void stop() {
+    log.info("Unregistering produce metrics with JMX");
+    metrics.close();
+  }
```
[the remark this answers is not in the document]
thomaskwscott: we never call stop anyway (it was for cases where we would remove metrics during application lifecycle) so I've removed it.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+
+  /** Sets up a {@link ProduceMetricMBean} for each sensor */
+  private void setupMetricBeans() {
+    ProduceMetricMBean metricMBean =
+        mbean(ProducerMetricsRegistry.GROUP_NAME, Collections.emptyMap());
+
+    // Clear out any previous metrics and sensors that might have been present
+    if (metricMBean != null) {
+      metricMBean.close();
+    }
```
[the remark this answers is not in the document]
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+            ProducerMetricsRegistry.RECORD_ERROR_TOTAL_WINDOWED_DOC)
+        .build();
+
+    new SensorBuilder(metricMBean, ProducerMetricsRegistry.REQUEST_LATENCY_SENSOR)
+        .addMax(
+            ProducerMetricsRegistry.REQUEST_LATENCY_MAX,
+            ProducerMetricsRegistry.REQUEST_LATENCY_MAX_DOC)
+        .addAvg(
+            ProducerMetricsRegistry.REQUEST_LATENCY_AVG,
+            ProducerMetricsRegistry.REQUEST_LATENCY_AVG_DOC)
```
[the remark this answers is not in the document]
thomaskwscott: makes sense, I've added 95, 99 and 999

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+    private final BeanCoordinate beanCoordinate;
+
+    /**
+     * Create a produce metrics MBean.
+     *
+     * @param beanCoordinate the identifier of the bean; may not be null and must be valid
+     */
+    ProduceMetricMBean(BeanCoordinate beanCoordinate) {
+      Objects.requireNonNull(beanCoordinate);
+      this.beanCoordinate = beanCoordinate;
```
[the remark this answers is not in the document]
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    public void recordMetrics(String sensorName, double value) {
+      Sensor sensor =
+          metrics.getSensor(
+              String.join(":", JMX_PREFIX, ProducerMetricsRegistry.GROUP_NAME, sensorName));
+      if (sensor != null) {
+        sensor.record(value);
+      }
+    }
```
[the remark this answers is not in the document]
thomaskwscott: I think join is quicker than format but I get the point about concatenation every time. I've factored them out into separate methods against constants to avoid this.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,395 @@
...
+
+    /**
+     * The {@link Metrics} that this MBean belongs to. Do not use this to add {@link Sensor
+     * Sensors}, since they will not be removed when this group is{@link #close() closed}. Metrics
```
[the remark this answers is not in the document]
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
thomaskwscott: fixed both ways

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
thomaskwscott: it was just staying consistent with the java producer metrics but i've removed it now as we are not consistent elsewhere.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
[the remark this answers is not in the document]
thomaskwscott: @dimitarndimitrov My worry is that we don't offer any comparable metrics between other clients and the rest producer. All of the Confluent developed clients offer a rate metric defined in this way and I think it is expected. And, if present, should be consistent with the other clients.

kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
thomaskwscott: as per discussion on slack I've made producerMetrics volatile and guarded it with an IllegalStateException if provide() is called with a non null metrics object. I've manually tested this with multiple connections but not with multiple contexts, I believe that test would be in ce-kafka-rest as we don't do multi context in kafka-rest.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -128,10 +140,26 @@ public void produce(
...
+              return result;
+            })
+        .thenApply(
+            result -> {
+              ProduceResponse response =
+                  toProduceResponse(
+                      clusterId, topicName, keyFormat, keySchema, valueFormat, valueSchema, result);
+              double responseSize = getResponseSize(response.getKey(), response.getValue());
+              double latency =
+                  Duration.between(requestInstant, result.getCompletionTimestamp()).toMillis();
```
dimitarndimitrov, line 159: Isn't the actual latency result here a `long`? I think we only deceive ourselves that we have better precision than what we actually have. I'd suggest using `long` or even `int`, and if some metrics API requires `double`, let's convert at the call-site.

P.S. This change would then propagate all the way down.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,426 @@
...
+          1000,
+          250,
+          Percentiles.BucketSizing.CONSTANT,
+          percentiles.entrySet().stream()
+              .map(
+                  namePercentile ->
+                      new Percentile(namePercentile.getKey(), namePercentile.getValue()))
+              .toArray(Percentile[]::new));
+    }
+  }
```
dimitarndimitrov, line 392: This means that the max latency we can track is just `250`, right? That seems way too low to me. I think we also track latency as an integer number, which means that we actually have almost no precision loss for the percentile histogram. I'd suggest bumping the `sizeInBytes` a bit (as this would be a global metric) and then bumping significantly the max latency that we can track.

kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
```diff
@@ -130,11 +130,17 @@ public void dispose(Producer<byte[], byte[]> producer) {
...
+    ProducerMetricsFactory(Provider<KafkaRestContext> context) {
+      this.context = requireNonNull(context);
+    }
+
     @Override
     public ProducerMetrics provide() {
-      producerMetrics = new ProducerMetrics(Time.SYSTEM);
+      producerMetrics = new ProducerMetrics(context.get().getConfig(), Time.SYSTEM);
       return producerMetrics;
     }
```
dimitarndimitrov: Did you find a nicer approach than what we discussed offline about initializing `producerMetrics` and ensuring that's done only once, and done in a thread-safe manner?
If not I'd suggest using a DCL-like structure that can ensure we are not reinitializing the metrics multiple times.
In its simplest form (when relying on the singleton factory-singleton provided object) this will just check that `producerMetrics` is always `null` when we try to initialize it.

dimitarndimitrov (commented):
Thanks for addressing a bunch of these super-fast! I got a second, smaller batch of comments.

kafka-rest/src/main/java/io/confluent/kafkarest/backends/kafka/KafkaModule.java
dimitarndimitrov: Hm, I think I got an outdated diff in the mix, sorry about that. Let me check all my other comments too.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -23,40 +25,83 @@
...
-    this.metrics = new MetricsBuilder(JMX_PREFIX).withTime(time).build();
+  public ProducerMetrics(KafkaRestConfig config, Time time) {
+    this.metrics = new MetricsBuilder(config.getMetricsContext()).withTime(time).build();
+    this.jmxPrefix = config.getString(KafkaRestConfig.METRICS_JMX_PREFIX_CONFIG);
+    this.fullyQualifiedRecordErrorSensor =
+        String.join(
+            ":",
+            jmxPrefix,
+            ProducerMetricsRegistry.GROUP_NAME,
+            ProducerMetricsRegistry.RECORD_ERROR_SENSOR);
```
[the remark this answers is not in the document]
thomaskwscott: I think this is more efficient as we only compute these strings once on ProducerMetrics creation. Given the overhead for creating templates this is more efficient than String.format too.

For readability I think the separate convenience methods for recording metrics are more readable than a templating solution. If we had a lot of them it would become unmanageable but given there are only 5 i think we are ok.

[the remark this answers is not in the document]
dimitarndimitrov: I think we might be disagreeing because we are talking about different things :)

My suggestion is not to remove the actual precomputation, as that was the main optimization from the previous version - just to replace the `String.join` calls here with either a `String.format` with a single parameter, or even directly concatenating the common prefix/template with the actual sensor name. There is no meaningful overhead in creating what I call template above - it is a single `String` which won't escape YoungGen as it is needed solely in the constructor.

The separate convenience methods won't be affected at all.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -73,93 +118,88 @@ public ProduceMetricMBean mbean(String groupName, Map<String, String> tags) {
...
+      this.tags = Collections.unmodifiableMap(tags);
     }
 
     @Override
     public boolean equals(Object o) {
       if (this == o) {
         return true;
       }
-      if (o == null || getClass() != o.getClass()) {
+      if (o == null || !(o instanceof BeanCoordinate)) {
```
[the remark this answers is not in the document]
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -73,93 +118,88 @@ public ProduceMetricMBean mbean(String groupName, Map<String, String> tags) {
...
-    }
-
-    public String getBeanName() {
-      return beanName;
-    }
-
-    public Map<String, String> getTags() {
-      return tags;
+      this.beanName = Objects.requireNonNull(beanName);
+      this.tags = Collections.unmodifiableMap(tags);
```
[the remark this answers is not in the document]
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -187,31 +227,40 @@ public int hashCode() {
...
+
+    public void recordRequest() {
+      recordMetric(fullyQualifiedRequestSensor, 1.0);
+    }
+
+    public void recordRequestSize(double value) {
+      recordMetric(fullyQualifiedRequestSizeSensor, value);
+    }
+
+    public void recordMetric(String sensorName, double value) {
```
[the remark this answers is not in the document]
thomaskwscott: nope, fixed

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
thomaskwscott: I'd pulled the values from fetch metrics in ce-kafka but looks like i was looking in the wrong place. Have updated with sensible 30s max now.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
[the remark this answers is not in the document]
thomaskwscott: ah, i'm with you and have changed.

kafka-rest/src/main/java/io/confluent/kafkarest/ProduceRequestSerdeModule.java
```diff
@@ -0,0 +1,88 @@
...
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.deser.BeanDeserializerModifier;
+import com.fasterxml.jackson.databind.deser.ResolvableDeserializer;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.datatype.guava.PackageVersion;
+import io.confluent.kafkarest.entities.v3.ProduceRequest;
+import java.io.IOException;
+
+public class ProduceRequestSerdeModule extends SimpleModule {
```
dimitarndimitrov, line 32: I assume this is needed for the correct recording of the original produce message size?

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,390 @@
...
+        if (metricName.group().equals(beanCoordinate.beanName)
+            && beanCoordinate.tags.equals(metricName.tags())) {
+          metrics.removeMetric(metricName);
+        }
+      }
+    }
+  }
+
+  // I'm not going to lie this exists only to defeat checkstyle ;-)
+  public static class SensorBuilder {
```
dimitarndimitrov, line 264: nit: This can be `private`, I think?

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,390 @@
...
+    private MetricName getMetricName(String name, String doc) {
+      MetricName metricName =
+          bean.metricName(
+              new MetricNameTemplate(
+                  name, ProducerMetricsRegistry.GROUP_NAME, doc, Collections.emptySet()));
+      return metricName;
+    }
+  }
+
+  public static class MeasuredStatSupplier {
```
dimitarndimitrov, line 327: nit: This can be `private`, I think?

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
```diff
@@ -0,0 +1,390 @@
...
+          Percentiles.BucketSizing.CONSTANT,
+          percentiles.entrySet().stream()
+              .map(
+                  namePercentile ->
+                      new Percentile(namePercentile.getKey(), namePercentile.getValue()))
+              .toArray(Percentile[]::new));
+    }
+  }
+
+  public static class MetricsBuilder {
```
dimitarndimitrov, line 358: Do we need this class after the latest changes? Either I'm missing something, or it can be replaced by a simple `private` method now.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+  public static final String REQUEST_SIZE_AVG_DOC = "The average request size in bytes.";
+
+  public static final String REQUEST_COUNT_WINDOWED = "request-count-windowed";
+  public static final String REQUEST_COUNT_WINDOWED_DOC =
+      "The total number of requests sent within the given window.";
+
+  // response
+  public static final String RESPONSE_SEND_RATE = "response-rate";
+  public static final String RESPONSE_SEND_RATE_DOC =
+      "The average number of response sent per second.";
```
dimitarndimitrov, line 43: nit: response -> responses

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+  public static final String RECORD_ERROR_RATE = "record-error-rate";
+  public static final String RECORD_ERROR_RATE_DOC =
+      "The average per-second number of record sends that resulted in errors.";
+
+  // latency
+  public static final String REQUEST_LATENCY_AVG = "request-latency-avg";
+  public static final String REQUEST_LATENCY_AVG_DOC = "The average request latency";
+
+  public static final String REQUEST_LATENCY_MAX = "request-latency-max";
+  public static final String REQUEST_LATENCY_MAX_DOC = "The max request latency";
```
dimitarndimitrov, line 63: That's also a problem for my "favorite" rate metrics, but here too the mention of average, max, and percentiles brings up the question of the aggregation window. Am I correct in thinking that latency is tracked as Kafka Metrics `SampledStat` and therefore it uses some window that's not the whole lifetime of the metrics sensor? Can you please document that here?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -128,10 +140,25 @@ public void produce(
...
+            })
+        .thenApply(
+            result -> {
+              ProduceResponse response =
+                  toProduceResponse(
+                      clusterId, topicName, keyFormat, keySchema, valueFormat, valueSchema, result);
+              long latency =
+                  Duration.between(requestInstant, result.getCompletionTimestamp()).toMillis();
+              recordResponseMetrics(latency);
+              return response;
```
dimitarndimitrov, line 160: nit: What do you think about also recording latency for error requests? It feels like Coordinated Omission otherwise.

kafka-rest/src/test/java/io/confluent/kafkarest/ProduceRequestSerdeModuleTest.java
```diff
@@ -0,0 +1,75 @@
+package io.confluent.kafkarest;
```
dimitarndimitrov, line 1: nit: Missing License header.

kafka-rest/src/test/java/io/confluent/kafkarest/ProducerMetricsTest.java
```diff
@@ -0,0 +1,193 @@
+package io.confluent.kafkarest;
```
dimitarndimitrov, line 1: nit: Missing License header.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/ProduceActionIntegrationTest.java
```diff
@@ -64,15 +64,14 @@
...
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
 // TODO ddimitrov This continues being way too flaky.
 //  Until we fix it (KREST-1542), we should ignore it, as it might be hiding even worse errors.
-@Ignore
+// @Ignore
```
dimitarndimitrov, line 74: I have mixed feelings about this - I totally want to enable it, but I think until we have figured out how to avoid SR calling `System.exit(1)` on us from time to time, it will be a pretty annoying problem for our CI results.

dimitarndimitrov (commented):
The changes look pretty good to me overall, some comments below.

kafka-rest/src/main/java/io/confluent/kafkarest/ProduceRequestSerdeModule.java
thomaskwscott: yep, it injects a customer deserializer that wraps the regular deserializer and records the position in the input stream pre and post deserialization. It uses these to determine input request size.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
thomaskwscott: iirc it blows up checkstyle with too many instantiations.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/ProduceActionIntegrationTest.java
thomaskwscott: I didn't mean to re-enable this, was just using the test locally to test the deserialization changes.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
thomaskwscott: I've added a comment to this effect, the default window size is 30s

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
thomaskwscott: fixed

kafka-rest/src/test/java/io/confluent/kafkarest/ProduceRequestSerdeModuleTest.java
thomaskwscott: fixed

kafka-rest/src/test/java/io/confluent/kafkarest/ProducerMetricsTest.java
thomaskwscott: fixed

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetrics.java
dimitarndimitrov: That really seems odd to me. I would prefer not having a separate class for satisfying what seems to be a weird rule, but I don't think this should hold up the changes.

kafka-rest/src/main/java/io/confluent/kafkarest/ProducerMetricsRegistry.java
dimitarndimitrov: I mostly meant the `_DOC` constants, as that's what brought the question of the aggregation window to my attention. If these are actually exposed to end users, we should probably include the clarification there; if not, your comment seems perfectly fine to me.

dimitarndimitrov (approved):
+1.

Thanks for the super-fast (as always) review follow-ups!

PR #893, opened 2021-09-07
Remove ZooKeeperClient/KafkaZkClient usage from tests

ijuma:
These are internal classes and hence can change in incompatible ways.
One such incompatible change committed to Apache Kafka 3.0 is the reason
why we need this in 7.0.x too.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/AuthorizationErrorTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/AuthorizationErrorTest.java
@@ -72,9 +72,7 @@ public class AuthorizationErrorTest
   @Before
   public void setUp() throws Exception {
     super.setUp();
-    kafka.utils.TestUtils.createTopic(zkClient, TOPIC_NAME, 1, 1,
-        JavaConverters.asScalaBuffer(this.servers),
-        new Properties());
+    createTopic(TOPIC_NAME, 1, (short) 1);
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


ijuma:
One test failed for build 2:

> io.confluent.kafkarest.v2.KafkaConsumerManagerTest.testBackoffMsControlsPollCalls
https://jenkins.confluent.io/job/Confluentinc%20Contributors/job/kafka-rest/job/PR-893/2/

This looks to be flaky, it fails occasionally without this change. And it passed locally after a few retries.

rondagostino (approved):
LGTM assuming builds pass.

dimitarndimitrov:
This seems to have caused all integration tests based on `ClusterTestHarness` to fail for `7.0.x` and `master`. As a unit test, `KafkaConsumerManagerTest` failing seems to have hidden all of these by skipping their phase. I'll provide an update when I have a fix available, hopefully soon.

ijuma:
Sorry for that. Can we fix the build not to ignore integration tests in such cases?

dimitarndimitrov:
> Sorry for that. Can we fix the build not to ignore integration tests in such cases?

I don't think you (or anyone) could have done anything better in this specific situation.

As for not ignoring other phases, we should very much look into that.
I am currently experimenting with the following: (1) comment all but one unit tests and all but one integration tests; (2) reproduce the skipping locally by failing the unit test; (3) test whether running with `--fail-at-end` also runs the integration test;

PR #894, opened 2021-09-08
Fix ambiguous describeConfigs reference in tests

bob-barrett:
A recent change to Apache Kafka (https://github.com/apache/kafka/commit/1d22b0d70686aef5689b775ea2ea7610a37f3e8c) added an overload to `Admin.describeTopics()`, which causes tests in `TopicManagerImplTest` to have ambiguous argument matchers in their mocks. This PR fixes that by specifically matching a `Collection` argument.


rigelbm (approved):
LGTM

dimitarndimitrov (approved):
+1.

Looks good to me, but keep in mind that it looks like https://github.com/confluentinc/kafka-rest/pull/893 has caused tons and tons of failures on `7.0.x`, so you might not get the result you want until we fix that too.

PR #896, opened 2021-09-08
KREST-1965 Ignore some known very flaky tests

dimitarndimitrov:
The two tests ignored here are way too flaky and we have seen at least one occasion were one of them was hiding an even worse problem: https://github.com/confluentinc/kafka-rest/pull/893.

We have added tasks for actually fixing them, but in the meantime they should not be causing as much trouble as they currently are.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
@@ -401,6 +401,9 @@ public class KafkaConsumerManagerTest {
         consumerManager.deleteConsumer(groupName, consumer.cid());
     }
 
+  // TODO ddimitrov This continues being way too flaky, even after some fix attempts.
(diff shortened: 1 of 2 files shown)


ehumber (approved):

dimitarndimitrov:
@ehumber Thank you very much for the super-quick review!
With #894 and #895 merged, I'm now very interested to see what will be the results from the Jenkins run.

kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -401,6 +401,9 @@ public void onCompletion(List<TopicPartitionOffset> offsets, Exception e) {
         consumerManager.deleteConsumer(groupName, consumer.cid());
     }
 
+  // TODO ddimitrov This continues being way too flaky, even after some fix attempts.
+  //  Until we fix it (KREST-2300), we should ignore it, as it might be hiding even worse errors.
+  @Ignore
```
ijuma, line 406: This one is definitely too flaky.

PR #897, opened 2021-09-09
KREST-2260: Use super user when creating topic in AuthorizationErrorTest.

rigelbm:
Prior to https://github.com/confluentinc/kafka-rest/pull/893, this topic was being created directly in Zookeeper, bypassing Kafka authorization entirely.

Now that we are using Kafka to create the topic, we need to pass through authorization. This changes uses the super-user to create the topic, so that we don't fail authorization.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/AuthorizationErrorTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/AuthorizationErrorTest.java
@@ -13,7 +13,6 @@
  * specific language governing permissions and limitations under the License.
  */
 
-
 package io.confluent.kafkarest.integration;
 
 import static io.confluent.kafkarest.TestUtils.assertErrorResponse;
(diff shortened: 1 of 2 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/integration/AuthorizationErrorTest.java
```diff
@@ -72,7 +71,14 @@
...
   public void setUp() throws Exception {
     super.setUp();
-    createTopic(TOPIC_NAME, 1, (short) 1);
+    Properties properties = restConfig.getAdminProperties();
+    properties.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, brokerList);
+    properties.put(
+        "sasl.jaas.config",
+        "org.apache.kafka.common.security.plain.PlainLoginModule required "
+            + " username=\"admin\""
+            + " password=\"admin-secret\";");
```
ijuma, line 80: Can we extract this to a constant? It's used in other places in this file too.

ijuma: There's also the `overrideKafkaRestConfigs`

rigelbm: Done.

ehumber (approved):

PR #898, opened 2021-09-09
Pin google-java-format version to 1.11.0.

rigelbm:
Different versions of `google-java-format` will format the code in slightly different ways. This pin is go guarantee we always use the same version everywhere.


dimitarndimitrov (approved):
+1.

Maybe just add a comment why are we doing this.

PR #910, opened 2021-09-20
Log setup/teardown of integration tests

dimitarndimitrov:
Some of kafka-rest's integration tests seem to take more than 80% of their running time just for setup + teardown. In addition to that, during the teardown, some errors that would usually indicate a serious failure can be seen.

This change adds some simple logging that will help identifying the rough ratios of setup to actual test to teardown. It will also help us identify if some usually unexpected error is happening during teardown (which might actually make the error expected).


dimitarndimitrov:
@rigelbm JFYI, I actually wanted this to land on `6.2.x`, I hope you don't mind (if CI approves).

PR #912, opened 2021-09-22
KREST-1976 add producer rate limiting

ehumber:
Rate limiting based on config properties in rest

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ProduceControllerImpl.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ProduceControllerImpl.java
@@ -17,10 +17,17 @@ package io.confluent.kafkarest.controllers;
 
 import static java.util.Objects.requireNonNull;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Multimap;
 import com.google.protobuf.ByteString;
+import io.confluent.kafkarest.KafkaRestConfig;
+import io.confluent.kafkarest.SystemTime;
+import io.confluent.kafkarest.Time;
 import io.confluent.kafkarest.entities.ProduceResult;
+import io.confluent.kafkarest.exceptions.RateLimitGracePeriodExceededException;
+import io.confluent.kafkarest.exceptions.TooManyRequestsException;
 import java.time.Instant;
+import java.util.LinkedList;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
@@ -32,10 +39,28 @@ import org.apache.kafka.common.header.internals.RecordHeader;
 final class ProduceControllerImpl implements ProduceController {
 
   private final Producer<byte[], byte[]> producer;
+  private final KafkaRestConfig config;
+  private final LinkedList<Long> rateCounter = new LinkedList<>();
+  private Optional<Long> gracePeriodStart = Optional.empty();
+  private static final int ONE_SECOND = 1000;
+
+  @VisibleForTesting public Time time = new SystemTime();
 
   @Inject
-  ProduceControllerImpl(Producer<byte[], byte[]> producer) {
+  ProduceControllerImpl(Producer<byte[], byte[]> producer, KafkaRestConfig config) {
     this.producer = requireNonNull(producer);
+    this.config = config;
+  }
+
+  private void addToAndCullRateCounter(long now) {
+
+    rateCounter.add(now);
+    while (rateCounter.peek() < now - ONE_SECOND) {
+      rateCounter.pop();
+    }
+    if (rateCounter.size() <= config.getInt(KafkaRestConfig.PRODUCE_MAX_REQUESTS_PER_SECOND)) {
+      gracePeriodStart = Optional.empty();
+    }
   }
 
   @Override
@@ -47,7 +72,26 @@ final class ProduceControllerImpl implements ProduceController {
       Optional<ByteString> key,
       Optional<ByteString> value,
       Instant timestamp) {
+
+    long now = time.milliseconds();
+    addToAndCullRateCounter(now);
+
     CompletableFuture<ProduceResult> result = new CompletableFuture<>();
+    if (rateCounter.size() > config.getInt(KafkaRestConfig.PRODUCE_MAX_REQUESTS_PER_SECOND)) {
+
+      if (!gracePeriodStart.isPresent()
+          && config.getInt(KafkaRestConfig.PRODUCE_GRACE_PERIOD) != 0) {
+        gracePeriodStart = Optional.of(now);
+      } else if (config.getInt(KafkaRestConfig.PRODUCE_GRACE_PERIOD) == 0
+          || config.getInt(KafkaRestConfig.PRODUCE_GRACE_PERIOD) < now - gracePeriodStart.get()) {
+        result.completeExceptionally(new RateLimitGracePeriodExceededException());
+        return result;
+      }
+
+    } else {
+      gracePeriodStart = Optional.empty();
+    }
+
     producer.send(
         new ProducerRecord<>(
             topicName,
@@ -65,6 +109,11 @@ final class ProduceControllerImpl implements ProduceController {
         (metadata, exception) -> {
           if (exception != null) {
             result.completeExceptionally(exception);
+          } else if (gracePeriodStart.isPresent()) {
+            result.completeExceptionally(
+                new TooManyRequestsException(
+                    config.getInt(KafkaRestConfig.PRODUCE_MAX_REQUESTS_PER_SECOND),
+                    config.getInt(KafkaRestConfig.PRODUCE_GRACE_PERIOD)));
           } else {
             result.complete(ProduceResult.fromRecordMetadata(metadata));
           }
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
@@ -24,10 +24,13 @@ import com.fasterxml.jackson.databind.MappingIterator;
 import com.fasterxml.jackson.jaxrs.base.JsonMappingExceptionMapper;
 import com.fasterxml.jackson.jaxrs.base.JsonParseExceptionMapper;
 import com.google.auto.value.AutoValue;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import io.confluent.kafkarest.common.CompletableFutures;
+import io.confluent.kafkarest.exceptions.RateLimitGracePeriodExceededException;
 import io.confluent.kafkarest.exceptions.StatusCodeException;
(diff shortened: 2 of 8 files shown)


thomaskwscott:
@ehumber Thanks for the PR, looks great. 

I have a couple of questions:

1. Does this impose a rate limit per connection or across the entire proxy instance?
2. This PR does the configured hard rate limit, are we going to account for producer throttling too?

ehumber:
> @ehumber Thanks for the PR, looks great.
> 
> I have a couple of questions:
> 
> 1. Does this impose a rate limit per connection or across the entire proxy instance?
> 2. This PR does the configured hard rate limit, are we going to account for producer throttling too?

@thomaskwscott 

1- It does now!  Thanks for spotting this, I'd done the initial change per connection
2- I'll do this as a follow on PR as the two are independent changes.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -59,6 +62,7 @@
 public abstract class StreamingResponse<T> {
 
   private static final Logger log = LoggerFactory.getLogger(StreamingResponse.class);
+  @VisibleForTesting static ChunkedOutputFactory factory;
```
rigelbm, line 65: We can:
1. Create a (default-visibility) `StreamingResponseFactory` that takes a `ChunkedOutputFactory` in the constructor, and has the `from(MappingIterator)` method.
2. Add a `ChunkedOutputFactory` to the StreamingResponse constructors. Pass it down from the `StreamingResponseFactory`.
3. Have our static `from(MappingIterator)` use a static final `StreamingResponseFactory`.
4. Enjoy our life free of mutable static variables. :)

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -218,7 +251,8 @@ private void close() {
     }
   }
 
-  private abstract static class ResultOrError {
+  @VisibleForTesting
```
rigelbm, line 254: I wouldn't bother adding `@VisibleForTesting` in default visibility members.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ProduceControllerImpl.java
```diff
@@ -32,10 +40,31 @@
 final class ProduceControllerImpl implements ProduceController {
 
   private final Producer<byte[], byte[]> producer;
+  private final KafkaRestConfig config;
```
rigelbm, line 43: I think we should rate-limit in the resource layer, not in the controller layer. :|

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponseFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.response;
+
+import com.fasterxml.jackson.databind.MappingIterator;
+
+public class StreamingResponseFactory {
```
rigelbm, line 20: Optional: I would make this class an inner class of `StreamingResponse`.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -83,15 +83,20 @@
               response -> ((ErrorMessage) response.getEntity()).getMessage())
           .build();
 
-  StreamingResponse() {}
+  private ChunkedOutputFactory chunkedOutputFactory;
```
rigelbm, line 86: I like the changes in `StreamingResponse` very much. Thank you :)

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -92,9 +128,46 @@ public void produce(
       @PathParam("topicName") String topicName,
       MappingIterator<ProduceRequest> requests)
       throws Exception {
-    ProduceController controller = produceController.get();
-    StreamingResponse.from(requests)
-        .compose(request -> produce(clusterId, topicName, request, controller))
+
+    if (rateLimit()) {
```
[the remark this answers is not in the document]
ehumber: For your first point:
If the rest-producer rate limit ends up being lower than the overall limit then we want this check in place too don't we?  We shouldn't just rely on the overall limit for produces in case we need to set the limit really low.  Especially as it is possible to call produce in a non-streaming way.

For your second point:
The intent of having this block of code was to fail fast, before we even make a produceController, in the case where the produce endpoint is getting hit with lots of individual requests. It doesn't matter if there are multiple messages in the requests MappingIterator (ie in the case where it is a streaming produce, rather than individual calls) - that flow is picked up by the second block of rate limiting code.   The intent of this code is to block too many individual produce calls. That means I don't think we want this check hanging off the .from(requests) part here, it should be at the top level.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -67,18 +75,46 @@
...
-  private final Provider<RecordSerializer> recordSerializer;
-  private final Provider<ProduceController> produceController;
+  private final Provider<SchemaManager> schemaManagerProvider;
+  private final Provider<RecordSerializer> recordSerializerProvider;
+  private final Provider<ProduceController> produceControllerProvider;
+  private final ChunkedOutputFactory chunkedOutputFactory;
+  private final StreamingResponseFactory streamingResponseFactory;
+
+  static final ConcurrentLinkedQueue<Long> rateCounter = new ConcurrentLinkedQueue<>();
+  static Optional<AtomicLong> gracePeriodStart = Optional.empty();
```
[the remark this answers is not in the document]
ehumber: That makes sense, I'll go update it.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/ProduceResponse.java
```diff
@@ -54,6 +54,10 @@
   @JsonInclude(Include.NON_ABSENT)
   public abstract Optional<ProduceResponseData> getValue();
 
+  @JsonProperty("resume_after_ms")
+  @JsonInclude(Include.NON_ABSENT)
+  public abstract Optional<Long> getResumeAfterMs();
```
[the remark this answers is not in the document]
ehumber: I picked a long as it was smaller to store than an Instant (because this map could end up holding a lot of values if the rate limit is high so I wanted to keep it as small as possible), and it's a number, so I don't have to do String -> number conversion later in the code as that will slow down the flow.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponseFactory.java
ehumber: I thought Streaming response had enough inner classes, so made it separate :) No strong feelings though

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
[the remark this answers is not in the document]
ehumber: I think we still need to worry about concurrency, but I do like the singleton object so I've added that in.

Even if it is a singleton object, we can for example have one thread reading from the rateCounter list, and one adding to it - the underlying object needs to be set up to not throw a concurrentModificationException.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -208,6 +302,53 @@ private static ProduceResponse toProduceResponse(
...
+      String clusterId, String topicName) {
+
+    CompletableFuture<ProduceResult> result = new CompletableFuture<>();
+
+    result.completeExceptionally(
+        new RateLimitGracePeriodExceededException(
+            config.getInt(KafkaRestConfig.PRODUCE_MAX_REQUESTS_PER_SECOND),
+            config.getInt(KafkaRestConfig.PRODUCE_GRACE_PERIOD)));
+
+    return result.thenApply(
```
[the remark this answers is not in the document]
ehumber: Yes, you're entirely right (the code was needed to make the types match, even though it was never used).  I've removed the method and now the code just passes the exception on to the .resume call inline.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponseFactory.java
rigelbm: Ack

kafka-rest/src/main/java/io/confluent/kafkarest/resources/ProduceRateLimitCounters.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  public static void setGracePeriodStart(final Optional<AtomicLong> gracePeriodStart) {
+    ProduceRateLimitCounters.gracePeriodStart = gracePeriodStart;
+  }
+
+  public static boolean gracePeriodPresent() {
+    return gracePeriodStart.isPresent();
+  }
+}
```
rigelbm, line 63: Why so many `static`? :fearful: There should be no static here. You need to bind `ProduceRateLimitCounters` as itself in the V3 resources Module, then inject it in the Action.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/ProduceRateLimitCounters.java
```diff
@@ -0,0 +1,63 @@
...
+  public static void add(Long time) {
+    rateCounter.add(time);
+  }
+
+  public static void resetGracePeriodStart() {
+    gracePeriodStart = Optional.empty();
+  }
+
+  // This throws an exception if the gracePeriodStart is Optional.empty()
+  public static Long get() {
```
rigelbm, line 52: `Optional<Long> getGracePeriodStart()`

kafka-rest/src/main/java/io/confluent/kafkarest/resources/ProduceRateLimitCounters.java
```diff
@@ -0,0 +1,63 @@
...
+    return gracePeriodStart.get().get();
+  }
+
+  public static void setGracePeriodStart(final Optional<AtomicLong> gracePeriodStart) {
+    ProduceRateLimitCounters.gracePeriodStart = gracePeriodStart;
+  }
+
+  public static boolean gracePeriodPresent() {
+    return gracePeriodStart.isPresent();
+  }
```
rigelbm, line 62: Remove this method, since you can return an Optional above.

PR #915, opened 2021-10-14
KREST-2837 move produce resource cleanup to end of streaming response.

thomaskwscott:
As noticed in previous testing we are seeing producers closed before all messages are written to them. This is because the context shutdown() call that closes the producer is made in a filter called when a response is created rather than completed. For streaming concerns the response is created (and resumed) a long time before it has completed writing. 

To keep things generic a new 'dispose' Runnable has been added to ComposingStreamingResponse, this is called in an overriden resume() method directly after the superclass resume (that closes the response channel) is called.

I this case the runnable calls a new method on the ProducerController for disposing of resources. 

This behaviour is covered in the existing integration tests.


thomaskwscott:
Closing as @rigelbm has a more elegant solution

PR #916, opened 2021-10-15
KREST-2835: Move ownership of Admin/Producer to HK2.

rigelbm:
Why: Currently, Admin/Producer are owned by the KafkaRestContext. Their lifetime is controlled by a mix of HK2 (creation) and ContainerResponseFilter (disposal). That doesn't work for streaming responses because JAX-RS ContainerResponseFilters and Interceptors are triggered immediately as soon as a response is provided, even if the request hasn't finished yet. That means we are closing, for example, the Producer, as soon as we provide a ChunkedOuput OutputStream as response, even though we haven't closed the ChunkedOuput yet. Pending `Producer#send` calls will fail, since the Producer is already closed.

What: This PR removes the KafkaRestContext ownership over Admin/Producer. It lets HK2 handle the ownership and lifetime of the instances. We can then avoid the previous problem by overriding the default bindings with bindings using the appropriate lifetime, e.g. RequestScoped will dispose only after the whole request has finished.

This PR also fixes two bugs:

  * In ProduceAction, if a Kafka exception has been thrown, there's no result available. Querying it for the finish time (for metrics purposes) causes NPE.
  * In StreamingResponse, we are not handling `RestException`.

The changes in `KafkaRestFixture` are to allow it to create a standalone Kafka REST server without a backing Kafka server. `KafkaModuleOverridingTest` contains an example of how the bindings can be overridden.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/unit/DefaultKafkaRestContextTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/unit/DefaultKafkaRestContextTest.java
@@ -33,21 +33,6 @@ public class DefaultKafkaRestContextTest {
   }
 
   @Test
-  public void testGetProducerThreadSafety() throws InterruptedException {
-    Set<Object> refs = new CopyOnWriteArraySet<>();
-
-    ExecutorService executor = Executors.newFixedThreadPool(100);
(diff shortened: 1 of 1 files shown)


rigelbm:
The failed downstream build is from `blueway`. Unrelated, thus ignoring.

PR #924, opened 2021-11-09
KREST-1936 Remove mask filtering of producer configs.

dimitarndimitrov:
This change removes the filtering of all properties not explicitly known to be valid for producer configs. This has demonstrated to be a problem for producer configs before (see [KREST-481](https://confluentinc.atlassian.net/browse/KREST-481)), and now causes a problem for other interceptor configs.

Note that such filtering doesn't work in the general case in the face of extensibility mechanisms that Kafka has (like interceptors) that don't provide a way to register their expected and allowed config in advance.

That same filtering has been left unchanged for the different SR-related configs, despite a history of causing problems there too (see [KREST-1187](https://confluentinc.atlassian.net/browse/KREST-1187)). Removing that same filtering for SR-related configs requires more refactoring so it won't be addressed in this change.

P.S. This is pending testing of the exact use-case described by @conan-goldsmith in [KREST-1936](https://confluentinc.atlassian.net/browse/KREST-1936).

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -44,7 +44,6 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.ws.rs.core.MediaType;
-import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
 import org.apache.kafka.common.config.ConfigDef.Range;
@@ -54,7 +53,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Settings for the REST proxy server.
+ * Settings for the Kafka REST server.
  */
 public class KafkaRestConfig extends RestConfig {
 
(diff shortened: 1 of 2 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -925,9 +912,13 @@ public RestMetricsContext getMetricsContext() {
   }
 
   private final class ConfigsBuilder {
-    private final Set<String> mask ;
+    private final Set<String> mask;
     private final Map<String, ConfigValue> configs = new HashMap<>();
 
+    private ConfigsBuilder() {
+      this.mask = null;
```
rigelbm, line 919: nit: Can we use empty set? I'm nullophobic.

rigelbm (approved):
LGTM

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
dimitarndimitrov: You got it.

ehumber (approved):

dimitarndimitrov:
The sole failure in the last PR CI runs is the notorious `KafkaConsumerManagerTest.testBackoffMsControlsPollCalls` which has been flaky for a long time now and should be unrelated to the changes here.

PR #929, opened 2021-11-16
KREST-3136 process the produce responses in their own thread

ehumber:
We suspect there is conention with the Kafka network thread when using the rest producer.  This commit ensures that produce responses are processed on their own thread.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -195,6 +195,15 @@ public class KafkaRestConfig extends RestConfig {
           + "is 1 hour.";
   public static final String PRODUCE_RATE_LIMIT_CACHE_EXPIRY_MS_DEFAULT = "3600000";
 
+  public static final String PRODUCE_RESPONSE_THREAD_POOL_SIZE =
+      "api.v3.produce.response.thread.pool.size";
+  private static final String PRODUCE_RESPONSE_THREAD_POOL_SIZE_DOC =
+      "Number of threads in the executor thread pool used to process produce responses.";
+  public static final String PRODUCE_RESPONSE_THREAD_POOL_SIZE_DEFAULT =
+      Integer.toString(Runtime.getRuntime().availableProcessors());
+  public static final ConfigDef.Range PRODUCE_RESPONSE_THREAD_POOL_SIZE_VALIDATOR =
+      ConfigDef.Range.between(1, Integer.MAX_VALUE);
+
   public static final String CONSUMER_ITERATOR_TIMEOUT_MS_CONFIG = "consumer.iterator.timeout.ms";
   private static final String CONSUMER_ITERATOR_TIMEOUT_MS_DOC =
       "Timeout for blocking consumer iterator operations. This should be set to a small enough "
@@ -520,6 +529,13 @@ public class KafkaRestConfig extends RestConfig {
             PRODUCE_RATE_LIMIT_CACHE_EXPIRY_MS_DEFAULT,
             Importance.LOW,
             PRODUCE_RATE_LIMIT_CACHE_EXPIRY_MS_DOC)
+        .define(
+            PRODUCE_RESPONSE_THREAD_POOL_SIZE,
+            Type.INT,
+            PRODUCE_RESPONSE_THREAD_POOL_SIZE_DEFAULT,
+            PRODUCE_RESPONSE_THREAD_POOL_SIZE_VALIDATOR,
+            Importance.LOW,
+            PRODUCE_RESPONSE_THREAD_POOL_SIZE_DOC)
         .define(
             CONSUMER_ITERATOR_TIMEOUT_MS_CONFIG,
             Type.INT,
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
@@ -37,6 +37,7 @@ import io.confluent.kafkarest.entities.v3.ProduceResponse.ProduceResponseData;
 import io.confluent.kafkarest.exceptions.BadRequestException;
 import io.confluent.kafkarest.extension.ResourceAccesslistFeature.ResourceName;
 import io.confluent.kafkarest.ratelimit.DoNotRateLimit;
+import io.confluent.kafkarest.resources.v3.V3ResourcesModule.ProduceResponseThreadPool;
 import io.confluent.kafkarest.response.StreamingResponseFactory;
 import io.confluent.rest.annotations.PerformanceMetric;
 import java.time.Duration;
@@ -45,6 +46,7 @@ import java.util.Collections;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutorService;
 import java.util.function.Function;
 import java.util.stream.Collector;
 import javax.inject.Inject;
@@ -84,6 +86,7 @@ public final class ProduceAction {
   private final Provider<ProducerMetrics> producerMetrics;
   private final StreamingResponseFactory streamingResponseFactory;
   private final ProduceRateLimiters produceRateLimiters;
+  private final ExecutorService executorService;
 
   @Inject
   public ProduceAction(
@@ -92,13 +95,15 @@ public final class ProduceAction {
       Provider<ProduceController> produceControllerProvider,
       Provider<ProducerMetrics> producerMetrics,
       StreamingResponseFactory streamingResponseFactory,
-      ProduceRateLimiters produceRateLimiters) {
+      ProduceRateLimiters produceRateLimiters,
+      @ProduceResponseThreadPool ExecutorService executorService) {
     this.schemaManagerProvider = requireNonNull(schemaManagerProvider);
     this.recordSerializerProvider = requireNonNull(recordSerializer);
     this.produceControllerProvider = requireNonNull(produceControllerProvider);
     this.producerMetrics = requireNonNull(producerMetrics);
     this.streamingResponseFactory = requireNonNull(streamingResponseFactory);
     this.produceRateLimiters = requireNonNull(produceRateLimiters);
+    this.executorService = requireNonNull(executorService);
   }
 
   @POST
@@ -172,7 +177,7 @@ public final class ProduceAction {
             request.getTimestamp().orElse(Instant.now()));
 
     return produceResult
-        .handle(
+        .handleAsync(
             (result, error) -> {
               if (error != null) {
                 long latency = Duration.between(requestInstant, Instant.now()).toMillis();
@@ -180,8 +185,9 @@ public final class ProduceAction {
                 throw new CompletionException(error);
               }
               return result;
-            })
-        .thenApply(
+            },
+            executorService)
+        .thenApplyAsync(
             result -> {
               ProduceResponse response =
                   toProduceResponse(
@@ -197,7 +203,8 @@ public final class ProduceAction {
                   Duration.between(requestInstant, result.getCompletionTimestamp()).toMillis();
               recordResponseMetrics(latency);
               return response;
-            });
+            },
+            executorService);
   }
 
   private Optional<RegisteredSchema> getSchema(
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
@@ -15,10 +15,24 @@
 
 package io.confluent.kafkarest.resources.v3;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import io.confluent.kafkarest.config.ConfigModule.ProduceResponseThreadPoolSizeConfig;
 import io.confluent.kafkarest.response.ChunkedOutputFactory;
 import io.confluent.kafkarest.response.StreamingResponseFactory;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 import java.time.Clock;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import javax.inject.Inject;
+import javax.inject.Qualifier;
 import javax.inject.Singleton;
+import org.glassfish.hk2.api.AnnotationLiteral;
+import org.glassfish.hk2.api.Factory;
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
 
 public final class V3ResourcesModule extends AbstractBinder {
@@ -29,5 +43,48 @@ public final class V3ResourcesModule extends AbstractBinder {
     bindAsContract(ChunkedOutputFactory.class);
     bindAsContract(StreamingResponseFactory.class);
     bind(Clock.systemUTC()).to(Clock.class);
+    bindFactory(ProduceResponseExecutorServiceFactory.class)
+        .qualifiedBy(new ProduceResponseThreadPoolImpl())
+        .to(ExecutorService.class)
+        .in(Singleton.class);
+  }
+
+  @Qualifier
+  @Retention(RetentionPolicy.RUNTIME)
+  @Target({ElementType.TYPE, ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER})
+  public @interface ProduceResponseThreadPool {}
+
+  private static final class ProduceResponseThreadPoolImpl
+      extends AnnotationLiteral<ProduceResponseThreadPool> implements ProduceResponseThreadPool {}
+
+  private static final class ProduceResponseExecutorServiceFactory
+      implements Factory<ExecutorService> {
+
+    private final int produceResponseThreadPoolSize;
+
+    @Inject
+    ProduceResponseExecutorServiceFactory(
+        @ProduceResponseThreadPoolSizeConfig Integer produceExecutorThreadPoolSize) {
+      this.produceResponseThreadPoolSize = produceExecutorThreadPoolSize;
+    }
+
+    @Override
+    public ExecutorService provide() {
+      ThreadFactory namedThreadFactory =
+          new ThreadFactoryBuilder().setNameFormat("Produce-response-thread-%d").build();
+      return Executors.newFixedThreadPool(produceResponseThreadPoolSize, namedThreadFactory);
+    }
+
+    @Override
+    public void dispose(ExecutorService executorService) {
+      executorService.shutdown();
+      try {
+        if (!executorService.awaitTermination(60, TimeUnit.SECONDS)) {
+          executorService.shutdownNow();
+        }
+      } catch (InterruptedException e) {
+        executorService.shutdownNow();
+      }
+    }
   }
 }
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/config/ConfigModule.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/config/ConfigModule.java
@@ -123,6 +123,10 @@ public final class ConfigModule extends AbstractBinder {
         .qualifiedBy(new ProduceRateLimitEnabledConfigImpl())
         .to(Boolean.class);
 
+    bind(config.getInt(KafkaRestConfig.PRODUCE_RESPONSE_THREAD_POOL_SIZE))
+        .qualifiedBy(new ProduceResponseThreadPoolSizeImpl())
+        .to(Integer.class);
+
     bind(config.getProducerConfigs())
         .qualifiedBy(new ProducerConfigsImpl())
         .to(new TypeLiteral<Map<String, Object>>() {});
@@ -303,6 +307,15 @@ public final class ConfigModule extends AbstractBinder {
   private static final class ProduceRateLimitEnabledConfigImpl
       extends AnnotationLiteral<ProduceRateLimitEnabledConfig>
       implements ProduceRateLimitEnabledConfig {}
+
+  @Qualifier
+  @Retention(RetentionPolicy.RUNTIME)
+  @Target({ElementType.TYPE, ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER})
+  public @interface ProduceResponseThreadPoolSizeConfig {}
+
+  private static final class ProduceResponseThreadPoolSizeImpl
+      extends AnnotationLiteral<ProduceResponseThreadPoolSizeConfig>
+      implements ProduceResponseThreadPoolSizeConfig {}
 
   @Qualifier
   @Retention(RetentionPolicy.RUNTIME)
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProduceActionTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProduceActionTest.java
@@ -12,6 +12,7 @@ import static org.easymock.EasyMock.mock;
 import static org.easymock.EasyMock.replay;
 
 import com.fasterxml.jackson.databind.MappingIterator;
+import com.google.common.util.concurrent.MoreExecutors;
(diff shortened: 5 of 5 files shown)


ehumber:
@rigelbm I'd like your opinion on the injection of the ExecutorService.  I copied what was done for metrics, but I think the singleton pattern in the factory and .as(Singleton.class) aren't both necessary?

dimitarndimitrov (commented):
I think that we should start with just the common FJ pool instead of a fixed, dedicated thread pool just for Produce v3's responses, as the dedicated pool would be redundant whenever we don't use Produce v3. It's possible to limit the impact of that by reducing core threads/allowing core thread timeouts, but I feel it would be an unnecessary complication to our thread model at that stage.
This will also significantly simplify this change which I think is good - just changing that first `handle` to `handleAsync` without the executor param should suffice - once we've jumped, the `thenApply` should not even need to be made async, which would be slightly more efficient.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -30,5 +37,40 @@ protected void configure() {
     bindAsContract(ChunkedOutputFactory.class);
     bindAsContract(StreamingResponseFactory.class);
     bind(Clock.systemUTC()).to(Clock.class);
+    bindFactory(ExecutorServiceFactory.class, Singleton.class)
```
rigelbm, line 40: No need for `Singleton` here.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -30,5 +37,40 @@ protected void configure() {
...
+        .to(ExecutorService.class)
+        .in(Singleton.class);
```
rigelbm, line 42: Yes for `Singleton` here.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -30,5 +37,40 @@ protected void configure() {
...
+  }
+
+  private static final class ExecutorServiceFactory implements Factory<ExecutorService> {
+
+    private final int produceExecutorThreadPoolSize;
+    private volatile ExecutorService executorService;
```
rigelbm, line 48: Don't save the executor. HK2 will do it for you.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -30,5 +37,40 @@ protected void configure() {
...
+
+    @Inject
+    ExecutorServiceFactory(
+        @ProduceExecutorThreadPoolSizeConfig Integer produceExecutorThreadPoolSize) {
+      this.produceExecutorThreadPoolSize = requireNonNull(produceExecutorThreadPoolSize);
```
rigelbm, line 53: No need for `requireNonNull`. Unboxing will throw NPE already.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -30,5 +37,40 @@ protected void configure() {
...
+
+    @Override
+    public ExecutorService provide() {
+      ExecutorService localExecutorService = executorService;
+      if (localExecutorService == null) {
+        synchronized (this) {
+          localExecutorService = executorService;
+          if (localExecutorService == null) {
+            executorService =
+                localExecutorService = Executors.newFixedThreadPool(produceExecutorThreadPoolSize);
```
rigelbm, line 64: Just return this directly. HK2 will save the instance for you int the `Singleton` scope.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -166,16 +170,17 @@ public void produce(
...
                 long latency = Duration.between(requestInstant, Instant.now()).toMillis();
                 recordErrorMetrics(latency);
                 throw new CompletionException(error);
               }
               return result;
-            })
-        .thenApply(
+            },
+            executorService)
+        .thenApplyAsync(
```
rigelbm, line 183: No need to use a separate executor for the `apply`. Whoever closes the previous future (which is already a thread in `executorService` will execute the `apply` here.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -169,6 +169,14 @@
   public static final ConfigDef.Range PRODUCE_GRACE_PERIOD_MS_VALIDATOR =
       ConfigDef.Range.between(0, Integer.MAX_VALUE);
 
+  public static final String PRODUCE_EXECUTOR_THREAD_POOL_SIZE =
+      "api.v3.produce.executor.thread.pool.size";
+  private static final String PRODUCE_EXECUTOR_THREAD_POOL_SIZE_DOC =
+      "Number of threads in the executor thread pool used to process produce responses.";
+  public static final String PRODUCE_EXECUTOR_THREAD_POOL_SIZE_DEFAULT = "50";
+  public static final ConfigDef.Range PRODUCE_EXECUTOR_THREAD_POOL_SIZE_VALIDATOR =
+      ConfigDef.Range.between(1, Integer.MAX_VALUE);
```
rigelbm, line 178: Instead of being exclusive for produce, I would just call this non-blocking-thread-pool, or something like that. It is useful even for Admin (which we should also fix eventually). So, `non.blocking.thread.pool.size`, or something like that.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -94,14 +96,16 @@ public ProduceAction(
...
+      RateLimiter rateLimiter,
+      ExecutorService executorService) {
     this.schemaManagerProvider = requireNonNull(schemaManagerProvider);
     this.recordSerializerProvider = requireNonNull(recordSerializer);
     this.produceControllerProvider = requireNonNull(produceControllerProvider);
     this.producerMetrics = requireNonNull(producerMetrics);
     this.chunkedOutputFactory = requireNonNull(chunkedOutputFactory);
     this.streamingResponseFactory = requireNonNull(streamingResponseFactory);
     this.rateLimiter = requireNonNull(rateLimiter);
+    this.executorService = requireNonNull(executorService);
```
dimitarndimitrov, line 108: HK2 doesn't look good in our profiles, so I am starting to be much more skeptical to its usage. Unlike most of the other things we inject (like controllers or configs), this also is not really gaining much from being injectable, as it has much stronger locality.

If we have to configure the pool (although we can do without that - see my pool size comments below) we have to consume the injectable config, so I am not recommending to put this as a `static` field, but I will think about a better recommendation.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -30,5 +37,40 @@ protected void configure() {
     bindAsContract(ChunkedOutputFactory.class);
     bindAsContract(StreamingResponseFactory.class);
     bind(Clock.systemUTC()).to(Clock.class);
+    bindFactory(ExecutorServiceFactory.class, Singleton.class)
+        .to(ExecutorService.class)
+        .in(Singleton.class);
+  }
+
+  private static final class ExecutorServiceFactory implements Factory<ExecutorService> {
```
dimitarndimitrov, line 45: nit: It probably makes sense to use a new marker interface which is a subtype of `ExecutorService`. This is not really an `ExecutorSeervice` provider, it's a `ProduceResponseExecutor` provider.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -169,6 +169,14 @@
   public static final ConfigDef.Range PRODUCE_GRACE_PERIOD_MS_VALIDATOR =
       ConfigDef.Range.between(0, Integer.MAX_VALUE);
 
+  public static final String PRODUCE_EXECUTOR_THREAD_POOL_SIZE =
+      "api.v3.produce.executor.thread.pool.size";
```
dimitarndimitrov, line 173: nit: Maybe also include "response" in the name here? As this pool will be used mostly for blocking tasks, it won't be easy to also make use of it for other tasks in the Produce pipeline.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
dimitarndimitrov: What do you mean by non-blocking? This ends up blocking on `writev` syscalls - the way I understand it, this is a big part of why we are moving it away from the Kafka threads.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
rigelbm: +1, not for marker interface, but for annotation.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Yes. I forgot the very last future will be blocking. Let's keep it separate then just for produce. +1 for `produce.response` or `produce.callback` or something like that.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
ehumber: I ran one of the unit tests with the debugger on, with Executors.newFixedThreadPool(3) set as the executor service and saw that in some cases the same thread was used for both handle and thenApply, but not always:

handleAsync:
image


thenApply:
image


or using thenApplyAsycn I see pool-1-thread-1 for the handleAsync and then this for the thenApply

image

So, I think we need to explicitly specify the executor for both the handle and the thenApply.

The thread don't always match, but they are always from our thread pool

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
ehumber: The blocking doesn't affect whether we use this for admin calls or produce calls really though - we just want to not get tin the way of the kafka processing.

I've gone with `api.v3.rest.callback.thread.pool.size` but can easily change it if there are concerns

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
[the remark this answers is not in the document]
ehumber: See the results of the testing of which thread is used above - leaving thenApplyAsync and the executor service in place.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
ehumber: `api.v3.rest.callback.thread.pool.size` see my comment above.  Let me know if you're not keen on that !

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
ehumber: I'm not sure I agree that it's not an executorService provider - all it does is create and return an executor service with the right config.

@rigelbm can you explain what you mean by annotation.  Looking at the other factories that do the same type of thing in the codebase, they are written very similarly to what is here so I don't understand the request.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
dimitarndimitrov: The last picture seems very much expected - `thenApplyAsync` without the executor param will run on the common FJ pool.
Now on to the second picture, I am interested to see the lower part of the stack that reaches into the `thenApply` function. The `thenApply` applies to the result of `handle`, which is a new `CompletableFuture`. The contents of `handle` need to be executed (on the custom executor) in order to complete with the result that new `CompletableFuture`, and I am curious what reschedules on the `main` thread (which I assume is the one completing `controller.produce()`).

In order not to fret over one redundant context switch per response, we can keep the executor params on both right now, but I think we should:
1. collapse the `thenApply` into the `handle`.
2. examine the rescheduling on `main`.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
dimitarndimitrov: > I'm not sure I agree that it's not an executorService provider - all it does is create and return an executor service with the right config.

I think my main point is that we don't want to use this `ExecutorService` for generic purposes. It's used specifically for the blocking produce responses - e.g. it won't be a good idea to reuse the same instance for the AdminClient purposes that were mentioned above, but even if it wasn't the same instance, that "right" config will be right specifically for our case, not for other cases where we might need `ExecutorService`.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
ehumber: Yup I agree on the third picture, just included it for completeness.

I did think when I started this about collapsing the thenApply into the handle as I couldn't see anything async in the body of the handleAsync, and it just returns the result passed in anyway.

I decided against it as it always seems best to touch things as little as possible just before we are trying to release a product, but maybe that's the way forward.

handleAsync
image

thenApply (no executor)
image

dimitarndimitrov: > I decided against it as it always seems best to touch things as little as possible just before we are trying to release a product, but maybe that's the way forward.

Absolutely!

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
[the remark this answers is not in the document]
ehumber: I think the custom ThreadFactory is a nice to have, if we see problems in the field with the execution here then we can rename appropriately.  

https://github.com/adoptium/adoptium-support/issues/60#issuecomment-640030186 was interesting reading wrt now many processors are returned (if not directly relevant here).  

I've set the default for the value to be availabileProcessors() but left it configurable.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
ehumber: I'm going to leave this as an HK2 injected field - it's just one of many instances of this pattern.  If we determine it's something we want to alter we should come back to it and sort them all, not just this one.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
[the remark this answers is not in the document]
dimitarndimitrov: > I think the custom ThreadFactory is a nice to have, if we see problems in the field with the execution here then we can rename appropriately.

The problem is that we will be seeing a non-descript thread name in the logs / in the thread dumps that might belong to any of the thread pools that are created in the Kafka JVM without a dedicated `ThreadFactory` - and I have no idea how many of these are out there, but my guess is more than one.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
dimitarndimitrov: FWIW I prefer what you originally had, but with `response` replacing `executor`.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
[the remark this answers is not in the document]
ehumber: I think it makes the code fiddlier, but I've made the change, let me know what you think

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
ehumber: (And in the end I didn't, based on another of Dimitar's comments)

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -30,5 +39,38 @@ protected void configure() {
     bindAsContract(ChunkedOutputFactory.class);
     bindAsContract(StreamingResponseFactory.class);
     bind(Clock.systemUTC()).to(Clock.class);
+    bindFactory(ProduceResponseExecutorServiceFactory.class)
+        .to(ExecutorService.class)
```
dimitarndimitrov, line 43: I still think it's better to not have this act as a generic `ExecutorService` factory.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -30,5 +39,38 @@ protected void configure() {
...
+      ThreadFactory namedThreadFactory =
+          new ThreadFactoryBuilder().setNameFormat("Produce-response-thread-%d").build();
+      return new ThreadPoolExecutor(
+          produceResponseThreadPoolSize,
+          produceResponseThreadPoolSize,
+          0L,
+          TimeUnit.MILLISECONDS,
+          new LinkedBlockingQueue<Runnable>(),
+          namedThreadFactory);
+    }
```
dimitarndimitrov, line 69: nit: I actually think we should use this ctor, but with the core pool size being `Runtime.getRuntime().availableProcessors()` and the max pool size being the configurable size, defaulting to `Runtime.getRuntime().availableProcessors()` as well. If you prefer not to go for that, which I think is OK, because we will be fine tuning this code soon, you can just use `Executors.newFixedThreadPool(produceResponseThreadPoolSize, namedThreadFactory)`.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -30,5 +39,38 @@ protected void configure() {
...
+
+    @Override
+    public void dispose(ExecutorService executorService) {
+      executorService.shutdown();
+    }
```
dimitarndimitrov, line 74: not-a-comment: I still haven't learned how does HK2 behave for singletons - do you know when will this be called (also CC @rigelbm as he probably knows)?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
rigelbm: @ehumber can you add a `@ProduceResponseThreadPool` annotation to this executor?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
rigelbm: At the end of the instance life-cycle. For a `@Singleton` instance, that's when then `Injector` itself is destroyed, which is when the server shuts down.

rigelbm: @ehumber, do we need to `awaitTermination`/`shutdownNow` here, to make in-flight produce responses are dispatched?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
rigelbm: I would make everything configurable, and using `Runtime.getRuntime().availableProcessors()` as a reasonable default.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
rigelbm: @ehumber , by annotation, I mean what we do in `ConfigModule`, so that we inject it as `@MyAnnotation MyType` instead of just `MyType`.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
ehumber: @dimitarndimitrov is just naming it to have ProduceResponse in the title sufficient, or were you thinking of something additional?  In which case, can you explain in a little more detail please.

ehumber: @rigelbm I think I've done that in the commit I'm about to push, but let me know if it's not what you mean.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
ehumber: The default is set to Runtime.getRuntime().availableProcessors() in KafkaRestConfig, so hopefully that covers the default setting, and then it's configurable just via the normal properties.

Dimitar - thanks for pointing out that method, I looked and looked for it but obviously my eyes weren't working, so I made my own based on what Executors.newFixedThreadPool() does.  I've flipped back to using that.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
dimitarndimitrov: I think the problem is not with the name of the factory type, but with the registered type provided via that factory, which currently still seems to be the generic `ExecutorService` type. My inelegant suggestion would have been to introduce an `ExecutorService` subtype which to provide with this factory, and inject that subtype where we need it. I think Rigel is saying that this can also be done by simply annotating the injection point and the binding here with a new annotation specifically for the executor "flavor". Not sure how exactly does this need to look, but it sounds preferable to me.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
ehumber: Good spot Rigel :) fixed.

ehumber:
@rigelbm I'm not sure I have addressed the Annotation change as you'd expect.  Can you take a look in particular at V3ResourcesModule and check it's what you had in mind.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
rigelbm:
```
bindFactory(ProduceResponseExecutorServiceFactory.class)
    .qualifiedBy(new ProduceResponseThreadPoolImpl())
    .to(ExecutorService.class)
    .in(Singleton.class);

//...

@Qualifier
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.TYPE, ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER})
public @interface ProduceResponseThreadPool {}

private static final class ProduceResponseThreadPoolImpl
    extends AnnotationLiteral<ProduceResponseThreadPool>
    implements ProduceResponseThreadPool {}

// ...

private static final class ProduceResponseExecutorServiceFactory
    implements Factory<ExecutorService> {
    // ....
}
```

ehumber:
Rigel has asked that this be merged into 7.1.x, as that hasn't shipped yet.  https://github.com/confluentinc/kafka-rest/pull/934 opened to cover that.

rigelbm (approved):
LGTM! Nice!!

PR #931, opened 2021-11-17
Implement request rate-limiting for Kafka REST.

rigelbm:
This PR adds 6 configs:

 * rate.limit.enable
 * rate.limit.backend ('guava' or 'resilience4j')
 * rate.limit.permits.per.sec
 * rate.limit.timeout.ms
 * rate.limit.default.cost
 * rate.limit.costs (map of endpoint=cost,...)

If rate-limiting is enabled, requests to all endpoints are rate-limited to at most `rate.limit.permits.per.sec` per second, using the permit cost configured via `rate.limit.costs`, or `rate.limit.default.cost` if no cost has been configured for the endpoint.

Each request will wait up to `rate.limit.timeout.ms` milliseconds for the necessary permits to proceed, or fail with HTTP 429 if can't obtain the permits within the timeout. Endpoints annotated with `@DoNotRateLimit` (currently only `ProduceAction`) ignore rate-limiting.

There are two implementations of the rate-limiting algorithm: Guava and Resilence4j. The implementation can be chosen via `rate.limit.backend`. See https://www.alibabacloud.com/blog/detailed-explanation-of-guava-ratelimiters-throttling-mechanism_594820 and https://resilience4j.readme.io/docs/ratelimiter for more details about each implementation.

On a code level, this PR introduces two public types: `RateLimitFeature` and `RequestRatelimiter`. `RateLimitFeature` configures rate-limiting according the rules above. `RequestRatelimiter` is the singleton rate-limiter responsible for keeping track of all request rates. This rate-limiter can be injected for manual injection, but make sure the endpoint is annotated with `@DoNotRateLimit`, otherwise you might rate-limit twice.

The automatic rate-limiting of requests (using `rate.limit.costs` and `rate.limit.default.cost`) is accomplished by `FixedCostRateLimiRequestFilter`.

This PR also renames the existing `RateLimiter` to `v3.ProduceRateLimiter` since its logic is very specific to `v3.ProduceAction`. Ideally we should unify both rate-limiters, but I ran out of time trying to do that.

The rate-limiter introduced in this PR is supposed to be used in conjuction with rest-utils `DosFilter`. While `DosFilter` protects the server against all requests (including unauthenticated ones), the requests rate-limited by `RequestRateLimiter` are already authenticated.

### DIFF
--- a/kafka-rest/pom.xml
+++ b/kafka-rest/pom.xml
@@ -93,7 +93,10 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>io.github.resilience4j</groupId>
+            <artifactId>resilience4j-ratelimiter</artifactId>
+        </dependency>
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>rest-utils-test</artifactId>
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -15,6 +15,8 @@
 
 package io.confluent.kafkarest;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static io.confluent.kafka.serializers.AbstractKafkaSchemaSerDeConfig.AUTO_REGISTER_SCHEMAS;
 import static io.confluent.kafka.serializers.AbstractKafkaSchemaSerDeConfig.USE_LATEST_VERSION;
 import static java.util.Collections.emptyMap;
@@ -24,6 +26,7 @@ import static java.util.Objects.requireNonNull;
 import static org.apache.kafka.clients.CommonClientConfigs.METRICS_CONTEXT_PREFIX;
 
 import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.MapDifference;
 import com.google.common.collect.MapDifference.ValueDifference;
@@ -32,11 +35,14 @@ import io.confluent.kafka.serializers.AbstractKafkaSchemaSerDeConfig;
 import io.confluent.kafka.serializers.KafkaJsonSerializerConfig;
 import io.confluent.kafka.serializers.json.KafkaJsonSchemaSerializerConfig;
 import io.confluent.kafka.serializers.protobuf.KafkaProtobufSerializerConfig;
+import io.confluent.kafkarest.ratelimit.RateLimitBackend;
 import io.confluent.rest.RestConfig;
 import io.confluent.rest.RestConfigException;
 import io.confluent.rest.metrics.RestMetricsContext;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.time.Duration;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -365,6 +371,43 @@ public class KafkaRestConfig extends RestConfig {
       "Whether to enable REST Proxy V3 API. Default is true.";
   private static final boolean API_V3_ENABLE_DEFAULT = true;
 
+  public static final String RATE_LIMIT_ENABLE_CONFIG = "rate.limit.enable";
+  private static final String RATE_LIMIT_ENABLE_DOC =
+      "Whether to enable request rate-limiting. Default is false.";
+  private static final boolean RATE_LIMIT_ENABLE_DEFAULT = false;
+
+  public static final String RATE_LIMIT_BACKEND_CONFIG = "rate.limit.backend";
+  private static final String RATE_LIMIT_BACKEND_DOC =
+      "The rate-limiting backend to use. The options are 'guava' and 'resilience4j'. Default is "
+          + "'guava'.";
+  private static final String RATE_LIMIT_BACKEND_DEFAULT = "guava";
+
+  public static final String RATE_LIMIT_PERMITS_PER_SEC_CONFIG = "rate.limit.permits.per.sec";
+  private static final String RATE_LIMIT_PERMITS_PER_SEC_DOC =
+      "The maximum number of permits to emit per second. A permit is an unit of cost for a "
+          + "request. More expensive requests will consume more permits. The cost for each "
+          + "resource/method can configured via rate.limit.default.cost and rate.limit.costs. "
+          + "Default is 50.";
+  private static final Integer RATE_LIMIT_PERMITS_PER_SEC_DEFAULT = 50;
+
+  public static final String RATE_LIMIT_TIMEOUT_MS_CONFIG = "rate.limit.timeout.ms";
+  private static final String RATE_LIMIT_TIMEOUT_MS_DOC =
+      "How much to wait for the necessary permits to proceed, in milliseconds. A request that "
+          + "fails to acquire the necessary permits within the timeout will fail with HTTP 429. "
+          + "Default is 0ms.";
+  private static final long RATE_LIMIT_TIMEOUT_MS_DEFAULT = 0;
+
+  public static final String RATE_LIMIT_DEFAULT_COST_CONFIG = "rate.limit.default.cost";
+  private static final String RATE_LIMIT_DEFAULT_COST_DOC =
+      "The default cost for resources/methods without an explicit cost configured via"
+          + "rate.limit.costs. Default is 1.";
+  private static final int RATE_LIMIT_DEFAULT_COST_DEFAULT = 1;
+
+  public static final String RATE_LIMIT_COSTS_CONFIG = "rate.limit.costs";
+  private static final String RATE_LIMIT_COSTS_DOC =
+      "Map of rate-limit cost per endpoint. The map should be in the form 'api1=10,api2=20'.";
+  private static final String RATE_LIMIT_COSTS_DEFAULT = "";
+
   private static final ConfigDef config;
 
   static {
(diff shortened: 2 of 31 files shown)


kafka-rest/pom.xml
```diff
@@ -93,7 +93,10 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>io.github.resilience4j</groupId>
```
ehumber, line 97: Do we need to do any open source clearance or project management type stuff around adding a new dependency?

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -365,6 +371,43 @@
       "Whether to enable REST Proxy V3 API. Default is true.";
   private static final boolean API_V3_ENABLE_DEFAULT = true;
 
+  public static final String RATE_LIMIT_ENABLE_CONFIG = "rate.limit.enable";
```
ehumber, line 374: These new configs will all need the equivalent LD code adding too in the appropriate repo

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -365,6 +371,43 @@
...
+  private static final String RATE_LIMIT_ENABLE_DOC =
+      "Whether to enable request rate-limiting. Default is false.";
```
ehumber, line 376: Why false?  Previously any rate limiting is on by default, I think we should make this on by default too as it's safer to have rate limiting in place, and then it can be disabled if needed.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -365,6 +371,43 @@
...
+  private static final boolean RATE_LIMIT_ENABLE_DEFAULT = false;
+
+  public static final String RATE_LIMIT_BACKEND_CONFIG = "rate.limit.backend";
+  private static final String RATE_LIMIT_BACKEND_DOC =
```
ehumber, line 380: Why the choice of rate limiting back end?  Are we giving the option to test which is more efficient? If so, we should probably remove the option again once we know which is the preferred limiter.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -365,6 +371,43 @@
...
+  private static final String RATE_LIMIT_PERMITS_PER_SEC_DOC =
+      "The maximum number of permits to emit per second. A permit is an unit of cost for a "
+          + "request. More expensive requests will consume more permits. The cost for each "
+          + "resource/method can configured via rate.limit.default.cost and rate.limit.costs. "
+          + "Default is 50.";
+  private static final Integer RATE_LIMIT_PERMITS_PER_SEC_DEFAULT = 50;
+
+  public static final String RATE_LIMIT_TIMEOUT_MS_CONFIG = "rate.limit.timeout.ms";
+  private static final String RATE_LIMIT_TIMEOUT_MS_DOC =
+      "How much to wait for the necessary permits to proceed, in milliseconds. A request that "
```
ehumber, line 395: This is like the grace period for Produce. I was wondering if it should have "grace" in the variable name to indicate that (although I'm 50:50 on that).

The text should say "How much time to wait for the necessary...."  or "How long to wait for the necessary.... "

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -365,6 +371,43 @@
...
+
+  public static final String RATE_LIMIT_DEFAULT_COST_CONFIG = "rate.limit.default.cost";
+  private static final String RATE_LIMIT_DEFAULT_COST_DOC =
+      "The default cost for resources/methods without an explicit cost configured via"
+          + "rate.limit.costs. Default is 1.";
+  private static final int RATE_LIMIT_DEFAULT_COST_DEFAULT = 1;
+
+  public static final String RATE_LIMIT_COSTS_CONFIG = "rate.limit.costs";
+  private static final String RATE_LIMIT_COSTS_DOC =
+      "Map of rate-limit cost per endpoint. The map should be in the form 'api1=10,api2=20'.";
```
ehumber, line 408: Can the doc give an explicit example here of what "api" looks like. eg is it "api.v3.produce.*" or "api.v3.produce" or "/v3/clusters/{clusterId}/topics/{topicName}/records" or...

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -862,6 +941,42 @@ public boolean isV3ApiEnabled() {
...
+  public final ImmutableMap<String, Integer> getRateLimitCosts() {
+    String value = getString(RATE_LIMIT_COSTS_CONFIG);
+    return Arrays.stream(value.split(","))
+        .filter(entry -> !entry.isEmpty())
+        .peek(
+            entry ->
+                checkArgument(
+                    entry.contains("="),
+                    String.format(
+                        "Invalid value for config %s: %s", RATE_LIMIT_COSTS_CONFIG, value)))
```
ehumber, line 973: Potentially say what the format should be too (in error messages it's generally best to give a next action as well as the error).

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -862,6 +941,42 @@ public boolean isV3ApiEnabled() {
...
+        .collect(
+            toImmutableMap(
+                entry -> entry.substring(0, entry.indexOf('=')),
```
ehumber, line 976: Consider trimming any whitespace here.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RateLimitExceededException.java
```diff
@@ -0,0 +1,34 @@
...
+
+import io.confluent.kafkarest.exceptions.StatusCodeException;
+import javax.ws.rs.core.Response;
+
+/**
+ * An exception thrown when the request rate-limit has been exceeded.
+ *
+ * @see RequestRateLimiter#rateLimit(int)
+ */
+public final class RateLimitExceededException extends StatusCodeException {
```
ehumber, line 26: Consider merging with the RateLimitGracePeriodExceededException - they are doing the same thing.  (Maybe the Grace period exception needs a rename)

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/FixedCostRateLimitRequestFilter.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+
+/**
+ * A {@link ContainerRequestFilter} that automatically applies a request rate-limit at a fixed cost
+ * based on the resource/method being requested, according to the {@link
+ * io.confluent.kafkarest.KafkaRestConfig#RATE_LIMIT_COSTS_CONFIG} and {@link
+ * io.confluent.kafkarest.KafkaRestConfig#RATE_LIMIT_DEFAULT_COST_CONFIG} configs.
+ */
+final class FixedCostRateLimitRequestFilter implements ContainerRequestFilter {
```
ehumber, line 31: I like how you've injected this behaviour using the Filter, nice thinking :)

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
ehumber: I think having two backends (actually 3, if you include the null rate limiter) significantly increases the complexity of the code too - we need delegates and different factories.  With only one limiter, I think a lot of the layering of the classes could potentially be removed.  

I'm not convinced a null rate limiter is required if we only have one rate limiter option - for each call to an endpoint the code is having to work out which rate limiter to use.  If you have just the one, it's probably more efficient to go "if rate limiting disabled, just return" next to the call to the guava/resilience4j call.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -365,6 +371,43 @@
...
+  private static final boolean RATE_LIMIT_ENABLE_DEFAULT = false;
+
+  public static final String RATE_LIMIT_BACKEND_CONFIG = "rate.limit.backend";
+  private static final String RATE_LIMIT_BACKEND_DOC =
+      "The rate-limiting backend to use. The options are 'guava' and 'resilience4j'. Default is "
+          + "'guava'.";
+  private static final String RATE_LIMIT_BACKEND_DEFAULT = "guava";
+
+  public static final String RATE_LIMIT_PERMITS_PER_SEC_CONFIG = "rate.limit.permits.per.sec";
+  private static final String RATE_LIMIT_PERMITS_PER_SEC_DOC =
```
ehumber, line 386: I like that it's possible to give different endpoints a different rate limit, but I think the hard work here is on the user to do the maths - our logic should be able to do it.

What we want is to allow the user to configure that endpointA can do N requests per second (which is the rate limit measurement we use elsewhere, rather than costs and permits) and endpointB can do N+5 requests per second, as that's the familiar way of thinking about rate limits.

Having both costs per endpoint and a permit to aim for makes the user do the calculation.

I wonder if we can have a default rate limit per second (as there is already) and then costs is renamed to eg endpoints.rate.limits, and then from a user point of view each endpoint just has a rate limit in the map specified.

Then permits could go.  We'd have to work out how to translate the users requests per second for an endpoint in the rateLimit(costs) method... but actually, isn't costs just the rate limit for the endpoint anyway (although I suspect I'm missing something in thinking through that)?

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: This configuration is false to make upgrading easier. People upgrading to a newer version won't have to do additional work. They then can enable it at their own convenience.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: I actually removed a whole layer of dependency injection by using a dynamic-feature instead (in the latest commit). That should make it much more efficient (the rate-limiter decision is only made once, at startup, per endpoint). I have two backends because I don't have enough data to decide which one is better. They use different algorithms, and on the tests, behave differently. I think once we merge this PR, it's unlikely we will remove the backend choice (for backwards compatibility). I don't think that's a problem though.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: This is not like the grace period. The grace period allows requests to proceed even though they are over the limit, for a small amount of time. This timeout actually blocks the request from proceeding until the rate is below the limit.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: I actually think it would be harder to explain it to users. When you say "endpointA can do N requests per second" what you really mean is "in the absence of any other requests, endpointA can do N requests per second, but if there are other concurrent requests, then go figure". It's very hard to encode what the total-permits is in this case. The cost/permit way is, in my opinion, the most natural way of thinking of rate-limiting/QoS. The math is easy to reason about, and not a huge problem.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/FixedCostRateLimitRequestFilter.java
rigelbm: The one thing you liked in the PR is the one thing I removed, sorry :). Removed the injection to avoid the per-request cost of figuring out the cost/rate-limiter. Now this is configured at server start up.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RateLimitExceededException.java
rigelbm: I explicitly didn't reuse anything from this rate-limiter in the produce-rate-limiter. I think there's much to be reused, but I didn't want more time (or rather didn't have more time to spend).

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Yes.

kafka-rest/pom.xml
rigelbm: This library uses Apache 2 license. I don't think we need to do any extra work.

ehumber (approved):

rigelbm:
Red build is due to already failing `ce-kafka-rest` downstream validation. Ignoring.

PR #933, opened 2021-11-19
Bind NoSchemaRecordSerializer as Singleton.

rigelbm:
This is to avoid creating a new NoSchemaRecordserializer for each record in a stream. Normally, this instances should be very light and short-lived, so instantiating them should not be a problem. Unfortunately, a new `KafkaJsonSerializerConfig` is created for each `NoSchemaRecordSerializer`, and that thing is everything but light. In special, `AbstractConfig` (super-type) has the bad habit of logging all configs at creation, and when you that doing thousands of times per second, that usually takes a lot of cpu.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
@@ -15,6 +15,7 @@
 
 package io.confluent.kafkarest.controllers;
 
+import javax.inject.Singleton;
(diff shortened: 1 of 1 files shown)


dimitarndimitrov (approved):
+1.

rigelbm:
Red build is due to downstream validation failure on the already broken `ce-kafka-rest`. Ignoring.

ijuma:
`AbstractConfig` has the following constructor:

> public AbstractConfig(ConfigDef definition, Map<?, ?> originals, boolean doLog) {

PR #940, opened 2021-11-25
Add an extra slack around rate-limiting tests.

rigelbm:
That would remove most of all the rate-limit test failures in previous runs.


ehumber:
Is there an understanding of why this test isn't exact?  Do the rate limiters not give consistent responses, or is it because we have the equivalents of sleeps in the test (with the executor service scheduling messages using the system clock every n ms).

Can you also link to the jira that covers this please.

ehumber (approved):

Issue #341, opened 2017-09-14
Consumer read takes long time with multiple consumers in consumer group


PlatEng:
A fundamental issue is seen with REST Proxy. Suppose if we create two consumer instances under the same consumer group, read call is taking longer time with calls with subsequent instance.

Steps to reproduce:
1. Create my_consumer_instance1 and my_consumer_instance2 under my_json_consumer1:
curl -H "Content-Type: application/vnd.kafka.v2+json" -d '{"name": "my_consumer_instance1", "format": "json", "auto.offset.reset": "earliest"}' http://localhost:8082/consumers/my_json_consumer1
curl -H "Content-Type: application/vnd.kafka.v2+json" -d '{"name": "my_consumer_instance2", "format": "json", "auto.offset.reset": "earliest"}' http://localhost:8082/consumers/my_json_consumer1

2. Subscribe both the instances to topic test1:
curl -H "Content-Type: application/vnd.kafka.v2+json" -d '{"topics":["test1"]}' http://localhost:8082/consumers/my_json_consumer1/instances/my_consumer_instance1/subscription
curl -H "Content-Type: application/vnd.kafka.v2+json" -d '{"topics":["test1"]}' http://localhost:8082/consumers/my_json_consumer1/instances/my_consumer_instance2/subscription

3. Read the message from my_consumer_instance1 several times and we will get responses back:
curl -X GET -H "Accept: application/vnd.kafka.json.v2+json" http://localhost:8082/consumers/my_json_consumer1/instances/my_consumer_instance1/records

4. Read the message from my_consumer_instance2 and the call will hang. It will take around 8 minutes to get a response back:
curl -X GET -H "Accept: application/vnd.kafka.json.v2+json" http://localhost:8082/consumers/my_json_consumer1/instances/my_consumer_instance2/records

5. Again read the message from my_consumer_instance1 and the call will hang. It will take some time to get a response back:
curl -X GET -H "Accept: application/vnd.kafka.json.v2+json" http://localhost:8082/consumers/my_json_consumer1/instances/my_consumer_instance1/records

ZachTB123:
Any updates on this?

woodlee:
See my comment over on https://github.com/confluentinc/kafka-rest/issues/396

tPl0ch:
This is pretty much happening for all our teams using the REST proxy. So everyone is only able to run a single consumer within a group currently and this defies all advantages you want to have.

fkaufer:
[...]

Issue #789, opened 2021-01-12
Multiple consumers for a single topic fail for rest-proxy starting with branch 6.0.1-post


sbange-reply:
We are consuming a single topic with multiple consumers using the rest-proxy, which works fine up to branch 6.0.0-post. Recently we updated to branch 6.0.1-post (or the latest docker hub confluent image) and encountered an error while trying to join additional consumers on a single topic. On join it would give us the (level INFO) message:

`Attempt to heartbeat failed since group is rebalancing (org.apache.kafka.clients.consumer.internals.AbstractCoordinator)`

until the initially joined consumer would timeout and get disposed. Only then the next consumer would be able to join the same topic and read. To replicate the error:

1. clone confluentinc/cp-all-in-one
2. checkout branch 6.0.1-post
3. change to dir cp-all-in-one
4. docker-compose up
5. run the following script
6. check kafka rest-proxy logs and script output

```bash
#!/bin/bash
# get cluster id for all subsequent operations
KAFKA_CLUSTER_ID=$(docker-compose exec rest-proxy curl -X GET "http://localhost:8082/v3/clusters/" | jq -r ".data[0].cluster_id")
echo "... deleting test topic"
docker-compose exec rest-proxy curl -X DELETE -H "Content-Type: application/json" -d "{\"topic_name\":\"test-topic\",\"partitions_count\":3,\"replication_factor\":1,\"configs\":[]}" "http://localhost:8082/v3/clusters/${KAFKA_CLUSTER_ID}/topics/test-topic" | jq '.'
echo "... creating test topic"
docker-compose exec rest-proxy curl -X POST -H "Content-Type: application/json" -d "{\"topic_name\":\"test-topic\",\"partitions_count\":3,\"replication_factor\":1,\"configs\":[]}" "http://localhost:8082/v3/clusters/${KAFKA_CLUSTER_ID}/topics" | jq '.'
echo "...  producing a single message to every of the three partitions"
docker-compose exec rest-proxy curl -X POST -H "Content-Type: application/vnd.kafka.json.v2+json" -H "Accept: application/vnd.kafka.v2+json" --data '{"records":[{"key":"alice","value":{"count":0}},{"key":"alice","value":{"count":1}},{"key":"alice","value":{"count":2}}]}' "http://localhost:8082/topics/test-topic" | jq '.'
[...]

PR #942, opened 2021-11-26
Fix(#341): Remove deadlock when rebalancing by multiple consumer instances

merrygoround-of-life:
this is to fix the issue: #341
when rebalancing starts by multiple instances in a consumer group, a pseudo deadlock condition occurs:

thread a. (Consumer Expiration Thread)
 1) acquired the KafkaConsumerManager instance (lock A) within KafkaConsumerManager$ExpirationThread::Run
 2) waiting to acquire the KafkaConsumerState instance (lock B) within KafkaConsumerState::expired
thread-a-monitor

thread b.
 1) acquired the KafkaConsumerState instance (lock B) within KafkaConsumerState::hasNext in KafkaConsumerReadTask::addRecords
 2) waiting to acquire the ?? (lock C) within KafkaConsumer::poll(0) by rebalancing
thread-b-running

thread c.
 1) acquired the ?? (lock C) within jettyHandler
 2) waiting to acquire the KafkaConsumerManager instance (lock A) within KafkaConsumerManager::getConsumerInstance in readRecords
thread-c-monitor

a cyclic lock order is generated by 3 locks and 3 threads, resulting in deadlock.
to resolve this, split the KafkaConsumerState instance lock (lock B) into two locks - a KafkaConsumerState instance lock(existing) and KafkaConsumerState.expiration member lock(new).
as a result, lock B of thread a and lock B of thread b become different locks, breaking the cycle.

since thread b is actually in the running state, the lock C is not exactly the the Java synchronization object.
perhaps it needs permission(lock) to use jetty's handler when certain events are triggered by rebalancing inside KafkaConsumer::poll, which I've referred to as the lock C.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
@@ -65,16 +65,8 @@ public abstract class KafkaConsumerState<KafkaKeyT, KafkaValueT, ClientKeyT, Cli
 
   private final Queue<ConsumerRecord<KafkaKeyT, KafkaValueT>> consumerRecords = new ArrayDeque<>();
 
-  class Expiration {
-
-    long value;
-
-    Expiration(long expiration) {
-      this.value = expiration;
-    }
-  }
-
-  volatile Expiration expiration;
+  volatile long expiration;
+  private final Long expirationLock;
 
   KafkaConsumerState(
       KafkaRestConfig config,
@@ -84,8 +76,9 @@ public abstract class KafkaConsumerState<KafkaKeyT, KafkaValueT, ClientKeyT, Cli
     this.config = config;
     this.instanceId = instanceId;
     this.consumer = consumer;
-    this.expiration = new Expiration(config.getTime().milliseconds()
-                      + config.getInt(KafkaRestConfig.CONSUMER_INSTANCE_TIMEOUT_MS_CONFIG));
+    this.expiration = config.getTime().milliseconds()
+                      + config.getInt(KafkaRestConfig.CONSUMER_INSTANCE_TIMEOUT_MS_CONFIG);
+    this.expirationLock = this.expiration;
   }
 
   public ConsumerInstanceId getId() {
@@ -345,14 +338,14 @@ public abstract class KafkaConsumerState<KafkaKeyT, KafkaValueT, ClientKeyT, Cli
   }
 
   public boolean expired(long nowMs) {
-    synchronized (this.expiration) {
-      return expiration.value <= nowMs;
+    synchronized (this.expirationLock) {
+      return expiration <= nowMs;
     }
   }
 
   public void updateExpiration() {
-    synchronized (this.expiration) {
-      this.expiration.value = config.getTime().milliseconds()
+    synchronized (this.expirationLock) {
+      this.expiration = config.getTime().milliseconds()
                         + config.getInt(KafkaRestConfig.CONSUMER_INSTANCE_TIMEOUT_MS_CONFIG);
     }
   }
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
@@ -66,7 +66,7 @@ public abstract class KafkaConsumerState<KafkaKeyT, KafkaValueT, ClientKeyT, Cli
   private final Queue<ConsumerRecord<KafkaKeyT, KafkaValueT>> consumerRecords = new ArrayDeque<>();
 
   volatile long expiration;
-  private final Long expirationLock;
+  private final Object expirationLock = new Object();
 
   KafkaConsumerState(
       KafkaRestConfig config,
@@ -78,7 +78,6 @@ public abstract class KafkaConsumerState<KafkaKeyT, KafkaValueT, ClientKeyT, Cli
     this.consumer = consumer;
     this.expiration = config.getTime().milliseconds()
                       + config.getInt(KafkaRestConfig.CONSUMER_INSTANCE_TIMEOUT_MS_CONFIG);
-    this.expirationLock = this.expiration;
   }
 
   public ConsumerInstanceId getId() {
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
@@ -65,7 +65,7 @@ public abstract class KafkaConsumerState<KafkaKeyT, KafkaValueT, ClientKeyT, Cli
 
   private final Queue<ConsumerRecord<KafkaKeyT, KafkaValueT>> consumerRecords = new ArrayDeque<>();
 
-  volatile long expiration;
+  volatile Long expiration;
 
   KafkaConsumerState(
       KafkaRestConfig config,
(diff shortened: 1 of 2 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -66,6 +66,7 @@
   private final Queue<ConsumerRecord<KafkaKeyT, KafkaValueT>> consumerRecords = new ArrayDeque<>();
 
   volatile long expiration;
+  private final Long expirationLock;
```
plantrue, line 69: Added a separate final object for synchronizing the expiration field.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -335,13 +337,17 @@ synchronized long getEndOffset(String topic, int partition) {
...
+    synchronized (this.expirationLock) {
+      return expiration <= nowMs;
+    }
   }
 
-  public synchronized void updateExpiration() {
-    this.expiration = config.getTime().milliseconds()
-                      + config.getInt(KafkaRestConfig.CONSUMER_INSTANCE_TIMEOUT_MS_CONFIG);
+  public void updateExpiration() {
+    synchronized (this.expirationLock) {
```
plantrue, line 347: Resolve a deadlock by making the scope of synchronized small.
(replacing the KafkaConsumerState instance lock with a new KafkaConsumerState::expirationLock instance member lock)

klboke:
@merrygoround-of-life  Hi, what a coincidence, I also found this problem yesterday, my jstack analysis results are somewhat similar to your exposition, but not exactly the same. Finally, did you solve your problem?

merrygoround-of-life:
@klboke the stack information may be slightly different depending on the kafka-rest version you use. but i think it would be the same problem i found if you could see a thread waiting indefinitely for the KafkaConsumer::poll function to return.
Finally, it was very difficult to reproduce that issue with this modification.

klboke:
@merrygoround-of-life  It should be the same problem, this is the stack information on my side, eventually all waiting for the poll operation, while the poll operation is waiting for rebalancing
image

merrygoround-of-life:
according to my analysis, a deadlock was caused by 3 or more locks and threads, but it was not clear who the C lock was.
- thread a: lock A -> lock B
- thread b: lock B -> lock C
- thread c: lock C -> lock A
In your diagram, it appears that lock A (c22fa4d0) corresponds to a KafkaConsumerManager singleton instance and lock B (c3bdf508) corresponds to a KafkaConsumerState instance. (A KafkaConsumerState instance is created for each Consumer instance) however it's still not clear from your diagram who the third lock that completes the circular waiting is.

Presumably some of the threads called getConsumerInstance function are expected to be in the context of request handler of the jetty web server and it is presumed that the thread waiting for the return of KafkaConsumer::poll(0) in addRecords function need to acquire some lock associated with the jetty handler for rebalancing which I've referred to as the lock C. but no matter who lock C is, my solution is to split the lock B to break the circular waiting.

what version of kafka-rest have you tested?
I'll check to see if this solution is equally valid in the kafka-rest of your version,

merrygoround-of-life:
it seems to be related to issues #789 and #587 as well.

klboke:
I'm using the https://github.com/confluentinc/kafka-rest/tree/6.2.1-post version. Your analysis is correct, but it seems that there is no lock on C. All thread operations end up being blocked by the addRecords lock, which makes it impossible for the thread to clean up the expired consumer. The following diagram.
image
addRecords holds a hasNext lock, blocking getConsumerInstance, blocking ExpirationThread.run

merrygoround-of-life:
@klboke I confirmed that in my use case the deadlock was equally reproduced in the 6.2.1-post version and it was not reproduced when this modification was applied.
Can you check if it is solved in your use case by applying the attached patch file?
- [patch.txt](https://github.com/confluentinc/kafka-rest/files/7821000/patch.txt)
> patch -p1 < patch.txt

merrygoround-of-life:
@klboke I confirmed that in my use case the deadlock was equally reproduced in the 6.2.1-post version and it was not reproduced when this modification was applied.
Can you check if it is solved in your use case by applying the attached patch file?
- [patch.txt](https://github.com/confluentinc/kafka-rest/files/7821103/patch.txt)
> patch -p1 < patch.txt

klboke:
It should be solved, as I posted above, according to your pr so modified, ExpirationThread.run will be able to run properly, I think the key point is here

merrygoround-of-life:
@klboke
the hypothesis of my analysis is that 'KafkaConsumer::poll function is not returning because a deadlock has occurred'.
it is known that a deadlock occurs when multiple threads need the same locks but obtain them in different order,
so if there are only two locks that caused the deadlock, you should be able to see the A -> B lock acquisition order and vice versa.
however in our analysis, we found an A -> B lock acquisition order but not a reverse acquisition order (B -> A).

if there is another undiscovered lock (C) associated with the deadlock, it can be assumed that the lock acquisition order of B -> (C) -> A exists (though not explicitly seen) and thus the deadlock is established.

148019839-49e986ee-eb28-46f3-b5ea-62990ffbb948

A scenario is possible that the KafkaConsumer::poll function is not returning because the KafkaConsumer::poll function is trying to acquire the lock C internally, but has already been acquired by another thread (finally calling the KafkaConsumer::poll function).
i'm guessing this has something to do with the issues found when searching for 'jetty deadlock'.

klboke:
@merrygoround-of-life 
To be sure, there is no deadlock found in my program stack. If there was a deadlock, it would print [Found 1 deadlock] in the jstack result, however, there is not, and this is also true from the analysis of the program stack. All the blocking ends up waiting for KafkaConsumer::poll, ExpirationThread is also blocked, and updating the expiration time is also blocked, here is the key point

merrygoround-of-life:
@klboke you are right. this issue is not a ‘real’ deadlock consisting of only java sync objects. that's why i used the term 'pseudo deadlock' (or logical deadlock).
i don't think it can explain why the KafkaConsumer::poll function doesn't block when i split the lock B, unless it’s a deadlock (circular wait).

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
rigelbm: Can you just do `Object expirationLock = new Object()` instead? We don't actually need a `Long` for anything.

merrygoround-of-life: @rigelbm No problem. I'll fix it soon based on your comment.

rigelbm (approved):
LGTM

merrygoround-of-life:
@vharitonsky 'Only those with write access to this repository can merge pull requests.'
@rigelbm Could you by any chance merge this? or is there a guide to do this own my own?

vharitonsky:
Guys, this merge is very important for our infrastructure, if you could maybe look into it?

ehumber:
@vharitonsky  I'm looking at this at the moment - aiming to merge it very soon.

ehumber (approved):
Have tested locally and reviewed the changes.  All looks good to me.

ehumber:
@merrygoround-of-life Thanks for solving this, and delivering the code fix, really appreciate you to doing it.

I've merged your change into 6.0.x, and also merged it into all later release branches.

techexplorer0310:
Hi, 

In which version should we get the fix?

Thanks,

PR #946, opened 2021-12-03
KREST-3255 Fix typo in 10 MB

ehumber:
Default for number of bytes per second was 10K not 10MB

All our other rate limiters are off by default, so I've updated the produce limiter to match


dimitarndimitrov (approved):
+1.

Thanks for addressing these! Now that I re-read the docs, I wonder if we should explicitly mention something like "Produce v3" there instead of just a generic "produce", because I can imagine that until we sunset the Produce v2 API, people might mistakenly think that these also apply to it.

rigelbm (approved):
LGTM

PR #948, opened 2021-12-06
KREST-3255 Fix missing annotations for produce limit

ehumber:
The counter limit for rest produce was being set to 1000, regardless of what the config set it to.  It's not clear where the 1000 default config comes from as we would have expected an error here, but it was due to missing annotations in ConfigModule.


dimitarndimitrov (approved):
+1.

ehumber:
Known flakey test failure - [ERROR]   Resilience4jRateLimitTest.rateLimitWithDefaultCost:22->AbstractRateLimitEnabledTest.rateLimitWithDefaultCost:61->AbstractRateLimitEnabledTest.assertInRange:87 718 not in [500, 500]

PR #949, opened 2021-12-06
KREST-3255 fix rest producer count limit missing annotations

ehumber:
The counter limit for rest produce was being set to 1000, regardless of what the config set it to. It's not clear where the 1000 default config comes from as we would have expected an error here, but it was due to missing annotations in ConfigModule.


dimitarndimitrov (approved):
+1.

ehumber:
CI failure due to known flakey test

[ERROR]   Resilience4jRateLimitTest.rateLimitWithClassCost:22->AbstractRateLimitEnabledTest.rateLimitWithClassCost:73->AbstractRateLimitEnabledTest.assertInRange:87 217 not in [250, 250]

PR #950, opened 2021-12-06
KREST-3430 Optimize rate-limiting exceptions

dimitarndimitrov:
The `StatusCodeExceptions` that we throw in great numbers when rate-limiting can take a significant amount of time for filling their stack traces (up to 1% of a flamegraph taken during rate-limiting). These stack traces however are not really interesting for us, as the exceptions themselves always originate from our rate-limiting infrastructure.

The change overrides `fillInStackTrace()` instead of utilizing the `RuntimeException` ctor that takes a `writableStackTrace` flag, because while slightly more efficient, the latter approach is much more cumbersome.

Tested vs a local Kafka REST instance loaded using JMeter, by checking that the `java/lang/RuntimeException.<init>` -> `java/lang/Exception.<init>` -> `java/lang/Throwable.<init>` -> `java/lang/Throwable.fillInStackTrace` stack trace was optimized away from the captured flamegraph after this change.


ehumber (approved):

dimitarndimitrov:
The flaky `io.confluent.kafkarest.ratelimit.Resilience4jRateLimitTest.rateLimitWithClassCost` strikes again... It is unrelated to this though, so I will proceed with merging it.

PR #953, opened 2021-12-08
Close produce stream on idle time and unauthenticated.

rigelbm:
When handling produce requests, the request handling thread is held for as long as the input stream is open. Seeing we use a fixed size thread pool for request handling, a bad user could potentially deny any request from being handled by starving all request handling threads with open produce streams. That's attack is made worse by:

 1) we don't close the stream on unauthenticated from Kafka, so the user doesn't even need to have valid credentials
 2) we don't close the stream on idle time, so the user doesn't even need to send any data

This PR closes the two holes mentioned above. The server will now close the stream both if it hasn't seen requests in the stream within some or if any request produces an unauthenticated response from Kafka. Effectively, what that achieves is that we are forcing now each stream to authenticate with Kafka within some predefined time.

You can configure the idle time for the stream using the property `streaming.response.max.idle.time.ms` (default to 30s). Notice that there's an initial negotiation step where the stream is open but no bytes have actually been sent over the stream. In this case, the idle time is controlled not by Kafka REST, by Jetty, which already defaults it to 30s. A follow up PR will be done in `rest-utils` to add a way to configure this other Jetty idle time. It's strongly recomended that both idle times to be configured to the same value.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
@@ -15,15 +15,21 @@
 
 package io.confluent.kafkarest.resources.v3;
 
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.util.concurrent.SimpleTimeLimiter;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import io.confluent.kafkarest.config.ConfigModule.ProduceResponseThreadPoolSizeConfig;
+import io.confluent.kafkarest.config.ConfigModule.StreamingResponseMaxIdleTimeConfig;
 import io.confluent.kafkarest.response.ChunkedOutputFactory;
 import io.confluent.kafkarest.response.StreamingResponseFactory;
+import io.confluent.kafkarest.response.StreamingResponseIdleTimeLimiter;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.time.Clock;
+import java.time.Duration;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
@@ -42,9 +48,18 @@ public final class V3ResourcesModule extends AbstractBinder {
     bindAsContract(ProduceRateLimiters.class).in(Singleton.class);
     bindAsContract(ChunkedOutputFactory.class);
     bindAsContract(StreamingResponseFactory.class);
+    bindFactory(StreamingResponseIdleTimeLimiterFactory.class)
+        .to(StreamingResponseIdleTimeLimiter.class)
+        .in(Singleton.class);
+
     bind(Clock.systemUTC()).to(Clock.class);
+
     bindFactory(ProduceResponseExecutorServiceFactory.class)
         .qualifiedBy(new ProduceResponseThreadPoolImpl())
+        .to(ExecutorService.class)
+        .in(Singleton.class);
+    bindFactory(StreamingResponseIdleTimeExecutorFactory.class)
+        .qualifiedBy(new StreamingResponseIdleTimeExecutorImpl())
         .to(ExecutorService.class)
         .in(Singleton.class);
   }
@@ -56,6 +71,15 @@ public final class V3ResourcesModule extends AbstractBinder {
 
   private static final class ProduceResponseThreadPoolImpl
       extends AnnotationLiteral<ProduceResponseThreadPool> implements ProduceResponseThreadPool {}
+
+  @Qualifier
+  @Retention(RetentionPolicy.RUNTIME)
+  @Target({ElementType.TYPE, ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER})
+  public @interface StreamingResponseIdleTimeExecutor {}
+
+  private static final class StreamingResponseIdleTimeExecutorImpl
+      extends AnnotationLiteral<StreamingResponseIdleTimeExecutor>
+      implements StreamingResponseIdleTimeExecutor {}
 
   private static final class ProduceResponseExecutorServiceFactory
       implements Factory<ExecutorService> {
@@ -71,7 +95,7 @@ public final class V3ResourcesModule extends AbstractBinder {
     @Override
     public ExecutorService provide() {
       ThreadFactory namedThreadFactory =
-          new ThreadFactoryBuilder().setNameFormat("Produce-response-thread-%d").build();
+          new ThreadFactoryBuilder().setNameFormat("kafka-rest-produce-response-thread-%d").build();
       return Executors.newFixedThreadPool(produceResponseThreadPoolSize, namedThreadFactory);
     }
 
@@ -86,5 +110,51 @@ public final class V3ResourcesModule extends AbstractBinder {
         executorService.shutdownNow();
       }
     }
+  }
+
+  private static final class StreamingResponseIdleTimeExecutorFactory
+      implements Factory<ExecutorService> {
+
+    @Override
+    public ExecutorService provide() {
+      return Executors.newCachedThreadPool(
+          new ThreadFactoryBuilder()
+              .setNameFormat("kafka-rest-streaming-response-idle-time-thread-%d")
+              .build());
+    }
+
+    @Override
+    public void dispose(ExecutorService executorService) {
+      executorService.shutdown();
+      try {
+        if (!executorService.awaitTermination(60, TimeUnit.SECONDS)) {
+          executorService.shutdownNow();
+        }
+      } catch (InterruptedException e) {
+        executorService.shutdownNow();
+      }
+    }
+  }
+
+  private static final class StreamingResponseIdleTimeLimiterFactory
+      implements Factory<StreamingResponseIdleTimeLimiter> {
+    private final ExecutorService executor;
+    private final Duration maxIdleTime;
+
+    @Inject
+    StreamingResponseIdleTimeLimiterFactory(
+        @StreamingResponseIdleTimeExecutor ExecutorService executor,
+        @StreamingResponseMaxIdleTimeConfig Duration maxIdleTime) {
+      this.executor = requireNonNull(executor);
+      this.maxIdleTime = requireNonNull(maxIdleTime);
+    }
+
+    @Override
+    public StreamingResponseIdleTimeLimiter provide() {
+      return new StreamingResponseIdleTimeLimiter(SimpleTimeLimiter.create(executor), maxIdleTime);
+    }
+
+    @Override
+    public void dispose(StreamingResponseIdleTimeLimiter idleTimeLimiter) {}
   }
 }
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
@@ -25,16 +25,21 @@ import com.fasterxml.jackson.jaxrs.base.JsonMappingExceptionMapper;
 import com.fasterxml.jackson.jaxrs.base.JsonParseExceptionMapper;
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 import io.confluent.kafkarest.common.CompletableFutures;
 import io.confluent.kafkarest.exceptions.StatusCodeException;
 import io.confluent.kafkarest.exceptions.v3.ErrorResponse;
 import io.confluent.kafkarest.exceptions.v3.V3ExceptionMapper;
+import io.confluent.kafkarest.response.StreamingResponseIdleTimeLimiter.MaxIdleTimeExceededException;
+import io.confluent.kafkarest.response.StreamingResponseIdleTimeLimiter.MaxIdleTimeExceededExceptionMapper;
 import io.confluent.rest.entities.ErrorMessage;
 import io.confluent.rest.exceptions.KafkaExceptionMapper;
 import io.confluent.rest.exceptions.WebApplicationExceptionMapper;
 import java.io.IOException;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
 import javax.annotation.Nullable;
 import javax.ws.rs.WebApplicationException;
@@ -75,6 +80,11 @@ public abstract class StreamingResponse<T> {
               response -> Status.BAD_REQUEST.getStatusCode(),
               response -> (String) response.getEntity())
           .putMapper(
+              MaxIdleTimeExceededException.class,
+              new MaxIdleTimeExceededExceptionMapper(),
+              response -> ((ErrorResponse) response.getEntity()).getErrorCode(),
+              response -> ((ErrorResponse) response.getEntity()).getMessage())
+          .putMapper(
               StatusCodeException.class,
               new V3ExceptionMapper(),
               response -> ((ErrorResponse) response.getEntity()).getErrorCode(),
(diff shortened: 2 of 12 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -87,4 +111,50 @@ public void dispose(ExecutorService executorService) {
       }
     }
   }
+
+  private static final class StreamingResponseIdleTimeExecutorFactory
+      implements Factory<ExecutorService> {
+
+    @Override
+    public ExecutorService provide() {
+      return Executors.newCachedThreadPool(
```
ehumber, line 120: Why a cached thread pool, rather than a fixed thread pool?

https://www.baeldung.com/java-executors-cached-fixed-threadpool

In particular 
"As a result, fixed thread pools are better suited for tasks with unpredictable execution times." - which is true here I think, our tasks will have unpredictable times

 "cached pool starts with zero threads and can potentially grow to have Integer.MAX_VALUE" - and while I think that's OK, as we shouldn't allow that many requests in parallel, it makes me a little nervous in case we are forgetting to clear up somewhere and leak them - blowing up the JVM is worse than stopping Produce due to lack of threads being available in a fixed size pool.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -141,31 +154,56 @@ private ResultOrError handleNext(T result, @Nullable Throwable error) {
...
     }
 
     public void close() {
-      try {
-        mappingIteratorInput.close();
-      } catch (IOException e) {
-        log.error("Error when closing the request stream.", e);
-      }
+      closed = true;
     }
```
ehumber, line 173: Why has the close of the mapping iterator been removed here?  If I remember correctly, if the mapping Iterator isn't closed off, then it continues to read data from the socket, even after the sink is closed.

testing/image/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM 368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc/cp-kafka-rest:master-latest
+FROM 368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc/cp-kafka-rest:7.1.x-latest
```
ehumber, line 15: How do we remember to flip this back to master in the pint merge?

testing/environments/minimal/run.sh
```diff
@@ -37,16 +37,15 @@ aws ecr get-login-password --region us-west-2 \
  | docker login --username AWS --password-stdin 368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc/
 
 # Download latest cp-zookeeper image.
-docker pull 368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc/cp-zookeeper:master-latest
-
+docker pull 368821881613.dkr.ecr.us-west-2.amazonaws.com/confluentinc/cp-zookeeper:7.1.x-latest
```
ehumber, line 40: Same as below - how do we remember to flip this back to master for the pint merge? I wonder if (in a follow up PR) this could be parameterised so we don't have to remember each time we branch a release to go and update it.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponseIdleTimeLimiter.java
```diff
@@ -0,0 +1,67 @@
...
+
+  public StreamingResponseIdleTimeLimiter(TimeLimiter delegate, Duration maxIdleTime) {
+    this.delegate = requireNonNull(delegate);
+    this.maxIdleTime = requireNonNull(maxIdleTime);
+  }
+
+  <T> T callWithTimeout(Callable<T> callable)
+      throws ExecutionException, InterruptedException, MaxIdleTimeExceededException {
+    try {
+      return delegate.callWithTimeout(callable, maxIdleTime);
```
ehumber, line 41: If we are planning on having reslience4j as our rate limiter (due to it being able to deal with spikes of data better than guava), should we also use it here?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -141,31 +154,56 @@ private ResultOrError handleNext(T result, @Nullable Throwable error) {
...
 
     @Override
     public boolean hasNext() {
-      return mappingIteratorInput.hasNext();
+      if (closed) {
+        return false;
+      }
+      try {
+        return idleTimeLimiter.callWithTimeout(mappingIteratorInput::hasNextValue);
```
dimitarndimitrov, line 181: I'd like to check my understanding of the subtle change regarding the checked vs the unchecked version of the hasNext idiom here. If something checked is thrown here, it would still be transformed into an `ExecutionException`, just like anything unchecked, right? Also the unchecked version uses the checked one behind the scenes, so using the checked one is slightly more efficient. Otherwise the two are interchangeable here. Is that correct?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -141,31 +154,56 @@ private ResultOrError handleNext(T result, @Nullable Throwable error) {
...
+        hasNextError = e;
+        return true;
+      }
     }
 
     @Override
     public CompletableFuture<T> next() {
+      if (closed) {
+        throw new NoSuchElementException();
+      }
```
dimitarndimitrov, line 195: How is this handled in our exception mapping? Does it result in HTTP 500?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -141,31 +154,56 @@ private ResultOrError handleNext(T result, @Nullable Throwable error) {
...
+        return false;
+      }
+      try {
+        return idleTimeLimiter.callWithTimeout(mappingIteratorInput::hasNextValue);
+      } catch (ExecutionException | UncheckedExecutionException e) {
+        hasNextError = e.getCause();
+        return true;
+      } catch (InterruptedException | MaxIdleTimeExceededException e) {
+        hasNextError = e;
+        return true;
```
dimitarndimitrov, line 187: Is returning `true` here meant to always delegate to `next()` to actually close the stream? Arguably in both cases here (the Jetty handler thread being interrupted for some reason, or our idle time exception being raised), we need to unconditionally close the stream.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -141,31 +154,56 @@ private ResultOrError handleNext(T result, @Nullable Throwable error) {
...
+        close();
+        return CompletableFutures.failedFuture(hasNextError);
+      }
       try {
-        return CompletableFuture.completedFuture(mappingIteratorInput.nextValue());
-      } catch (Throwable e) {
+        return CompletableFuture.completedFuture(
+            idleTimeLimiter.callWithTimeout(mappingIteratorInput::nextValue));
+      } catch (ExecutionException | UncheckedExecutionException e) {
+        return CompletableFutures.failedFuture(e.getCause());
```
dimitarndimitrov, line 204: Shouldn't we also close the stream here? What are the cases that should return an error without closing the stream? I was thinking about handling bad JSON, where arguably it might be OK to keep the stream open, but wouldn't that practically always be caught in `hasNext()`, where execution exceptions do close the stream (by setting `hasNextError`)? This is all pretty subtle, so can we please add generous comments around it?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -246,13 +284,17 @@ private void push(CompletableFuture<ResultOrError> result) {
...
                       sink.write(res);
-                      if (res instanceof ErrorHolder
-                          && ((ErrorHolder) res).getError().getErrorCode() == 429) {
-                        log.warn(
-                            "Connection closed as a result of rate limiting being exceeded "
-                                + "for longer than grace period.");
-                        sinkClosed = true;
-                        sink.close();
+                      if (res instanceof ErrorHolder) {
+                        if (res.asError().getError().getErrorCode() == 429) {
```
dimitarndimitrov, line 288: nit: Why not use `Status.TOO_MANY_REQUESTS.getStatusCode()`?

dimitarndimitrov, line 288: nit: Why not just start with `Error error = ((ErrorHolder) res).getError()` and avoid the redundant `asError()` method?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -141,31 +154,56 @@ private ResultOrError handleNext(T result, @Nullable Throwable error) {
   private static class InputStreamingResponse<T> extends StreamingResponse<T> {
 
     private final MappingIterator<T> mappingIteratorInput;
+    private final StreamingResponseIdleTimeLimiter idleTimeLimiter;
+
+    private Throwable hasNextError = null;
+    private boolean closed = false;
```
dimitarndimitrov, line 160: These being plain variables relies on them always being used in methods that can be called solely by the same Jetty handler thread. As this is not actually a typical thing in a multi-threaded setting, can we at least explicitly call this out in a comment?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
rigelbm: https://javadoc.scijava.org/Guava/com/google/common/util/concurrent/SimpleTimeLimiter.html#create-java.util.concurrent.ExecutorService-

> Warning: using a bounded executor may be counterproductive! If the thread pool fills up, any time callers spend waiting for a thread may count toward their time limit, and in this case the call may even time out before the target method is ever invoked.

In reality, this thread pool is limited to the number of parallel streams, which is limited to the number of request handling threads (since `callWithTimeout` will be invoked from a request handling thread). We could use the same configuration as the request handling thread pool, not sure if necessary though. I will do it if you two think it still a problem.

ehumber: " If the thread pool fills up" - then we have another problem I think - better for it to fill up than to grow beyond what we expect.

I'd strongly prefer a size bounded thread pool. @dimitarndimitrov what do you think?

ehumber:
An additional process thing - this needs linking to the jira's raised to cover the problems

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
rigelbm: You remember almost correctly, if I remember correctly :). The error that you were seeing is that there was no `!responseQueue.isClosed()` in the `while (hasNext() && !responseQueue.isClosed())` loop. Therefore the while loop was not exiting when you closed the sink before exhausting the source. We don't actually have to close the input byte-stream. That is taken care of by jetty in the onComplete of the HTTP request. We just need this while-loop to exit, which it will. The reason a removed the early close of the input byte-stream is that, in some cases, if we close the input byte-stream before flushing the output byte-stream, this will cause a pipe broken in the client, and it will refuse to read anything we have written back (that it hasn't read yet).

testing/environments/minimal/run.sh
rigelbm: Yeah, we need a better solution for updating this at release cutting. Right now we just have to remember, which we never do. I will make sure to add a change to master when pint merging.

testing/image/Dockerfile
rigelbm: Ditto.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
dimitarndimitrov: > Warning: using a bounded executor may be counterproductive! If the thread pool fills up, any time callers spend waiting for a thread may count toward their time limit, and in this case the call may even time out before the target method is ever invoked.

This assumes that the bounded executor enqueues tasks. Not the case for the suggestion above.
It's a moot point anyway, as we don't really care whether we'll count the timeout correctly, if we are past the suggested boundary.

> In reality, this thread pool is limited to the number of parallel streams, which is limited to the number of request handling threads (since callWithTimeout will be invoked from a request handling thread). We could use the same configuration as the request handling thread pool, not sure if necessary though. I will do it if you two think it still a problem.

Right, and with our config that's an even tighter bound than the one calculated with the configured idle time and the stream rate limit. I would prefer to use the min of these, but I realize this is a hotfix, so just putting something saner than `Integer.MAX_VALUE` is good by me.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -87,4 +111,50 @@ public void dispose(ExecutorService executorService) {
...
+  }
+
+  private static final class StreamingResponseIdleTimeLimiterFactory
+      implements Factory<StreamingResponseIdleTimeLimiter> {
+    private final ExecutorService executor;
+    private final Duration maxIdleTime;
+
+    @Inject
+    StreamingResponseIdleTimeLimiterFactory(
+        @StreamingResponseIdleTimeExecutor ExecutorService executor,
```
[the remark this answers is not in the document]
rigelbm: Because we want to close it at server shutdown. One can get about doing the same by keeping a reference to the executor in the `StreamingResponseIdleTimeLimiter` class, but I find it a kind of anti-pattern to have aa owning-reference to something that you don't use for anything except closing. It is, though, an alternative way to do it, which would require about the same amount the code.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
rigelbm: I think unchecked throws `UncheckedExecutionException`, and by "I think" I mean "i've seen it when doing manual tests". :)

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
rigelbm: Yes. We delegate the responsibility to `next()` so that we have the chance to push a last message to the user saying "timeout, get lost". Notice that `next()` will `close()` the stream immediately when it sees this error, and the next `hasNext` will return false.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
rigelbm: It should. Calling `next()` when `hasNext()` is false is a programming error.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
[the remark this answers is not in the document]
dimitarndimitrov: I disagree, and would have insisted more on this, have this not been a hotfix. I also think that the alternative would have been more readable, or at least easier to understand - see my rant about our usage of HK2.
And wouldn't that alternative just manage the lifecycle of that executor in the idle limiter factory? I.e. create the executor when the factory is created, dispose of it when the provided singleton limiter is disposed?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
dimitarndimitrov: Well, it's not the unchecked version of the hasNext idiom, is the subsequent wrapping in Guava, right?
I should have written just _execution exception_ because one is not a subtype of the other - what I meant is no matter what we use, we'll have any exceptions wrapped in the corresponding form of an _execution exception_, and no matter what form of an _execution exception_ is thrown, this change would have treated it the same (because both the checked and the unchecked exceptions are always handled, and always handled in the same `catch`).
So am I right to assume there is no reason to choose one over the other besides avoiding the redundant wrapping on both the Jackson side and the Guava side?

PR #960, opened 2022-01-04
KREST-3537: Fix HTTP 500 errors in Produce V3.

ehumber:
This PR attempts to address the numerous 500 errors that were returned by REST produce.  User input should never be able to cause a 500 error, and we shouldn’t return a 500 if we can predict the failure.

See krest-3537 for the manual testing done

kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
 - The SchemaRegistryClient is now Optional - it is not created if the schema registry is not enabled

kafka-rest/src/main/java/io/confluent/kafkarest/Errors.java
 - Added two new 422 error variations.  422 means that the  “server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.”
 - the messageSerialization variant takes SerializationExceptions taken when attempting to serialise the payload
 - the invalidPayload variant is thrown when the payload would be valid, but the schema registry is disabled so it is not

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
 - this alters the default url for the schema registry to be “” which disables the schema registry
 - the change in default should be OK : we have warned for several releases now that the schema registry default will be unset shortly 

kafka-rest/src/main/java/io/confluent/kafkarest/config/ConfigModule.java
 - changes inject whether the schema registry is enabled

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/RecordSerializerFacade.java
 - the constructor for a schemaRecordSerializer throws a 422 if the schema registry is disabled. As this is found by the h2k injection it gets turned into a MultiException (the first part is the 422 RestContraintViolationExcaption, the second part is an IllegalStateException) 
 - we need to turn this back into a 422. We also need to get the message from the causing exception, not the exception that is throw as otherwise the error given to the user is the MultiException and that looks a little odd record

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
I’ve turned the serializationExceptions into 422s, as as far as I can see they are caused by the data not working with the schema registry configuration
 - line 69:  if there is no schema registry configured, then format must be specified
 - line 75 : if there is no schema registry configured then we don’t need to do anything else here
 - line 119 : Serialization exceptions result in 500 errors.  I could have put an exception mapper in, but if we get a serialisation exception we’ve not thought about then we probably want to know about it, so I’ve swapped them to be 422s
 - line 139 : if the schema id isn’t known about then the parsedSchema ends up being called “int”, and would be returned in the error message.  This swaps it back to being null (ie the schema hasn’t been found based on the id/subject)
 - line 167 : getByVersion throws a RuntimeException if the subject/schemaVersion aren’t known, we need to deal with this as a bad request

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializer.java
 - shouldn’t try to serialize a record if schema registry is disabled

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/IllegalArgumentExceptionMapper.java
 - all the checkArguments throw IllegalArgumentExceptions, which previously caused 500 errors. Now they are mapped to BadRequests

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/RestConstraintViolationExceptionMapper.java
 - this doesn’t extend StatusException, and comes from rest-utils, so we need to add an exceptions mapper to avoid 500s

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
 - SerializationException was seen during testing, so catching it too

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
 - line 135:  if you send dodgy data to an already open stream even the hasNext() can throw eg a JSON mapping exception.  This bombs out of the while loop so the response is never written to the response queue.  Added in code to do this manually in a catch.
- line 171 : sending null data to a stream also caused a barf
 - Also if an exception is thrown the objects were never closed off

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/SchemaManagerImplTest.java
- Line 1036 : this caused a NPE that was causing the test to execute down an undesired code path


rigelbm:
> this alters the default url for the schema registry to be “” which disables the schema registry
Why do we need to change the default? I would argue that we can make this change backwards compatible (even if we don't have to) by not changing the default (i.e. requiring users to explicitly set schema registry url to "" to disable) and still achieve the same effect

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestContext.java
```diff
@@ -36,7 +37,7 @@
     return getProducerPool().getProducer();
   }
 
-  default SchemaRegistryClient getSchemaRegistryClient() {
+  default Optional<SchemaRegistryClient> getSchemaRegistryClient() {
```
rigelbm, line 40: We can't change this interface, or we will otherwise break `ce-kafka-rest` and `confluent-security-plugins`. Just use `@Nullable` and `null` instead of `Optional`. (This comes from someone who hates `null`)

ehumber:
> 

That's fine with me, and it does feel safer to not change the default. I think I must have misunderstood our conversation before Christmas where I thought we decided we would change it!

kafka-rest/src/main/java/io/confluent/kafkarest/backends/schemaregistry/SchemaRegistryModule.java
```diff
@@ -44,11 +51,11 @@ private SchemaRegistryClientFactory(Provider<KafkaRestContext> context) {
     }
 
     @Override
-    public SchemaRegistryClient provide() {
+    public Optional<SchemaRegistryClient> provide() {
```
rigelbm, line 54: You can transform the `nullable` to `Optional` here with `Optional.ofNullable`.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/RecordSerializerFacade.java
```diff
@@ -45,8 +47,15 @@
       Optional<RegisteredSchema> schema,
       JsonNode data,
       boolean isKey) {
+
     if (format.requiresSchema()) {
-      return schemaRecordSerializer.get().serialize(format, topicName, schema, data, isKey);
+      SchemaRecordSerializer schemaRecordSerializer;
+      try {
+        schemaRecordSerializer = schemaRecordSerializerProvider.get();
+      } catch (MultiException ise) {
```
rigelbm, line 55: We should not rely on the DI framework's exception-on-failed-create for business logic. Instead, you should:

Option 1:
1. Make `SchemaRecordSerializer` an interface with current class renamed `SchemaRecrodSerializerImpl`.
2. Create a `ThrowingSchemaRecordSerializer`
3. In the `ControllersModule`, bind `SchemaRecordSerializer` to either `SchemaRecrodSerializerImpl` or `ThrowingSchemaRecordSerializer` based on whether schema registry is configured.

Option 2:
1. In `ControllersModule`, bind `Optional<SchemaRecordSerializer>` instead of `SchemaRecordSerializer`, to either an instance or empty based on whether schema registry is configured or not.
2. Inject `Optional<SchemaRecordSerializer>`

I have preference for Option 1, but author's choice.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -55,31 +65,45 @@ public RegisteredSchema getSchema(
       Optional<String> rawSchema,
       boolean isKey) {
     // (subject|subjectNameStrategy)?, schemaId
+
+    if (!schemaRegistryEnabled && !format.isPresent()) {
```
rigelbm, line 69: If schema registry is not configured, it should always throw, regardless of `format`. Even with `format`, we still need schema registry. We are supposed to insert the schema in SR before producing the message (the message is produced to Kafka with the SR id, not with the schema).

Also, I have a preference for having a separate class implementation that just throws, and doing the if in the module, but author's choice. As a rule of thumb, if the logic is based on something we know at start up time, instead of request time, it's better to do it in the module, IMO.

rigelbm (commented):
Partial review. Thanks for the amazing PR @ehumber!!!

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerThrowing.java
```diff
@@ -0,0 +1,44 @@
...
+package io.confluent.kafkarest.controllers;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.protobuf.ByteString;
+import io.confluent.kafkarest.Errors;
+import io.confluent.kafkarest.entities.EmbeddedFormat;
+import io.confluent.kafkarest.entities.RegisteredSchema;
+import java.util.Optional;
+
+public class SchemaRecordSerializerThrowing implements SchemaRecordSerializer {
```
rigelbm, line 25: `final`

kafka-rest/src/main/java/io/confluent/kafkarest/exceptions/IllegalArgumentExceptionMapper.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package io.confluent.kafkarest.exceptions;
+
+import io.confluent.kafkarest.exceptions.v3.ErrorResponse;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+import javax.ws.rs.ext.ExceptionMapper;
+
+public class IllegalArgumentExceptionMapper implements ExceptionMapper<IllegalArgumentException> {
```
rigelbm, line 23: `final`

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -26,26 +26,32 @@
...
+      SubjectNameStrategy defaultSubjectNameStrategy,
+      @SchemaRegistryEnabledConfig Boolean schemaRegistryEnabled) {
     this.schemaRegistryClient = requireNonNull(schemaRegistryClient);
     this.defaultSubjectNameStrategy = requireNonNull(defaultSubjectNameStrategy);
+    this.schemaRegistryEnabled = schemaRegistryEnabled;
   }
 
   @Override
-  public RegisteredSchema getSchema(
+  public Optional<RegisteredSchema> getSchema(
```
rigelbm, line 54: Why is this `Optional`? If SR is off you throw. Otherwise you always return something. You never return `Optinal.empty`. Revert.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -55,31 +61,40 @@ public RegisteredSchema getSchema(
       Optional<String> rawSchema,
       boolean isKey) {
     // (subject|subjectNameStrategy)?, schemaId
+
+    if (!schemaRegistryEnabled) {
```
rigelbm, line 65: Instead of injecting `schemaRegistryEnabled`, what if you created a `ThrowingSchemaManager` and move this logic to the module. Then you can remove `schemaRegistryEnabled` altogether.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -109,12 +124,19 @@ private RegisteredSchema getSchemaFromSchemaId(
 
   private int getSchemaVersion(String subject, ParsedSchema schema) {
     try {
-      return schemaRegistryClient.getVersion(subject, schema);
+      return schemaRegistryClient.get().getVersion(subject, schema);
     } catch (IOException | RestClientException e) {
-      throw new SerializationException(
+      String schemaString;
+      if (schema.canonicalString().equals("\"int\"")) {
```
rigelbm, line 130: Can you add more detail about this? What's special about the `"int"` schema? Can you maybe write a comment about this special case?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -128,20 +150,35 @@ private RegisteredSchema getSchemaFromSchemaVersion(
...
-                        String.format(
-                            "Error when fetching schema by version. subject = %s, version = %d",
-                            actualSubject, schemaVersion)));
+    Schema schema;
+    try {
+      schema =
+          schemaRegistryClient
+              .get()
+              .getByVersion(actualSubject, schemaVersion, /* lookupDeletedSchema= */ false);
+    } catch (RuntimeException restClientException) {
```
rigelbm, line 159: nit: `e`

We usually only name exceptions when we have multiple in the same context. Otherwise, the idiomatic is just to use `e`.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -128,20 +150,35 @@ private RegisteredSchema getSchemaFromSchemaVersion(
...
+          String.format(
+              "Schema does not exist for subject: %s, version: %s", actualSubject, schemaVersion),
+          restClientException);
+    }
+
+    ParsedSchema parsedSchema;
+    try {
+      parsedSchema =
+          EmbeddedFormat.forSchemaType(schema.getSchemaType())
+              .getSchemaProvider() // throws unsupportedOperationException for bad config
```
rigelbm, line 170: We should minimize the context of the try-block. Something like:

```
SchemaProvider schemaProvider;
try {
  schemaProvider = ...;
} catch (...) { ... }

ParsedSchema parsedSchema = schemaProvider.parseSchema(...)...
```

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -156,16 +193,21 @@ private RegisteredSchema getSchemaFromRawSchema(
...
-                () ->
-                    Errors.invalidSchemaException(
-                        String.format(
-                            "Error when parsing raw schema. format = %s, schema = %s",
-                            format, rawSchema)));
+    ParsedSchema schema;
+    try {
+      schema =
+          format
+              .getSchemaProvider() // throws UnsupportedOperationException for bad config
```
rigelbm, line 200: Same thing about minimizing scope of try-block.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -205,23 +247,31 @@ private RegisteredSchema findLatestSchema(
...
-                () ->
-                    Errors.invalidSchemaException(
-                        String.format(
-                            "Error when fetching latest schema version. subject = %s",
-                            actualSubject)));
+    ParsedSchema schema;
+    try {
+      schema =
+          EmbeddedFormat.forSchemaType(metadata.getSchemaType())
+              .getSchemaProvider() // throws UnsupportedOperationException for bad config
```
rigelbm, line 261: Same thing about minimizing the scope of try-block.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializer.java
```diff
@@ -15,156 +15,18 @@
...
-      SchemaRegistryClient schemaRegistryClient,
-      @AvroSerializerConfigs Map<String, Object> avroSerializerConfigs,
-      @JsonschemaSerializerConfigs Map<String, Object> jsonschemaSerializerConfigs,
-      @ProtobufSerializerConfigs Map<String, Object> protobufSerializerConfigs) {
-    avroSerializer = new AvroSerializer(schemaRegistryClient, avroSerializerConfigs);
-    jsonschemaSerializer =
-        new JsonSchemaSerializer(schemaRegistryClient, jsonschemaSerializerConfigs);
-    protobufSerializer = new ProtobufSerializer(schemaRegistryClient, protobufSerializerConfigs);
-  }
+public interface SchemaRecordSerializer {
```
rigelbm, line 24: Does this need to be public?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
```diff
@@ -0,0 +1,178 @@
+/*
```
rigelbm, line 1: Has there been changes in this that I need to review. With the move from `SchemaRecordSerializer` to `SchemaRecordSerializerImpl`, everything appears as new, so it's hard to see changes.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -145,7 +168,10 @@ private ResultOrError handleNext(T result, @Nullable Throwable error) {
     private InputStreamingResponse(
         MappingIterator<T> mappingIteratorInput, ChunkedOutputFactory chunkedOutputFactory) {
       super(chunkedOutputFactory);
-      this.mappingIteratorInput = requireNonNull(mappingIteratorInput);
+      if (mappingIteratorInput == null) {
+        throw new BadRequestException("Null input provided. Data is required.");
+      }
```
rigelbm, line 173: Is this possible? That's surprising. I think it would be better to check it in the `ProduceAction` before calling the `StreamingResponse`. The correct exception for `null` is NPE, like it was before.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerThrowing.java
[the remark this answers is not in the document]
ehumber: I remember you saying that - my thought was that if they all start with the same word then they are next to each other in the project tree and you can see them all more easily, and that shouldn't be the way you check for implementing classes.  But no strong feelings so let me know if you want me to change this over.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
```diff
@@ -0,0 +1,178 @@
...
+import io.confluent.kafkarest.entities.RegisteredSchema;
+import io.confluent.kafkarest.exceptions.BadRequestException;
+import java.io.IOException;
+import java.util.Map;
+import java.util.Optional;
+import javax.inject.Inject;
+import org.apache.avro.AvroTypeException;
+import org.everit.json.schema.ValidationException;
+
+public final class SchemaRecordSerializerImpl implements SchemaRecordSerializer {
```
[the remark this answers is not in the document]
ehumber: SchemaRegistryModule complains unless this is public unfortunately. I can't think of an obvious way to get around this, let me know if you can.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
```diff
@@ -0,0 +1,178 @@
...
+
+  private final AvroSerializer avroSerializer;
+  private final JsonSchemaSerializer jsonschemaSerializer;
+  private final ProtobufSerializer protobufSerializer;
+
+  @Inject
+  public SchemaRecordSerializerImpl(
```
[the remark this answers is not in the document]
ehumber: As above

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerThrowing.java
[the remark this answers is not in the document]
ehumber: As above

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
ehumber: Yup, easy to hit:

 curl -X POST  -H "Content-Type: application/json"   "http://localhost:8082/v3/clusters/lkc-xrn7x1/topics/bob/records"  
{"error_code":400,"message":"Bad Request: Null input provided. Data is required."}%   

I don't think it should throw a NPE, that would cause a 500 error.  We just need to tell the user that they need to send us some data.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
ehumber: image

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProduceActionTest.java
```diff
@@ -649,18 +750,26 @@ private static ProduceAction getProduceAction(
...
       ProduceRateLimiters produceRateLimiters,
       ChunkedOutputFactory chunkedOutputFactory,
       int times,
-      int producerId) {
+      int producerId,
+      boolean errorSchemaRegistry) {
     Provider<SchemaManager> schemaManagerProvider = mock(Provider.class);
+    SubjectNameStrategy subjectNameStrategyMock = mock(SubjectNameStrategy.class);
+    SchemaManager schemaManager =
+        new SchemaManagerImpl(Optional.empty(), subjectNameStrategyMock, false);
```
dimitarndimitrov, line 766: Just out of curiosity, can't this be a mock (which will allow `SchemaManagerImpl` to have default visibility instead of `public`? Not pushing at all for changing it though, no matter if a mock would work or not.

kafka-rest/src/test/java/io/confluent/kafkarest/unit/DefaultKafkaRestContextTest.java
```diff
@@ -46,4 +43,22 @@ public void testGetKafkaConsumerManagerThreadSafety() throws InterruptedExceptio
...
     assertEquals(1, refs.size());
   }
+
+  @Test
+  public void testGetSchemaRegistryClientEnabled() {
+    Properties props = new Properties();
+    props.put(KafkaRestConfig.SCHEMA_REGISTRY_URL_CONFIG, "localhost:5234");
+    KafkaRestConfig restConfig = new KafkaRestConfig(props);
+    context = new DefaultKafkaRestContext(restConfig);
+    assertTrue(context.getSchemaRegistryClient() != null);
```
dimitarndimitrov, line 53: nit: You can also use `assertNotNull` which might be slightly easier to read.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -26,26 +26,32 @@
...
   @Inject
-  SchemaManagerImpl(
-      SchemaRegistryClient schemaRegistryClient, SubjectNameStrategy defaultSubjectNameStrategy) {
+  public SchemaManagerImpl(
+      Optional<SchemaRegistryClient> schemaRegistryClient,
+      SubjectNameStrategy defaultSubjectNameStrategy,
+      @SchemaRegistryEnabledConfig Boolean schemaRegistryEnabled) {
     this.schemaRegistryClient = requireNonNull(schemaRegistryClient);
     this.defaultSubjectNameStrategy = requireNonNull(defaultSubjectNameStrategy);
+    this.schemaRegistryEnabled = schemaRegistryEnabled;
```
dimitarndimitrov, line 50: Is there a case where SR is disabled but the SR client is not null, or SR is enabled but the SR client is null?
If not, we should be able to get away without the `Optional` decorator for the SR client - should we validate that?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -109,12 +124,19 @@ private RegisteredSchema getSchemaFromSchemaId(
...
-      return schemaRegistryClient.getVersion(subject, schema);
+      return schemaRegistryClient.get().getVersion(subject, schema);
     } catch (IOException | RestClientException e) {
-      throw new SerializationException(
+      String schemaString;
+      if (schema.canonicalString().equals("\"int\"")) {
+        schemaString = "";
+      } else {
+        schemaString = schema.canonicalString();
+      }
```
dimitarndimitrov, line 134: nit: I think this fits on a single line with a ternary op:
```
String schemaString = schema.canonicalString().equals("\"int\"") ? "" : schema.canonicalString();
```

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
```diff
@@ -0,0 +1,178 @@
...
+
+    if (schemaRegistryClient.isPresent()) {
+      avroSerializer = new AvroSerializer(schemaRegistryClient.get(), avroSerializerConfigs);
+      jsonschemaSerializer =
+          new JsonSchemaSerializer(schemaRegistryClient.get(), jsonschemaSerializerConfigs);
+      protobufSerializer =
+          new ProtobufSerializer(schemaRegistryClient.get(), protobufSerializerConfigs);
+    } else {
+      throw Errors.messageSerializationException(
+          "Schema registry not defined, no Schema Registry client "
```
dimitarndimitrov, line 69: nit: Schema registry -> Schema Registry

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerThrowing.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public SchemaRecordSerializerThrowing() {
+
+    throw Errors.messageSerializationException(
+        "Schema registry not defined, no Schema "
+            + "Registry client available to serialize message.");
```
dimitarndimitrov, line 31: nit: Redundant blank line. Also why do we need to throw in the ctor here?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -79,6 +84,16 @@
...
+          .putMapper(
+              RestConstraintViolationException.class,
+              new RestConstraintViolationExceptionMapper(),
+              response -> ((ErrorResponse) response.getEntity()).getErrorCode(),
+              response -> ((ErrorResponse) response.getEntity()).getMessage())
+          .putMapper(
+              IllegalArgumentException.class,
+              new IllegalArgumentExceptionMapper(),
+              response -> Status.BAD_REQUEST.getStatusCode(),
+              response -> ((ErrorResponse) response.getEntity()).getMessage())
```
dimitarndimitrov, line 96: This looks way too generic - can't we handle it more specifically? There's a chance we might end up masking some genuine programming errors as bad requests. I know we're manually using the mapper, but it still looks like too wide of a net.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImplTest.java
```diff
@@ -0,0 +1,35 @@
+package io.confluent.kafkarest.controllers;
```
dimitarndimitrov, line 1: nit: Missing copyright header.

dimitarndimitrov (commented):
First pass on everything but `SchemaManagerImplTest`.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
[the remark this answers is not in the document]
rigelbm: That's because things in the `controllers` package should be configured in the `ControllersModule`, not on `SchemaRegistryModule`. :)

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
ehumber: (I moved this to the produce action though, as suggested)

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
ehumber: I've taken it out - I can't get it to break without the mapper in place, as we catch the IllegalArgumentException in ProduceAction anyway, around the getSchema call.

It could be argued that even there the catch is too generic, but the alternative is to put a try catch around each/all of the checkArugments in getSchema, and that feels a bit long winded.

I think not having it in the exception mapper, and having it around the getSchema call in the produceAction is probably the right balance.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -869,28 +870,28 @@ public Properties getOriginalProperties() {
           "Using default value {} for config {}. In a future release this config won't have a"
               + "default value anymore. If you are using Schema Registry, please, specify {}"
               + "explicitly. Requests will fail in a future release if you try to use Schema "
-              + "Registry but have not specified a value for {}.",
+              + "Registry but have not specified a value for {}. An empty value for this property "
+              + "means that the Schema Registry is disabled.",
           SCHEMA_REGISTRY_URL_DEFAULT,
           SCHEMA_REGISTRY_URL_CONFIG,
-          SCHEMA_REGISTRY_URL_CONFIG,
```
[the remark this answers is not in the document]
ehumber: That's a really good spot Dimitar.  

It had been removed as pre-review comment change there was one fewer {} to populate.  That's not the case any more, so I've put it back in.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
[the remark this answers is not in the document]
rigelbm: As above.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
rigelbm: We should trust `google-java-format`.

dimitarndimitrov: Are you saying `google-java-format` will break the proposed one-liner? I would expect it won't if it fits within our line length limits, which this does.

rigelbm: Oh, sorry. I read too fast. I thought you just reformatted the code. Ignore me. It's Friday. :)

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
ehumber: I think I retract this with another look at the test case that I thought caused it 😊

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
ehumber: OK, but then we need two try blocks instead - both getSchemaProvider and parseSchema throw exceptions we want to catch - the orElseThrow just throws if the value provided to it is null if I understand the javadoc correctly.

It seems tidier to have the catch together.  

No strong feelings though, let me know if you want me to split it out to be what you have above, but with a try catch around the parseSchema line too.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
ehumber: As above

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializer.java
ehumber: Not after some of the other changes :)

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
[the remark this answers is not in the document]
ehumber: Moved back - I'd put it there originally but changed my mind for some reason!

ehumber:
@rigelbm @dimitarndimitrov Thanks for all the review comments, they've helped tidy this up a lot :)

I think I've either fixed them or commented, let me know if I missed any and any further comments.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/RecordSerializerFacadeTest.java
```diff
@@ -95,13 +96,43 @@ public void setUp() {
...
 
+  @Test
+  public void noSchemaRegistryClientConfigured() {
+
+    RecordSerializerFacade myRecordSerializer =
+        new RecordSerializerFacade(
+            new NoSchemaRecordSerializer(SCHEMA_SERIALIZER_CONFIGS),
+            () -> new SchemaRecordSerializerThrowing());
+
+    boolean checkpoint = false;
```
rigelbm, line 114: The idiomatic way of doing this with JUnit is:

```
RestConstraintViolationException exception =
    assertThrows(
        RestConstraintViolationException.class,
        () ->
            myRecordSerializer.serialize(
                EmbeddedFormat.AVRO,
                TOPIC_NAME,
                /* schema= */ Optional.empty(),
                TextNode.valueOf(
                    BaseEncoding.base64().encode("foobar".getBytes(StandardCharsets.UTF_8))),
                /* isKey= */ true));

assertEquals(
    "Error serializing message. Schema Registry not defined, no Schema Registry client available to serialize message.",
    exception.getMessage());
assertEquals(42207, exception.getErrorCode());
```

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class SchemaRecordSerializerTest {
+
+  @Test
+  public void errorWhenNoSchemaRegistryDefined() {
+    boolean checkpoint = false;
```
rigelbm, line 33: Same.

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProduceActionTest.java
```diff
@@ -482,16 +543,60 @@ public void produceWithByteLimit() throws Exception {
...
+    // setup
+    ChunkedOutputFactory chunkedOutputFactory = mock(ChunkedOutputFactory.class);
+
+    Clock clock = mock(Clock.class);
+    ProduceAction produceAction = getProduceAction(properties, chunkedOutputFactory, clock, 1);
+    MappingIterator<ProduceRequest> requests = null;
+
+    FakeAsyncResponse fakeAsyncResponse = new FakeAsyncResponse();
+
+    boolean checkpoint = false;
```
rigelbm, line 565: Same.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/RecordSerializerFacadeTest.java
```diff
@@ -95,13 +96,43 @@ public void setUp() {
...
+    try {
+      myRecordSerializer
+          .serialize(
+              EmbeddedFormat.AVRO,
+              TOPIC_NAME,
+              /* schema= */ Optional.empty(),
+              TextNode.valueOf(
+                  BaseEncoding.base64().encode("foobar".getBytes(StandardCharsets.UTF_8))),
+              /* isKey= */ true)
+          .get();
```
rigelbm, line 124: You don't actually need to call `Optional.get()`, do you?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
```diff
@@ -38,9 +50,75 @@ protected void configure() {
...
+    public SchemaRecordSerializer provide() {
+      if (schemaRegistryClient.isPresent()) {
+        return new SchemaRecordSerializerImpl(
+            schemaRegistryClient,
+            avroSerializerConfigs,
+            jsonschemaSerializerConfigs,
+            protobufSerializerConfigs);
+      } else {
+        return new SchemaRecordSerializerThrowing();
+      }
```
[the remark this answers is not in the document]
ehumber: That doesn't work unfortunately, the type returned by .orElse (SchemaRecordSerializerThrowing) needs to match the type returned by map (SchemaRecordSerializerImpl) - you can't do it on the super class as far as I can work out.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -116,11 +125,19 @@ public final void resume(AsyncResponse asyncResponse) {
...
+    try {
+      while (hasNext() && !responseQueue.isClosed()) {
+        responseQueue.push(next().handle(this::handleNext));
+      }
+    } catch (Exception e) {
+      log.debug("Exception thrown when processing stream ", e);
+      responseQueue.push(
+          CompletableFuture.completedFuture(
+              ResultOrError.error(EXCEPTION_MAPPER.toErrorResponse(e))));
+    } finally {
```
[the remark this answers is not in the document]
ehumber: My thinking is that the responseQueue.push in the catch might unexpectedly error (there is quite a lot of code in there), so it's best to always close so we don't leave anything hanging.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
```diff
@@ -38,9 +50,75 @@ protected void configure() {
...
+        SubjectNameStrategy defaultSubjectNameStrategy,
+        @SchemaRegistryEnabledConfig Boolean schemaRegistryEnabled) {
+      this.schemaRegistryClient = schemaRegistryClient;
+      this.defaultSubjectNameStrategy = requireNonNull(defaultSubjectNameStrategy);
+      this.schemaRegistryEnabled = requireNonNull(schemaRegistryEnabled);
+    }
+
+    @Override
+    public SchemaManager provide() {
+      if (schemaRegistryEnabled) {
```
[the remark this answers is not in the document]
ehumber: Yes to the schemaRegistry is there check, but no to the Optional.map due to the types not matching

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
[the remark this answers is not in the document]
rigelbm: You can force the types:

```
myOptional.<Foo>map(x -> subFoo(x)).orElse(() -> anotherSubFoo());
```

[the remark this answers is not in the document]
ehumber: Honestly, I think the if else is cleaner and nicer to read.

The really way to do it is a .ifPresentOrElse() but we don't have that until a later Java version.

ehumber:
@rigelbm @dimitarndimitrov ready to go again...

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -995,6 +997,10 @@ public final boolean isRateLimitEnabled() {
     return getBoolean(RATE_LIMIT_ENABLE_CONFIG);
   }
 
+  public final boolean isSchemaRegistryEnabled() {
+    return !getSchemaRegistryConfigs().get(SCHEMA_REGISTRY_URL_CONFIG).equals("");
+  }
+
```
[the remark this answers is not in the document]
ehumber: I need this utility method, or I need to call `getSchemaRegistryConfigs().get(SCHEMA_REGISTRY_URL_CONFIG).equals("")` in DefaultKafkaRestContext (which is the one place that calls the utility method).  I like the utility method better.

the getSchemaRegistryConfigs needs calling as that's what populates the url if defaults are needed, there is no `schemaRegistryUrl.isEmpty()`

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
```diff
@@ -38,9 +50,75 @@ protected void configure() {
...
+
+  private static final class SchemaManagerFactory implements Factory<SchemaManager> {
+
+    private final SchemaRegistryClient schemaRegistryClient;
+    private final SubjectNameStrategy defaultSubjectNameStrategy;
+    private final boolean schemaRegistryEnabled;
+
+    @Inject
+    private SchemaManagerFactory(
+        SchemaRegistryClient schemaRegistryClient,
```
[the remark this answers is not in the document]
rigelbm: You should never bind `null`. A binding of `null` should be considered a bug. If the binding is optional, you should mark it with `Optional` for clarity.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
```diff
@@ -38,9 +49,71 @@ protected void configure() {
...
+  }
+
+  private static final class SchemaManagerFactory implements Factory<SchemaManager> {
+
+    private final SchemaRegistryClient schemaRegistryClient;
+    private final SubjectNameStrategy defaultSubjectNameStrategy;
+
+    @Inject
+    private SchemaManagerFactory(
+        SchemaRegistryClient schemaRegistryClient, SubjectNameStrategy defaultSubjectNameStrategy) {
```
rigelbm, line 102: I still think you should inject `Optional<SchemaRegistryClient>`.

ehumber:
@rigelbm Ready for you again :)

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
```diff
@@ -38,9 +49,72 @@ protected void configure() {
...
+  private static final class SchemaManagerFactory implements Factory<SchemaManager> {
+
+    private final Optional<SchemaRegistryClient> schemaRegistryClient;
+    private final SubjectNameStrategy defaultSubjectNameStrategy;
+
+    @Inject
+    private SchemaManagerFactory(
+        Optional<SchemaRegistryClient> schemaRegistryClient,
+        SubjectNameStrategy defaultSubjectNameStrategy) {
+      this.schemaRegistryClient = schemaRegistryClient;
```
rigelbm, line 104: requireNonNull

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/RecordSerializerFacade.java
```diff
@@ -28,14 +28,14 @@
 final class RecordSerializerFacade implements RecordSerializer {
 
   private final NoSchemaRecordSerializer noSchemaRecordSerializer;
-  private final Provider<SchemaRecordSerializer> schemaRecordSerializer;
+  private final Provider<SchemaRecordSerializer> schemaRecordSerializerProvider;
```
rigelbm, line 31: You can completely revert this file, right? There's only the attribute rename now, which is not needed anymore.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -128,20 +129,42 @@ private RegisteredSchema getSchemaFromSchemaVersion(
...
+          schemaRegistryClient.getByVersion(
+              actualSubject, schemaVersion, /* lookupDeletedSchema= */ false);
+    } catch (RuntimeException e) {
+      throw new BadRequestException(
+          String.format(
+              "Schema does not exist for subject: %s, version: %s", actualSubject, schemaVersion),
+          e);
+    }
+
+    ParsedSchema parsedSchema;
```
rigelbm, line 144: Move variable definition closer to usage.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -154,18 +177,36 @@ private RegisteredSchema getSchemaFromRawSchema(
...
-                        String.format(
-                            "Error when parsing raw schema. format = %s, schema = %s",
-                            format, rawSchema)));
+    try {
+      checkArgument(format.requiresSchema(), "%s does not support schemas.", format);
+    } catch (IllegalArgumentException e) {
+      throw new BadRequestException(e.getMessage(), e);
+    }
+
+    ParsedSchema schema;
```
rigelbm, line 186: Move variable definition closer to usage.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -207,21 +248,37 @@ private RegisteredSchema findLatestSchema(
...
-        EmbeddedFormat.forSchemaType(metadata.getSchemaType())
-            .getSchemaProvider()
-            .parseSchema(metadata.getSchema(), metadata.getReferences(), /* isNew= */ false)
-            .orElseThrow(
-                () ->
-                    Errors.invalidSchemaException(
-                        String.format(
-                            "Error when fetching latest schema version. subject = %s",
-                            actualSubject)));
+    ParsedSchema schema;
```
rigelbm, line 256: Move variable definition closer to usage.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -264,4 +321,12 @@ private String getSchemaSubjectUnsafe(
 
     return subject;
   }
+
+  private static void checkArgumentWrapper(boolean argument) {
+    try {
+      checkArgument(argument);
+    } catch (IllegalArgumentException e) {
+      throw new BadRequestException("Unsupported argument: ", e.getMessage(), e);
+    }
```
rigelbm, line 330: Just reimplement it, i.e.:

```
if (!argument) {
  throw new BadRequestException(...);
}
```

Also, allow for messages, for us to get more descriptive errors, i.e.:

```
checkArgumentWrapper(!format.isPresent(), "`format` shouldn't be present because X.");
```

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -118,6 +120,10 @@ public void produce(
       MappingIterator<ProduceRequest> requests)
       throws Exception {
 
+    if (requests == null) {
+      throw Errors.invalidPayloadException("Null input provided. Data is required.");
+    }
```
rigelbm, line 125: I'm guessing `null` means a completely empty stream. Would the most consistent way to handle this be just to return empty instead?

rigelbm (approved):
Just a couple of nits. LGTM otherwise. Thanks!!!

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
ehumber: That's odd, I have that locally.  Will do a dummy commit and repush

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/RecordSerializerFacade.java
ehumber: True but it is a provider, so it doesn't really matter.  leaving.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
ehumber: Follow on - this PR could do with merging :) I'll raise an issue.  

This is no worse than it was before wrt to the messages but I do agree they should be better.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/SchemaManagerImplTest.java
```diff
@@ -383,101 +397,147 @@ public void getSchema_avro_latestSchema_notIsKey() throws Exception {
...
-            schemaManager.getSchema(
-                TOPIC_NAME,
-                /* format= */ Optional.empty(),
-                /* subject= */ Optional.empty(),
-                /* subjectNameStrategy= */ Optional.empty(),
-                /* schemaId= */ Optional.of(1000),
-                /* schemaVersion= */ Optional.empty(),
-                /* rawSchema= */ Optional.empty(),
-                /* isKey= */ true));
+    boolean checkpoint = false;
```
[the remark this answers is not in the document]
ehumber: also odd - the change is made locally.  again, I'll dummy commit and push and see if it appears.  Sorry for not spotting it in the git UI.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
ehumber: https://confluentinc.atlassian.net/browse/KREST-4069

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/SchemaManagerImplTest.java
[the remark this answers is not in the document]
ehumber: image

Both look like this in the diff for me

ehumber:
Failing test is a known flakey one - GuavaRateLimitTest.rateLimitWithDefaultCost:22->AbstractRateLimitEnabledTest.rateLimitWithDefaultCost:61->AbstractRateLimitEnabledTest.assertInRange:

Merging

PR #961, opened 2022-01-13
KREST-3322 First pass of removing grace period

ehumber:
This PR removes the grace period that was originally added for rate limiting REST produce requests, and also replaces the custom rate limiter with the Guava/Resiliance4J limiter that is currently used to limit admin requests.

Although grace period was originally included in the 1 pager for produce, with more experience of rate limiting we have determined that using a simple library limiter is a better solution.

When a streamed produce is limited, the connection is no longer closed, instead a 429 is returned and the message is not processed.

The Guava/Resiliance4J/Null high level request rate limiters are injected, with configuration taken from the overall rate limits.

For produce, the produce specific limits need to be used, so I have made new Factory classes for both bytes and count limiters (they take different values) that extend the original factory (as otherwise I can’t override the configuration values).  

The new factories call super on the original RequestRateLimiterFactory to initialise the RequestRateLimiter, and then store a local copy of the RequestRateLimiter to be used for the actual rate limiting.  Unfortunately we need to store a copy of the RequestRateLimiter, as I need to be able to inject either a produce bytes limiter or a produce count limiter but the new limiter classes can’t extend all of the Guava, Resiliance4J and Null Rate limiters (because Java). I had to make the new injectable classes just extend the RequestRateLimiter and override the rateLimit method to call the underlying saved RateLimiters limit method.  Convoluted, yes. Sorry.  Open to any suggestions to make it simpler!

The cost of a counter based limit is 1, the cost of a bytes based limit is the size of the message in bytes.

Produce action had some code that I didn’t understand (and I think was wrong), where it didn’t rate limit at all if there was no message size available - I’ve taken that out so that rate limiting always happens, but if there is no message size, the bytes based limiting is not done.  Looking through the code, the message size is always there (added in the deserialise method of the ProduceRequestSerdeModule).  

The alternative is the we barf if the size is missing - what do you think?

In the ProduceRateLimiters, I quick return if rate limiting is disabled - this feels more efficient than allowing it to fall through, for both bytes and count, to the null limiter to return on.

I’ve taken out a number of the ProduceActionTest unit tests because they were actually testing the rate limiter, not produce action, and that’s either now done in ProduceRateLimiters or is unnecessarily checking the behaviour of the external libraries.


kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytes.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.ratelimit;
+
+public class RequestRateLimiterProduceBytes extends RequestRateLimiter {
```
rigelbm, line 18: final

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytes.java
```diff
@@ -0,0 +1,34 @@
...
+
+  RequestRateLimiter requestRateLimiter;
```
rigelbm, line 20: private final

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytes.java
```diff
@@ -0,0 +1,34 @@
...
+
+  RequestRateLimiterProduceBytes(RequestRateLimiter requestRateLimiter) {
+    this.requestRateLimiter = requestRateLimiter;
```
rigelbm, line 23: requireNonNull

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytesFactory.java
```diff
@@ -0,0 +1,38 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.ratelimit;
+
+import io.confluent.kafkarest.config.ConfigModule;
+import java.time.Duration;
+import javax.inject.Inject;
+
+public class RequestRateLimiterProduceBytesFactory extends RequestRateLimiterFactory {
```
rigelbm, line 22: final

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytesFactory.java
```diff
@@ -0,0 +1,38 @@
...
+
+  RequestRateLimiterProduceBytes requestRateLimiterProduceBytes;
```
rigelbm, line 24: private final

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytesFactory.java
```diff
@@ -0,0 +1,38 @@
...
+  @Inject
+  public RequestRateLimiterProduceBytesFactory(
+      RateLimitBackend backend,
+      @ConfigModule.ProduceRateLimitBytesConfig Integer permitsPerSecond,
+      @ConfigModule.RateLimitTimeoutConfig Duration timeout) {
+    super(backend, permitsPerSecond, timeout);
+    requestRateLimiterProduceBytes = new RequestRateLimiterProduceBytes(super.provide());
+  }
+
+  public RequestRateLimiter provide() {
```
rigelbm, line 35: `@override`

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytesFactory.java
```diff
@@ -0,0 +1,38 @@
...
+public class RequestRateLimiterProduceBytesFactory extends RequestRateLimiterFactory {
+
+  RequestRateLimiterProduceBytes requestRateLimiterProduceBytes;
+
+  @Inject
+  public RequestRateLimiterProduceBytesFactory(
+      RateLimitBackend backend,
+      @ConfigModule.ProduceRateLimitBytesConfig Integer permitsPerSecond,
+      @ConfigModule.RateLimitTimeoutConfig Duration timeout) {
+    super(backend, permitsPerSecond, timeout);
```
rigelbm, line 31: Try instantiating `RequestRateLimiterFactory` directly here.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytesFactory.java
```diff
@@ -0,0 +1,38 @@
...
+    requestRateLimiterProduceBytes = new RequestRateLimiterProduceBytes(super.provide());
```
rigelbm, line 32: Don't cache the provided instance. Make sure to return a new instance everytime. HK2 will take care of the lifetime of the instance.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceCount.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.ratelimit;
+
+public class RequestRateLimiterProduceCount extends RequestRateLimiter {
```
rigelbm, line 18: final

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceCount.java
```diff
@@ -0,0 +1,34 @@
...
+
+  RequestRateLimiter requestRateLimiter;
```
rigelbm, line 20: private final

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceCount.java
```diff
@@ -0,0 +1,34 @@
...
+
+  RequestRateLimiterProduceCount(RequestRateLimiter requestRateLimiter) {
+    this.requestRateLimiter = requestRateLimiter;
```
rigelbm, line 23: requireNonNull

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceCountFactory.java
```diff
@@ -0,0 +1,38 @@
...
+      RateLimitBackend backend,
+      @ConfigModule.ProduceRateLimitCountConfig Integer permitsPerSecond,
+      @ConfigModule.RateLimitTimeoutConfig Duration timeout) {
+    super(backend, permitsPerSecond, timeout);
+    requestRateLimiterProduceCount = new RequestRateLimiterProduceCount(super.provide());
+  }
+
+  public RequestRateLimiter provide() {
+    return requestRateLimiterProduceCount;
+  }
```
rigelbm, line 37: Same as the other factory.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
```diff
@@ -20,66 +20,79 @@
...
 
-  private final int maxRequestsPerSecond;
-  private final int maxBytesPerSecond;
-  private final Duration gracePeriod;
   private final boolean rateLimitingEnabled;
-  private final LoadingCache<String, ProduceRateLimiter> cache;
+  private final LoadingCache<String, RequestRateLimiter> countCache;
+  private final LoadingCache<String, RequestRateLimiter> bytesCache;
+  Provider<RequestRateLimiterProduceCount> countLimiterProvider;
+  Provider<RequestRateLimiterProduceBytes> bytesLimiterProvider;
```
rigelbm, line 42: private final

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RateLimitModule.java
```diff
@@ -26,5 +26,9 @@
   @Override
   protected void configure() {
     bindFactory(RequestRateLimiterFactory.class).to(RequestRateLimiter.class).in(Singleton.class);
+    bindFactory(RequestRateLimiterProduceBytesFactory.class)
+        .to(RequestRateLimiterProduceBytes.class);
+    bindFactory(RequestRateLimiterProduceCountFactory.class)
+        .to(RequestRateLimiterProduceCount.class);
```
rigelbm, line 32: Should be maybe marked explicitly as PerLookup.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
```diff
@@ -20,66 +20,79 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(ProduceRateLimiters.class);
```
rigelbm, line 44: very optional: logger should be the first member in the class.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
```diff
@@ -20,66 +20,79 @@
...
-      @ProduceRateLimitCacheExpiryConfig Duration produceRateLimitCacheExpiryConfig,
-      Clock time) {
-    this.maxRequestsPerSecond = requireNonNull(produceRateLimitCountConfig);
-    this.gracePeriod = requireNonNull(produceGracePeriodConfig);
+      @ProduceRateLimitCacheExpiryConfig Duration produceRateLimitCacheExpiryConfig) {
     this.rateLimitingEnabled = requireNonNull(produceRateLimitEnabledConfig);
-    this.maxBytesPerSecond = requireNonNull(produceRateLimitBytesConfig);
-    requireNonNull(time);
+    this.countLimiterProvider = countLimiterProvider;
+    this.bytesLimiterProvider = bytesLimiterProvider;
```
rigelbm, line 54: requireNonNull

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
```diff
@@ -20,66 +20,79 @@
...
                   @Override
-                  public ProduceRateLimiter load(String key) {
-                    return new ProduceRateLimiter(
-                        gracePeriod,
-                        maxRequestsPerSecond,
-                        maxBytesPerSecond,
-                        produceRateLimitEnabledConfig,
-                        time);
+                  public RequestRateLimiter load(String key) {
+                    return countLimiterProvider.get().getLimiter();
```
rigelbm, line 63: You don't need `getLimiter`, and this class does not need to be `public`. Use binding annotations instead to differentiate between all the `RequestRateLimiters`.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
```diff
@@ -20,66 +20,79 @@
...
-    ProduceRateLimiter rateLimiter = cache.getUnchecked(clusterId);
-    Optional<Duration> waitTime = rateLimiter.calculateGracePeriodExceeded(requestSize);
-    return waitTime;
+    // requestSize should always be present, but if there is no message size available
+    // we allow the request through without limiting it
   }
 
   public void clear() {
-    cache.invalidateAll();
+    countCache.invalidateAll();
```
rigelbm, line 96: "Why is the byte one missing?", @ehumber

rigelbm (commented):
Niece PR! Thanks!

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
ehumber: N/A

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceCount.java
ehumber: N/A

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceCount.java
ehumber: N/A

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytesFactory.java
ehumber: Went with abstracting out the factory and extending it, as discussed.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytesFactory.java
ehumber: N/A

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytesFactory.java
ehumber: N/A

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytes.java
ehumber: N/A

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RequestRateLimiterProduceBytes.java
ehumber: N/A

ehumber:
@rigelbm Ready for round 2 :)

rigelbm (approved):
That's a clear improvement in the way we rate limit produce. Nice! Thanks!

PR #963, opened 2022-01-19
KREST-4113 Don't delete all ACLs if an empty DELETE request is received.

dimitarndimitrov:
This change prevents the simplest DELETE request (a no-parameter one) from deleting all ACLs. This is done by a simple change of the default value for the resource_type query param from a wildcard match to an invalid one (one that requires explicit input from the user).
- In addition to that, some inconsistencies in the delete request validation have been addressed.

For some reason, requests with partially filled query params are not currently hitting the "Delete ACL" endpoint - it should be investigated whether that's due to a testing problem, Jersey bug, or something else.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AclsResource.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AclsResource.java
@@ -212,25 +212,25 @@ public final class AclsResource {
   public void deleteAcls(
       @Suspended AsyncResponse asyncResponse,
       @PathParam("clusterId") String clusterId,
-      @QueryParam("resource_type") @DefaultValue("any") ResourceType resourceType,
+      @QueryParam("resource_type") @DefaultValue("unknown") ResourceType resourceType,
       @QueryParam("resource_name") @DefaultValue("") String resourceName,
-      @QueryParam("pattern_type") @DefaultValue("any") PatternType patternType,
+      @QueryParam("pattern_type") @DefaultValue("unknown") PatternType patternType,
       @QueryParam("principal") @DefaultValue("") String principal,
       @QueryParam("host") @DefaultValue("") String host,
-      @QueryParam("operation") @DefaultValue("any") Operation operation,
-      @QueryParam("permission") @DefaultValue("any") Permission permission
+      @QueryParam("operation") @DefaultValue("unknown") Operation operation,
+      @QueryParam("permission") @DefaultValue("unknown") Permission permission
   ) {
     if (resourceType == Acl.ResourceType.UNKNOWN) {
-      throw new BadRequestException("resource_type cannot be ANY");
+      throw new BadRequestException("resource_type cannot be unspecified or UNKNOWN");
     }
     if (patternType == Acl.PatternType.UNKNOWN) {
-      throw new BadRequestException("pattern_type cannot be UNKNOWN");
+      throw new BadRequestException("pattern_type cannot be unspecified or UNKNOWN");
     }
     if (operation == Acl.Operation.UNKNOWN) {
-      throw new BadRequestException("operation cannot be ANY");
+      throw new BadRequestException("operation cannot be unspecified or UNKNOWN");
     }
     if (permission == Acl.Permission.UNKNOWN) {
-      throw new BadRequestException("permission cannot be ANY");
+      throw new BadRequestException("permission cannot be unspecified or UNKNOWN");
     }
 
     CompletableFuture<DeleteAclsResponse> response =
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/AclsResourceIntegrationTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/AclsResourceIntegrationTest.java
@@ -19,6 +19,7 @@ import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import io.confluent.kafkarest.entities.Acl;
 import io.confluent.kafkarest.entities.v3.AclData;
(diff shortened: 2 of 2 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AclsResource.java
```diff
@@ -212,7 +212,7 @@ public void createAcl(
   public void deleteAcls(
       @Suspended AsyncResponse asyncResponse,
       @PathParam("clusterId") String clusterId,
-      @QueryParam("resource_type") @DefaultValue("any") ResourceType resourceType,
+      @QueryParam("resource_type") @DefaultValue("unknown") ResourceType resourceType,
```
rigelbm, line 215: Question 1: Should we also change `patternType`, `operation` and `permission` to `unknown`, for consistency?

Question 2: Is this backwards compatible? I don't think so. We might be forced to just update the documentation to explain it better. :(

rigelbm: In the Java API, you have to explicitly set all fields. There are no defaults. The defaults here are mostly for convenience, I believe. For the read operation, that's fine, I guess. But for this one here, we probably want to force something to be specified.

Now, another option to just forcing `resourceType`, or forcing everything, is to just force something. i.e.: all the defaults stay the same, but you have to set at least one field, anyone will do. That eliminates the empty == all case.

As for backwards compatibility, you are correct. I think it's fine to change the defaults if they aren't documented.

dimitarndimitrov: > The defaults here are mostly for convenience, I believe.

What I meant was mostly related to this:
_For some reason, requests with partially filled query params are not currently hitting the "Delete ACL" endpoint - it should be investigated whether that's due to a testing problem, Jersey bug, or something else._
My original desire was to add to the "Delete ACL" integration tests a test that does the following:
1. Sets up a couple ACLs.
2. Tries to delete all of them with empty query params and fails.
3. Tries to delete all of them by just specifying `?resource_type=all` and omitting the rest of the query params.

For some reason, while 2. worked as expected (and still is part of the test that we have in this change), 3. did not hit the "Delete ACL" endpoint. This was pretty surprising to me, but I did not spend too much time investigating it, as I still want to focus on other things this Sprint. Only when I specified all other params did I hit the endpoint. At that point I decided to just reuse the common search criteria.

> Now, another option to just forcing resourceType, or forcing everything, is to just force something. i.e.: all the defaults stay the same, but you have to set at least one field, anyone will do.

I am not sure if this is better, but more importantly I am not sure how can we implement it. I.e. how do we differentiate between all the parameters having the default values because of an empty request vs because the client has set exactly a single parameter to its default value?

dimitarndimitrov:
Apparently this change exploded the cyclomatic and NPath complexity of `deleteAcls` and now Checkstyle doesn't like it. I have to check in more depth how without any complexity besides the validation, the more complex validation of `createAcl` doesn't trigger Checkstyle. I might very well add a suppression here, because this looks like an overreaction from Checkstyle.

dimitarndimitrov:
In the end I went back close to the initial version of this change, as removing the default values altogether necessitates additional validation checks, and these lead to the aforementioned complexity problems. I think it's more acceptable to go this route as opposed to adding a Checkstyle exception, but am open to reconsidering if others feel different.
@rigelbm , @ehumber thoughts?

ehumber:
I'm happy - this is a nasty bug and we should fix it asap, what you've done looks good enough to me.

(Should we consider cherry picking the fix?)

ehumber (approved):

rigelbm (approved):
LGTM

dimitarndimitrov:
Thank you, @ehumber and @rigelbm - I'm not sure why did the PR CI merge validation step failed.
At the end one can see 
```
[2022-01-26T00:27:18.587Z] [INFO] ------------------------------------------------------------------------
[2022-01-26T00:27:18.587Z] [INFO] Reactor Summary for kafka-rest-parent 6.0.6-SNAPSHOT:
[2022-01-26T00:27:18.587Z] [INFO] 
[2022-01-26T00:27:18.587Z] [INFO] kafka-rest-parent .................................. SUCCESS [06:27 min]
[2022-01-26T00:27:18.587Z] [INFO] kafka-rest ......................................... SUCCESS [33:29 min]
[2022-01-26T00:27:18.587Z] [INFO] kafka-rest-console-scripts ......................... SUCCESS [  5.810 s]
[2022-01-26T00:27:18.587Z] [INFO] ------------------------------------------------------------------------
[2022-01-26T00:27:18.587Z] [INFO] BUILD SUCCESS
[2022-01-26T00:27:18.587Z] [INFO] ------------------------------------------------------------------------
[2022-01-26T00:27:18.587Z] [INFO] Total time:  40:15 min
[2022-01-26T00:27:18.587Z] [INFO] Finished at: 2022-01-26T00:27:18Z
[2022-01-26T00:27:18.587Z] [INFO] ------------------------------------------------------------------------
```
followed by some test cleanup where the most suspicious thing I could find was the following mess:
```
[2022-01-26T00:29:35.671Z] + curl -f https://dependency-mgmt.tools.confluent-internal.io/v1/downstream_repos/kafka-rest/[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginResolutionException
[2022-01-26T00:29:35.671Z]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[2022-01-26T00:29:35.671Z]                                  Dload  Upload   Total   Spent    Left  Speed
[2022-01-26T00:29:35.671Z] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[2022-01-26T00:29:35.671Z] curl: (22) The requested URL returned error: 531 
[2022-01-26T00:29:35.671Z] curl: (3) bad range in URL position 2:
[2022-01-26T00:29:35.671Z] [Help
[2022-01-26T00:29:35.671Z]  ^
[2022-01-26T00:29:35.671Z] curl: (3) unmatched close brace/bracket in URL position 2:
```

I tried rerunning the PR CI job, but got greeted with `Skipping this PR build because it was triggered by branch indexing.`, probably as a result of the recent CI changes, which will require empty commits to rerun tests easily.
I will proceed with merging this.

PR #964, opened 2022-01-21
KREST-3660 Move to Junit5

ehumber:
Move to junit 5

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/SchemaRegistrySaslInheritTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/SchemaRegistrySaslInheritTest.java
@@ -15,7 +15,7 @@
 
 package io.confluent.kafkarest.integration;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import com.fasterxml.jackson.databind.node.TextNode;
 import io.confluent.kafkarest.entities.EmbeddedFormat;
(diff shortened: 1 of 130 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/integration/SchemaRegistrySaslInheritTest.java
```diff
@@ -30,28 +30,30 @@
...
-import org.junit.rules.RuleChain;
-import org.junit.rules.TestRule;
-import org.junit.runner.RunWith;
-import org.junit.runners.JUnit4;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
-@RunWith(JUnit4.class)
```
rigelbm, line 41: Should you tag this as integration test, since it doesn't inherit from ClusterTestHarness?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/ProduceActionIntegrationTest.java
```diff
@@ -63,27 +63,24 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
 // TODO ddimitrov This continues being way too flaky.
 //  Until we fix it (KREST-1542), we should ignore it, as it might be hiding even worse errors.
-@Ignore
-@RunWith(JUnit4.class)
+@Disabled
```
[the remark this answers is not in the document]
ehumber: I ran it when i was doing my produce 500 errors work, and it was pretty temperamental then :)

As above, none of the other tests are marked integration test, leaving as is for now.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/SchemaRegistrySaslInheritTest.java
rigelbm: Most integration tests inherit from ClusterTestHarness, so now they get the annotation from inheritance. The fact this one missed the annotation is a bug. I never realized we needed it. Now we do. I think we should have fixed it. :(

ijuma: We can do this and other ones missing the annotation in their own PR.

PR #970, opened 2022-02-01
Fix context initialization in ConsumersResource.

dimitarndimitrov:
Recent muckrake failures pointed to problems in the initialization of our Consume v2 resource on 7.1.x and above.

It turned out that with the recent switch to fully handling the KafkaRestContext lifetime via our DI mechanisms, necessary context initialization happening in filters is done after the eager context initialization of KafkaRestContext for ConsumersResource. That lead to NPEs and HTTP 500s for calls to the Consume v2 API.

The fix here is to switch to lazy context initialization via javax.inject.Provider. This ensures that the necessary post-matching request filters have been executed by the time the context actually is needed.


ehumber (approved):

rigelbm (approved):
LGTM! Awesomeness!

dimitarndimitrov:
That's interesting... I have a local `google-java-format-1.13.0-all-deps.jar` that I use to format specific files that I've changed for PRs. Here the problem seems to be not with what it did output, but with copy-pasting that output to IntelliJ. I'm still not sure what the problem is (it's not tabbed output, it's not wrong auto-format on save, it's not rich-text pasting, I'm not sure what it is).

PR #972, opened 2022-02-08
KREST-4466 Run tests individually and repeatedly

ehumber:
Some of our integration tests were missing the "IntegrationTest" tag, so it wasn't possible to run them individually.

I've added a testing readme to explain how to run the tests individually, and also how to run them repeatedly, as requested.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/README.md
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/README.md
@@ -1,0 +1,34 @@
+Kafka REST Proxy Tests
+================
+
+Unit Tests
+----------
+
+The unit tests can be run using `mvn test`
+
+To run an individual unit test class use, for example, `mvn -Dtest=ProduceRequestTest test`
+
+To run an individual unit test within a test class use, for example, `mvn -Dtest=ProduceRequestTest#testProduceRequestDeserializerAddsSizeFromString test`
+
+Integration Tests
+-----------------
+
+The integration tests live in the integration folder.
+
+They can be run using `mvn failsafe:integration-test`
+
+Individual test classes can be run using, for example, `mvn -Dit.test=ProduceActionIntegrationTest failsafe:integration-test`
+
+And individual tests within a test class can be run using, for example, ` mvn -Dit.test=ProduceActionIntegrationTest#produceBinary failsafe:integration-test`
+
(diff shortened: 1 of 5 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/README.md
```diff
@@ -0,0 +1,34 @@
...
+-----
+
+To run individual tests you must be within the kafka-rest project folder (eg `~/git/kafka-rest/kafka-rest`) otherwise no tests will be found.
+
+If you make changes to the code, you will need to rebuild before running the tests for any code changes to be picked up. For example rebuild with `mvn package -DskipTests=true`
+
+To run the same test multiple times, there are several options, which include:
+
+- Replace `@Test` with `@RepeatedTest(5)` (remembering to rebuild)
+- Put a simple while loop around the mvn command `while true; do mvn -Dit.test=ProduceActionIntegrationTest failsafe:integration-test; done`
```
dimitarndimitrov, line 34: nit: We can even specify the exact number of runs/re-runs - we just need to replace `while true;` with `for i in {1..<the number of runs>};`.

kafka-rest/src/test/java/io/confluent/kafkarest/README.md
```diff
@@ -0,0 +1,34 @@
...
+Individual test classes can be run using, for example, `mvn -Dit.test=ProduceActionIntegrationTest failsafe:integration-test`
+
+And individual tests within a test class can be run using, for example, ` mvn -Dit.test=ProduceActionIntegrationTest#produceBinary failsafe:integration-test`
+
+Notes
+-----
+
+To run individual tests you must be within the kafka-rest project folder (eg `~/git/kafka-rest/kafka-rest`) otherwise no tests will be found.
+
+If you make changes to the code, you will need to rebuild before running the tests for any code changes to be picked up. For example rebuild with `mvn package -DskipTests=true`
```
dimitarndimitrov, line 29: nit: Without checking what exactly might be causing a problem and what not, I'd err on the side of safety and recommend to also include cleaning - `mvn clean package -DskipTests` instead.

dimitarndimitrov (approved):
+1.

@ehumber do you know if a similar change needs to be done for our non-public codebase?

PR #973, opened 2022-02-09
KREST-4356 Optimize CompletionExceptions when rate-limiting

dimitarndimitrov:
This change avoids the costly stack trace filling that is done for the `CompletionException` that decorates our `RateLimitExceededException` thrown in `ProduceAction.produce`.

A more detailed description of the related performance experiments and profiles can be found in the corresponding JIRA.


ehumber (approved):

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/StacklessCompletionException.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.ratelimit;
```
rigelbm, line 16: I would put this in `kafkarest.common` instead, where `CompletableFutures` is.

rigelbm (approved):
LGTM

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/StacklessCompletionException.java
dimitarndimitrov: Good suggestion, but now that you bring it, I think `kafkarest.exceptions` is an even better place for this.

PR #974, opened 2022-02-09
Move ProducerMetrics to resources/v3.

rigelbm:
It is in the wrong place. Found it when changing stuff in `KafkaModule`.


ehumber (approved):

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -355,35 +357,31 @@ public static CompoundStat percentiles(Map<MetricName, Double> percentiles) {
...
-    private MetricsContext metricsContext;
-    private JmxReporter reporter;
+    final MetricsContext metricsContext;
+    final JmxReporter reporter;
     // 1 sample per second
-    private int numSamples = 10;
-    private long sampleWindowMs = 10000;
-    private Time time = Time.SYSTEM;
+    final int numSamples = 10;
+    final long sampleWindowMs = 10000;
```
dimitarndimitrov, line 366: I'm not sure these make sense. If we are always going to use these values for them, let's make them... - you guessed it, compile time constants :)

dimitarndimitrov (approved):
+1.

A couple of nits below.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -153,12 +156,12 @@ private void setupMetricBeans() {
     log.info("Successfully registered kafka-rest produce metrics with JMX");
   }
 
-  private class BeanCoordinate {
-    private final String beanName;
-    private final Map<String, String> tags;
+  private static final class BeanCoordinate {
+    final String beanName;
+    final Map<String, String> tags;
```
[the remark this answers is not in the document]
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -50,339 +40,226 @@
...
-    this.fullyQualifiedRequestSizeSensor =
-        String.format(sensorNameTemplate, ProducerMetricsRegistry.REQUEST_SIZE_SENSOR);
-    this.fullyQualifiedResponseSensor =
-        String.format(sensorNameTemplate, ProducerMetricsRegistry.RESPONSE_SENSOR);
-    setupMetricBeans();
-  }
+  private static final String GROUP_NAME = "produce-api-metrics";
+  private static final ImmutableMap<String, String> METRIC_TAGS = ImmutableMap.of();
+  private static final int NUM_SAMPLES = 10;
+  private static final long SAMPLE_WINDOWS_MS = 10000;
```
dimitarndimitrov, line 48: nit: Use `L` to switch to a `long` literal. Also I'd suggest `SAMPLE_WINDOWS...` -> `SAMPLE_WINDOW...` - the fact that there are multiple windows during the metric recording is commonly understood I think.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -50,339 +40,226 @@
...
-      }
-      if (!(o instanceof BeanCoordinate)) {
-        return false;
-      }
-      BeanCoordinate that = (BeanCoordinate) o;
-      return Objects.equals(beanName, that.beanName) && Objects.equals(tags, that.tags);
-    }
+  static final String RECORD_ERROR_RATE_METRIC_NAME = "record-error-rate";
+  private static final String RECORD_ERROR_RATE_METRIC_DOC =
+      "The average per-second number of record sends that resulted in errors.";
```
dimitarndimitrov, line 90: nit: These sound way too different while in fact they seem to be rate and count metrics for the same event type (produce send errors). I'd suggest prepending `ERROR_COUNT_WINDOWED...` with `RECORD_`.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -50,339 +40,226 @@
...
-    ProduceMetricMBean(BeanCoordinate beanCoordinate) {
-      this.beanCoordinate = requireNonNull(beanCoordinate);
-    }
+  static final String REQUEST_LATENCY_MAX_METRIC_NAME = "request-latency-max";
+  private static final String REQUEST_LATENCY_MAX_METRIC_DOC = "The max request latency";
 
-    void recordResponse() {
-      recordMetric(fullyQualifiedResponseSensor, 1.0);
-    }
+  static final String REQUEST_LATENCY_PCT_METRIC_PREFIX = "request-latency-avg-";
```
dimitarndimitrov, line 99: nit: Remove `-avg`, percentiles and averages don't make too much sense together.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -50,339 +40,226 @@
...
-                "p95", 0.95,
-                "p99", 0.99,
-                "p999", 0.999),
-            ProducerMetricsRegistry.REQUEST_LATENCY_PCT_DOC)
-        .build();
-    log.info("Successfully registered kafka-rest produce metrics with JMX");
-  }
+  // response
+  static final String RESPONSE_SEND_RATE_METRIC_NAME = "response-rate";
+  private static final String RESPONSE_SEND_RATE_METRIC_DOC =
```
dimitarndimitrov, line 76: nit: `SEND` seems redundant here.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -50,339 +40,226 @@
...
-            && beanCoordinate.tags.equals(metricName.tags())) {
-          metrics.removeMetric(metricName);
-        }
+  private void removeExistingMetrics() {
+    for (MetricName metricName : metrics.metrics().keySet()) {
+      if (metricName.group().equals(GROUP_NAME) && METRIC_TAGS.equals(metricName.tags())) {
+        metrics.removeMetric(metricName);
       }
     }
   }
```
dimitarndimitrov, line 157: When are we expecting this not to match any metric? And additionally, if this is only used when setting up sensors, and we are using these solely for aggregate metrics (not e.g. per-tenant metrics), and we are treating the `ProducerMetrics` instance as a commonly provided singleton, do we really need that step?

dimitarndimitrov, line 157: When are we expecting this not to match any metric?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
```diff
@@ -97,33 +95,19 @@ public void dispose(ExecutorService executorService) {
...
-        synchronized (this) {
-          localProducerMetrics = producerMetrics;
-          if (localProducerMetrics == null) {
-            producerMetrics =
-                localProducerMetrics = new ProducerMetrics(context.get().getConfig(), Time.SYSTEM);
-          }
-        }
-      }
-      return localProducerMetrics;
+      return new ProducerMetrics(config.get(), Time.SYSTEM);
```
dimitarndimitrov, line 107: IIRC in the original changes, Tom had concerns that this is not being initialized only once and in a thread-safe manner with the scope binding he was using at that point. Is the current HK2 incantation ensuring that this won't be called more than once? The non-singleton factory providing a singleton instance via an instance method is pretty quirky.

dimitarndimitrov (commented):
It's probably a good idea to also include @thomaskwscott to this review. By now I have forgotten what was the reasoning behind the JMX bean + `BeanCoordinate` scaffolding that Tom had to implement.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: Done. Note that the resulting JMX metric name will be different. So, if we are using that in any dashboard, it will need to be updated.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: Done. Agreed. Seems futile, specially given this will be a singleton.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/V3ResourcesModule.java
rigelbm: Singleton property, including the fact this will be initialized only once, is enforced by HK2. The factory itself will be instantiated twice, once to provide the instance and another to dispose of it. Since the factory has no state, it's not a problem. Factories should not have any persistent state.

PR #975, opened 2022-02-10
KREST-4577 Fix Metadata API Test

ehumber:
Fix typo where we were checking the wrong response object

restServer sees "address already in use" very occassionally - if this happens shut down, wait and try again

On a restart, the original metrics have already been registered and may be be cleared up yet, so we need to clear them from the application.  I've gone belt and braces approach.

Sometimes the Kafka metric server is still left behind - so we close this off explicitly too with server.metrics().close();

Sometimes kafka hasn't quite finished making the topics, so when we get the topics, only the first is returned, so these components are run with a testWithRetry. There is an analysis of the Kafka debug logs in the jira that describes this in more detail.


ehumber:
branch in wrong repo

PR #977, opened 2022-02-10
KREST-4577 Fix Metadata API Test

ehumber:
Fix typo where we were checking the wrong response object

restServer sees "address already in use" very occasionally - if this happens shut down, wait and try again

On a restart, the original metrics have already been registered and may be be cleared up yet, so we need to clear them from the application.  I've gone belt and braces approach in clearing the metrics.

Sometimes the Kafka metric server is still left behind - so we close this off explicitly too with server.metrics().close();

Sometimes kafka hasn't quite finished making the topics, so when we get the topics, only the first is returned, so these components are run with a testWithRetry. There is an analysis of the Kafka debug logs in the jira that describes this in more detail.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
@@ -225,9 +225,15 @@ public abstract class ClusterTestHarness {
     try {
       startRest();
     } catch (IOException e) { // sometimes we get an address already in use exception
+      log.warn("IOException when attempting to start rest, trying again", e);
       stopRest();
       Thread.sleep(ONE_SECOND_MS);
-      startRest();
+      try {
+        startRest();
+      } catch (IOException e2) {
+        log.error("Restart of rest server failed", e2);
+        throw e2;
+      }
     }
     log.info("Completed setup of {}", getClass().getSimpleName());
   }
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
@@ -143,6 +143,8 @@ public abstract class ClusterTestHarness {
   protected Server restServer = null;
   protected String restConnect = null;
 
+  private static final long SLEEP_MS = 1000;
+
   public ClusterTestHarness() {
     this(DEFAULT_NUM_BROKERS, false);
   }
@@ -219,10 +221,33 @@ public abstract class ClusterTestHarness {
     restProperties.put("producer." + ProducerConfig.MAX_BLOCK_MS_CONFIG, "5000");
 
     restConfig = new KafkaRestConfig(restProperties);
+
+    try {
+      startRest();
+    } catch (IOException e) { // sometimes we get an address already in use exception
+      stopRest();
+      Thread.sleep(SLEEP_MS);
+      startRest();
+    }
+    log.info("Completed setup of {}", getClass().getSimpleName());
+  }
+
+  private void startRest() throws Exception {
     restApp = new KafkaRestApplication(restConfig);
     restServer = restApp.createServer();
     restServer.start();
-    log.info("Completed setup of {}", getClass().getSimpleName());
+  }
+
+  private void stopRest() throws Exception {
+    if (restApp != null) {
+      restApp.stop();
+      restApp.getMetrics().close();
+      restApp.getMetrics().metrics().clear();
+    }
+    if (restServer != null) {
+      restServer.stop();
+      restServer.join();
+    }
   }
 
   private void startBrokersConcurrently(int numBrokers) {
@@ -298,10 +323,7 @@ public abstract class ClusterTestHarness {
   @AfterEach
   public void tearDown() throws Exception {
     log.info("Starting teardown of {}", getClass().getSimpleName());
-    if (restServer != null) {
-      restServer.stop();
-      restServer.join();
-    }
+    stopRest();
 
     if (schemaRegServer != null) {
       schemaRegServer.stop();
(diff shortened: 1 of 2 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/integration/MetadataAPITest.java
```diff
@@ -189,4 +177,25 @@ public void testPartitionsList() throws InterruptedException {
...
+
+  private void verifyTopicGet() {
+    Response response = request("/topics").get();
+    assertOKResponse(response, Versions.KAFKA_V2_JSON);
+    final List<String> topicsResponse =
+        tryReadEntityOrLog(response, new GenericType<List<String>>() {});
+    assertEquals(Arrays.asList(topic1Name, topic2Name), topicsResponse);
+  }
+
+  private void verifyPartitionGet(String topicName, int size, int numPartitions) {
```
dimitarndimitrov, line 189: These parameters seem confusing to me. In the current code, it looks like both of them refer to the number of topic partitions. Was the idea to have `size` be `numPartitions`, and to have `numPartitions` be `numReplicas`, similar to the way the `createTopic` method works?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/MetadataAPITest.java
```diff
@@ -44,6 +45,8 @@
  */
 public class MetadataAPITest extends ClusterTestHarness {
 
+  private static final long SLEEP_MS = 500;
+
```
dimitarndimitrov, line 49: nit: This seems redundant.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -143,6 +143,8 @@ public static int choosePort() {
   protected Server restServer = null;
   protected String restConnect = null;
 
+  private static final long SLEEP_MS = 1000;
```
dimitarndimitrov, line 146: nit: Let's add an `L` suffix to make this an actual `long` primitive. I personally also think this should be called `ONE_SECOND_IN_MILLIS` or `ONE_SECOND_MS`.

ehumber: I left it without the time value in the name because I wonder if we will need to increase the sleep time in the future, it's an arbitrary length sleep, and it may not be right at 1s.

But, the flip side is, if we have to do that we can always rename it to SLEEP_MS so I don't have strong feeling.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -219,10 +221,33 @@ public void setUp() throws Exception {
...
+
+    try {
+      startRest();
+    } catch (IOException e) { // sometimes we get an address already in use exception
+      stopRest();
+      Thread.sleep(SLEEP_MS);
+      startRest();
+    }
+    log.info("Completed setup of {}", getClass().getSimpleName());
+  }
```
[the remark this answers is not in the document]
ehumber: But where do we stop, we could loop forever catching and retrying?

This happens frequently enough to observe on my laptop, and occasionally in jenkins too. I don't see it happen after this change.

Can we start with the single retry as it is, and it's simple to add in some more with an exponential sleep back off if it doesn't help.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/MetadataAPITest.java
```diff
@@ -189,4 +177,25 @@ public void testPartitionsList() throws InterruptedException {
...
         Versions.KAFKA_V2_JSON);
   }
+
+  private void verifyTopicGet() {
+    Response response = request("/topics").get();
+    assertOKResponse(response, Versions.KAFKA_V2_JSON);
+    final List<String> topicsResponse =
+        tryReadEntityOrLog(response, new GenericType<List<String>>() {});
+    assertEquals(Arrays.asList(topic1Name, topic2Name), topicsResponse);
+  }
```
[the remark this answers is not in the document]
ehumber: In the end it was only used in one place, so I've taken the method out and left the whole thing inline in the calling method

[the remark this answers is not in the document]
dimitarndimitrov: Right, but then we have a method that specifically checks for `topic1Name` and `topic2Name` in the topic response, which feels both unintuitive and harder to read. I personally don't like that, but am OK if you as an author prefer leaving it as is.

dimitarndimitrov (approved):
+1.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/MetadataAPITest.java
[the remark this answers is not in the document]
ehumber: Yup, I agree, we shouldn't check for specific topicNames in the method for the reasons you say, and that it's not generic.

I think the git commentary doesn't show the change very well now though, I've got rid of the method entirely and put it back to how it used to be, which is inline, but just with the addition of the testWithRetry around it.  At that point I think it is right to just pass in topic1Name and topic2Name.

If that's still not quite ideal, let me know.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
ehumber: Yup, I did change it already to say ONE_SECOND_MS

image

I wonder if git is having a moment....

ehumber:
@dimitarndimitrov happy for me to merge now?

dimitarndimitrov:
> @dimitarndimitrov happy for me to merge now?

Yep.

PR #979, opened 2022-02-15
KREST-4605 Deflake all the tests that extend AbstractConsumerTest

ehumber:
KafkaConsumerReadTask
- exception stack trace not logged, added in so we have a starting point for debugging the intermittently failing test

AbstractConsumerTest
 - a 404 is returned somewhere in startConsumerMessages (can't prove where, but one of the assertOKResponses fails with a 404) so check for the 404 on both, and then retry after a pause if it is returned.  Fail the second time
 - The number of messages consumed doesn't always match how many we expect.  I could reproduce this locally, and a single retry didn't always help, so I've added in two.  If it still fails we can look at reworking this a different way
 - consumeForTimeout is inherently flakey as it is so reliant on times.  I've doubled the wait period slack to 1s, from 501ms to see if that helps at all, but otherwise we will need to look at a mocked back end.

ClusterTestHarness
 - Made a generic produce method, as this was repeated with different object types
 - The produce was also failing in some cases so I've added some retry logic to try and resend each message 3 times with a pause each time before failure

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
@@ -19,6 +19,7 @@ import static io.confluent.kafkarest.TestUtils.assertOKResponse;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 import io.confluent.kafkarest.Errors;
 import io.confluent.kafkarest.KafkaRestConfig;
@@ -48,6 +49,8 @@ import javax.ws.rs.core.Response.Status;
 import org.apache.kafka.clients.producer.ProducerRecord;
 
 public class AbstractConsumerTest extends ClusterTestHarness {
+
+  private static final long ONE_SECOND_MS = 1000L;
 
   public AbstractConsumerTest() {}
 
@@ -100,8 +103,20 @@ public class AbstractConsumerTest extends ClusterTestHarness {
 
   protected String startConsumeMessages(
       String groupName, List<String> topics, EmbeddedFormat format, String expectedMediatype) {
-    Response createResponse = createConsumerInstance(groupName, null, null, format);
-    assertOKResponse(createResponse, Versions.KAFKA_V2_JSON);
+    Response createResponse = null;
+    try {
+      createResponse = createConsumerInstance(groupName, null, null, format);
+      assertOKResponse(createResponse, Versions.KAFKA_V2_JSON);
+    } catch (AssertionError e) {
+      if (createResponse != null
+          && createResponse.getStatus() == Response.Status.NOT_FOUND.getStatusCode()) {
+        pause();
+        createResponse = createConsumerInstance(groupName, null, null, format);
+        assertOKResponse(createResponse, Versions.KAFKA_V2_JSON);
+      } else {
+        fail("Create consumer instance failed.", e);
+      }
+    }
 
     CreateConsumerInstanceResponse instanceResponse =
         TestUtils.tryReadEntityOrLog(createResponse, CreateConsumerInstanceResponse.class);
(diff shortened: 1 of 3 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
```diff
@@ -119,15 +134,25 @@ protected String startConsumeMessages(
...
+    try {
+      response =
+          request(instanceResponse.getBaseUri() + "/records").accept(expectedMediatype).get();
+      assertOKResponse(response, expectedMediatype);
+    } catch (AssertionError e) {
+      if (response != null && response.getStatus() == Response.Status.NOT_FOUND.getStatusCode()) {
+        pause();
+        response =
+            request(instanceResponse.getBaseUri() + "/records").accept(expectedMediatype).get();
+        assertOKResponse(response, Versions.KAFKA_V2_JSON);
```
dimitarndimitrov, line 148: Let me first note that I don't currently understand how our Consume v2 weighted media types for responses work. But shouldn't the retry still expect the originally expected media type.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -89,6 +89,8 @@
 public abstract class ClusterTestHarness {
 
   private static final Logger log = LoggerFactory.getLogger(ClusterTestHarness.class);
+  private static final long ONE_SECOND_MS = 1000l;
```
dimitarndimitrov, line 92: super-nit: I think `L` is preferable over `l` as the latter kinda looks like a `1`.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
```diff
@@ -184,12 +209,42 @@ protected String startConsumeMessages(
...
     List<RecordType> consumed = TestUtils.tryReadEntityOrLog(response, responseEntityType);
-    assertEquals(records.size(), consumed.size());
+    try {
+      assertEquals(records.size(), consumed.size());
+    } catch (AssertionError e) {
+      pause();
+      Response response2 = request(instanceUri + "/records").accept(accept).get();
+      assertOKResponse(response2, responseMediatype);
+      List<RecordType> consumed2 = TestUtils.tryReadEntityOrLog(response2, responseEntityType);
+      consumed.addAll(consumed2);
```
dimitarndimitrov, line 219: nit: You can simplify `consumeMessages` by extracting this into a recursive method with an explicit recursion depth / counter argument.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -533,14 +527,36 @@ protected final void produceJsonMessages(List<ProducerRecord<Object, Object>> re
...
+    int errorCount = 0;
+    while (!sent) {
       try {
         producer.send(rec).get();
+        sent = true;
       } catch (Exception e) {
-        fail("Couldn't produce input messages to Kafka: " + e);
+        errorCount++;
+        if (errorCount > RETRY_COUNT) {
+          fail("Couldn't produce input messages to Kafka: " + e);
```
dimitarndimitrov, line 547: nit: Do we need to explicitly close the producer first?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -533,14 +527,36 @@ protected final void produceJsonMessages(List<ProducerRecord<Object, Object>> re
...
     props.setProperty(ProducerConfig.ACKS_CONFIG, "all");
     Producer<Object, Object> producer = new KafkaProducer<Object, Object>(props);
+
     for (ProducerRecord<Object, Object> rec : records) {
+      doProduce(producer, rec);
+    }
+    producer.close();
+  }
+
+  private <T, R extends Producer> void doProduce(R producer, ProducerRecord<T, T> rec) {
```
dimitarndimitrov, line 537: nit: I think this can be mostly replaced by `TestUtils.testWithRetry`.

dimitarndimitrov (approved):
+1.

Some comments below, but most of them are nits.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
ehumber: No, oops, that's a copy paste error.  Good spot.  The test still passes though... I'll go find out why

ehumber: It worked anyway (even through the failure case when I forced the exception throw to test the error path) because I tested with ConsumerJsonTest...

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
ehumber: Hmm, good question. If it's a transient KafkaException  (eg dodgy network) we want to retry without a close (which is what I'd written to start with)

But tried coding don't close for a RetriableException and do close for the rest, and it just got a bit tangled, so I now close for all errors.

dimitarndimitrov (approved):
+1, awesome work and thanks for addressing All The Things ™.

PR #980, opened 2022-02-16
Disable producer idempotency until KAFKA-13668 is fixed.

rigelbm:
For the idempotent producer, some errors are permanent, meaning the producer instance becomes unusable afterwards. Authorization errors are one of these. This behaviour disagrees with the Kafka REST model, which assumes the producer to be stateless across requests. Until we either adapt Kafka REST to the idempotent model, or authorization errors become non-fatal, we will disable idempotent producers in Kafka REST.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -873,6 +873,14 @@ public class KafkaRestConfig extends RestConfig {
 
     Properties producerProperties = new Properties();
     producerProperties.putAll(producerConfigs);
+
+    // KREST-4606: Disable idempotency until KAFKA-13668 is fixed.
+    if (Boolean.parseBoolean(producerProperties.getProperty("enable.idempotence"))) {
+      log.warn(
+          "Idempotent producers are not supported in Kafka REST. Ignoring enable.idempotence'.");
+    }
+    producerProperties.put("enable.idempotence", "false");
+
     addTelemetryReporterProperties(producerProperties);
 
     return producerProperties;
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -874,7 +874,8 @@ public class KafkaRestConfig extends RestConfig {
     Properties producerProperties = new Properties();
     producerProperties.putAll(producerConfigs);
 
-    // KREST-4606: Disable idempotency until KAFKA-13668 is fixed.
+    // KREST-4606: Disable idempotency until at the very least KAFKA-13668 is fixed, but maybe
+    // forever.
     if (Boolean.parseBoolean(producerProperties.getProperty("enable.idempotence"))) {
       log.warn(
           "Idempotent producers are not supported in Kafka REST. Ignoring 'enable.idempotence'.");
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -877,7 +877,7 @@ public class KafkaRestConfig extends RestConfig {
     // KREST-4606: Disable idempotency until KAFKA-13668 is fixed.
     if (Boolean.parseBoolean(producerProperties.getProperty("enable.idempotence"))) {
       log.warn(
-          "Idempotent producers are not supported in Kafka REST. Ignoring enable.idempotence'.");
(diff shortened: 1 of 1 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -873,6 +873,14 @@ public Properties getProducerProperties() {
 
     Properties producerProperties = new Properties();
     producerProperties.putAll(producerConfigs);
+
+    // KREST-4606: Disable idempotency until KAFKA-13668 is fixed.
+    if (Boolean.parseBoolean(producerProperties.getProperty("enable.idempotence"))) {
+      log.warn(
+          "Idempotent producers are not supported in Kafka REST. Ignoring enable.idempotence'.");
```
dimitarndimitrov, line 880: nit: A single quote seems to be missing just before _enable_.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -873,6 +873,14 @@ public Properties getProducerProperties() {
 
     Properties producerProperties = new Properties();
     producerProperties.putAll(producerConfigs);
+
+    // KREST-4606: Disable idempotency until KAFKA-13668 is fixed.
```
dimitarndimitrov, line 877: Is this actually correct though? Would we allow people enabling idempotency without warning them that Kafka REST might not support it the way they expect it, even after KAFKA-13668 is fixed? If not (which is what I assume), let's remove or reword this comment.

dimitarndimitrov (approved):
+1.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
ehumber: I agree - after our chat this morning, I think we came to the conclusion that idempotency in REST produce is a feature request.  While the REST produce client to kafka part of the flow might be idempotent, the http client to the REST produce client is not.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
ehumber: I'm not quite decided, but is warn strong enough here?  If someone thinks they are getting idempotence, and they are not, that's quite serious isn't it?  If it's something that requires once only delivery of messages isn't turned on, even after they set the setting is set to on (eg payment transfer happens twice instead of once) that could have consequences.

What do you think?

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: Yes, we would. That's current behaviour. I think Kafka REST is very explicit that passing a `producer.` config is equivalent to passing down the config to the underlying producer. We are not talking about having a `idempotence.enable` config in Kafka REST itself. People should be aware that passing that config to the Producer changes the producer only, but has no other effect.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
rigelbm: There are already no guarantees, due to HTTP retries, so I don't think it is very serious. `error` should only be used for things that are critical. In this case, if we did `error`, we should all exit the application. Since we aren't, I think `warn` is good enough.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
dimitarndimitrov: I think this might easily violate the [principle of least astonishement](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). People might e.g. assume that Kafka REST guarantees to reuse the same Producer underneath and that it somehow ensures idempotency guarantees across the whole write path. Without understanding the exact lifecycle of a Producer instance, which as we all know is not trivial, they can easily assume more than there is.
I think the added value to people that actually understand all the tricky details might be outweighed by the potential problems for people that don't, so I'd vote for not changing this even after KAFKA-13668 is fixed, and maybe adding an `force.enable.idempotency` Kafka REST config that can gate enabling the idempotency.

rigelbm: If that was the case, I think we should disable idempotent producer for all versions, not just `7.0.x+`. I think that is out of scope for this PR. We should disable it now for `7.0.x+` because we have a real problem. And then we can have a discussion whether we just don't support this at all for all versions.

rigelbm: Rephrased the comment to reflect our collective thinking.

PR #981, opened 2022-02-16
KREST-4608 unflake consumergroup tests

ehumber:
ConsumerGroupLagSummariesResourceIntegrationTest saw consumerGroupId=consumer-group-1, maxLag=3, totalLag=3 when it was expecting consumerGroupId=consumer-group-1, maxLag=6, totalLag=9

I can't reproduce locally but my best guess is that the lags just haven't been updated yet (see the issue for an explanation of the test flow), so give it another chance

listConsumerGroups_returnsConsumerGroups saw partitionAssignor=range, state=STABLE when it was expecting partitionAssignor=, state=PREPARING_REBALANCE

I would say that either is valid, so I've added in the "anyOf" matcher

### DIFF
--- a/kafka-rest/pom.xml
+++ b/kafka-rest/pom.xml
@@ -153,6 +153,11 @@
            <artifactId>bcprov-ext-jdk15on</artifactId>
            <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/ConsumerGroupsResourceIntegrationTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/ConsumerGroupsResourceIntegrationTest.java
@@ -16,6 +16,9 @@
 package io.confluent.kafkarest.integration.v3;
 
 import static java.util.Collections.singletonList;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.AnyOf.anyOf;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import io.confluent.kafkarest.entities.ConsumerGroup.State;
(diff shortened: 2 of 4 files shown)


kafka-rest/pom.xml
```diff
@@ -153,6 +153,11 @@
            <artifactId>bcprov-ext-jdk15on</artifactId>
            <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+            <scope>test</scope>
+        </dependency>
```
dimitarndimitrov, line 160: nit: Should this actually be `hamcrest-all` or it can be just `hamcrest-library`? Unfortunately we seem to be using 1.x - Hamcrest 2.x packages everything in a single JAR.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/ConsumerGroupsResourceIntegrationTest.java
```diff
@@ -107,12 +110,58 @@ public void listConsumerGroups_returnsConsumerGroups() {
...
+                                    .setResourceName(
+                                        "crn:///kafka="
+                                            + clusterId
+                                            + "/consumer-group=consumer-group-1")
+                                    .build())
+                            .setClusterId(clusterId)
+                            .setConsumerGroupId("consumer-group-1")
+                            .setSimple(false)
+                            .setPartitionAssignor("range")
+                            .setState(State.STABLE)
```
dimitarndimitrov, line 139: nit: This seems to be the only difference from `expected1` - let's use a helper method for the creation of the responses, this will make the code shorter and it will make the difference obvious.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/ConsumerGroupsResourceIntegrationTest.java
```diff
@@ -153,7 +202,41 @@ public void getConsumerGroup_returnsConsumerGroup() {
     consumer2.poll(Duration.ofSeconds(1));
     consumer3.poll(Duration.ofSeconds(1));
 
-    GetConsumerGroupResponse expected =
+    GetConsumerGroupResponse expected1 =
```
dimitarndimitrov, line 205: Same as above.

dimitarndimitrov (approved):
+1, some nits below.

dimitarndimitrov (approved):
+1.

PR #985, opened 2022-02-21
KREST-4687 avro consume still fails

ehumber:
Looking at the logs for this intermittent failure, the consumer attempts to consume messages 4 times, over a course of 15 seconds, but only gets one record, from the second attempt.  Each GET returns a 200, the failed consumes just time out, not finding any records.

There are no logs relating to the produce (these are only written at trace level for the produce client) so it’s not possible to tell if the produces all went through - is it the produce failing or the consume that's causing the problem here?

Produce is acks=all, the consumer’s auto.offset.reset is the default, which is latest.

I’ve added in diagnostics plus some guesses at fixing the problem:

trace level logging for the produce and consume clients so we can diagnose failures more easily

logging of which offset we are getting back on the consume, so that we can see if it’s first, last or something else.  As part of this tidied up a set of nested try/catches into a while loop in consumeMessage

auto.offset.reset can now be set, and in the majority of cases is set to earliest.  This means that a new consumer group will read from the beginning, rather than there being the potential of missing some messages.  It’s not clear which is the single message being consumed, so I’ve added this in as a guess as to one reason that the other three messages are not got.  (The one case it isn’t set to earliest is the consume by timestamp test, where it is set to default).

In case it’s the produce going awry, I check that the record metadata that is returned by the produce has an offset, and try again if it doesn’t.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
@@ -32,10 +32,13 @@ import io.confluent.kafkarest.entities.v2.CommitOffsetsResponse;
 import io.confluent.kafkarest.entities.v2.ConsumerSubscriptionRecord;
 import io.confluent.kafkarest.entities.v2.CreateConsumerInstanceRequest;
 import io.confluent.kafkarest.entities.v2.CreateConsumerInstanceResponse;
+import io.confluent.kafkarest.entities.v2.JsonConsumerRecord;
+import io.confluent.kafkarest.entities.v2.SchemaConsumerRecord;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -47,10 +50,13 @@ import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class AbstractConsumerTest extends ClusterTestHarness {
 
   private static final long ONE_SECOND_MS = 1000L;
+  private static final Logger log = LoggerFactory.getLogger(AbstractConsumerTest.class);
 
   public AbstractConsumerTest() {}
 
@@ -59,7 +65,7 @@ public class AbstractConsumerTest extends ClusterTestHarness {
   }
 
   protected Response createConsumerInstance(
-      String groupName, String id, String name, EmbeddedFormat format) {
+      String groupName, String id, String name, EmbeddedFormat format, String autoOffsetReset) {
     CreateConsumerInstanceRequest config = null;
     if (id != null || name != null || format != null) {
       config =
@@ -67,7 +73,7 @@ public class AbstractConsumerTest extends ClusterTestHarness {
               id,
               name,
               format != null ? format.toString() : null,
-              /* autoOffsetReset= */ null,
+              /* autoOffsetReset= */ autoOffsetReset,
               /* autoCommitEnable */ null,
               /* responseMinBytes= */ null,
               /* requestWaitMs= */ null);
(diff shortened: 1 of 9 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
```diff
@@ -206,33 +220,46 @@ protected String startConsumeMessages(
...
-      consumed.addAll(consumed2);
+
+    int retryCount = 0;
+    final int TOTAL_RETRIES = 3;
+    List<RecordType> consumed = new ArrayList<>();
+
+    while (retryCount < TOTAL_RETRIES) {
+      Response response = request(instanceUri + "/records").accept(accept).get();
+      assertOKResponse(response, responseMediatype);
+      consumed.addAll(TestUtils.tryReadEntityOrLog(response, responseEntityType));
```
dimitarndimitrov, line 231: I think that `testWithRetry` should work OK with externally kept `final` collection being modified - is there a different reason to prefer a manual retry approach?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
```diff
@@ -206,33 +220,46 @@ protected String startConsumeMessages(
...
+    log.info("Records produced : {}", records.size());
+    log.info("Consume failed.  Record offsets are:");
+    consumed.forEach(
+        record -> {
+          if (record instanceof SchemaConsumerRecord) {
+            log.info("{}", ((SchemaConsumerRecord) record).getOffset());
+          } else if (record instanceof BinaryConsumerRecord) {
+            log.info("{}", ((BinaryConsumerRecord) record).getOffset());
+          } else if (record instanceof JsonConsumerRecord) {
+            log.info("{}", ((JsonConsumerRecord) record).getOffset());
```
dimitarndimitrov, line 256: nit: Does it make sense to distinguish between the different record types? It shouldn't cost us anything, should it?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -565,8 +566,12 @@ protected final void produceJsonMessages(List<ProducerRecord<Object, Object>> re
           final KafkaProducer<T, T> producer = createProducer.get();
           boolean sent = false;
           try {
-            producer.send(rec).get();
-            sent = true;
+            RecordMetadata result = producer.send(rec).get();
+            if (result.hasOffset()) {
+              sent = true;
+            } else {
+              log.info("Failed to get offset back from produce {}", result);
```
dimitarndimitrov, line 573: nit: Maybe also log the originally sent record? If there's no offset maybe it's hard to identify what does this result correspond to.

kafka-rest/src/test/resources/log4j.properties
```diff
@@ -6,6 +6,9 @@ log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c:%L)%n
 
 log4j.logger.kafka=ERROR
 log4j.logger.org.apache.kafka=ERROR
+log4j.logger.org.apache.kafka.clients.producer=TRACE
+log4j.logger.org.apache.kafka.clients.consumer=TRACE
```
dimitarndimitrov, line 10: nit: I'm a bit concerned that `TRACE` will create an unmanageable amount of stuff in the logs, but that's entirely based on previous experience from different systems, I have no idea what is the verbosity difference between `TRACE` and `DEBUG` and whether `DEBUG` even contains what you need. If it does though, I'd suggest logging at that, not `TRACE`.

dimitarndimitrov (approved):
+1.

The change makes sense to me, some nits below.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/AbstractConsumerTest.java
ehumber: The different type of records don't extend/implement any super class. Ie there is no ConsumerRecord common parent that we can cast the unknown-typed record to to be able to call getOffset on it.  And without the cast, all you can call on record is Object's methods :(

So we have to deal with them one at a time.

kafka-rest/src/test/resources/log4j.properties
ehumber: I've taken this out, the rest of the unit tests log at error, so we'll stick with that.

But, I think it makes it virtually impossible to debug failures that happen on jenkins, so once our tests are a bit more robust, and we've cleared up all the easy to reproduce locally failures, we should consider flipping this to be TRACE for the clients and for REST so we have some chance of debugging the hard failures too.

PR #987, opened 2022-02-24
Add a Fake Kafka/SR test environment.

rigelbm:
Also convert ProduceActionIntegrationTest to use the new environment.

Using a real Kafka/SR implementation in our tests has the benefit that we can actually do an end-to-end test from Kafka REST to Kafka/SR. In the other hand, because of the distributed nature of Kafka, the tests become non-deterministic. This causes flakiness, which oftentimes we have to resort to sleep/retry to "fix".

This PR adds a fake Kafka/SR environment to address the aforementioned issues. The main design points are:

 1. No disk.
 2. No networking.
 3. No async.

By implementing Kafka following the above requirements, we can ensure tests are fully deterministic and fast(!!).

Note: I was mid-way implementing an AdminClient, until I noticed I didn't have to. This might explain some of the design decisions. I deleted all the code that was not used anymore, but the way the code is structured still reflects the extras needed for implementing the Admin operations. Eventually we will need to implement Admin anyway, so I think leaving the structure as is, without the extra code, is a good compromise.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/testing/fake/FakeKafkaProducer.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/testing/fake/FakeKafkaProducer.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2022 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.testing.fake;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.Objects.requireNonNull;
+
+import com.google.protobuf.ByteString;
+import io.confluent.kafkarest.common.KafkaFutures;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.Future;
+import java.util.stream.Stream;
+import org.apache.kafka.clients.consumer.ConsumerGroupMetadata;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.clients.producer.Callback;
+import org.apache.kafka.clients.producer.Partitioner;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.clients.producer.RoundRobinPartitioner;
+import org.apache.kafka.common.KafkaFuture;
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.errors.ProducerFencedException;
+import org.apache.kafka.common.serialization.Serializer;
+
+final class FakeKafkaProducer<K, V> implements Producer<K, V> {
+  private final FakeKafkaCluster cluster;
+  private final Serializer<K> keySerializer;
+  private final Serializer<V> valueSerializer;
+  private final Partitioner partitioner = new RoundRobinPartitioner();
+
+  FakeKafkaProducer(
+      FakeKafkaCluster cluster, Serializer<K> keySerializer, Serializer<V> valueSerializer) {
+    this.cluster = requireNonNull(cluster);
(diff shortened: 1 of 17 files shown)


ijuma:
> Using a real Kafka/SR implementation in our tests has the benefit that we can actually do an end-to-end test from Kafka REST to Kafka/SR. In the other hand, because of the distributed nature of Kafka, the tests become non-deterministic. This causes flakiness, which oftentimes we have to resort to sleep/retry to "fix".

Right, so ideally we'd have a small number of tests that use the real thing and most tests use the fake thing. Is that the plan?

rigelbm:
> > Using a real Kafka/SR implementation in our tests has the benefit that we can actually do an end-to-end test from Kafka REST to Kafka/SR. In the other hand, because of the distributed nature of Kafka, the tests become non-deterministic. This causes flakiness, which oftentimes we have to resort to sleep/retry to "fix".
> 
> Right, so ideally we'd have a small number of tests that use the real thing and most tests would use the fake thing. Is that the plan?

Exactly. I moved this one integration test to test the waters. The idea is that most of our coverage can be achieved with the more stable fake environment, and then we can have a few "system-test" style tests that setup the real thing, possibly using test-containers.

ehumber:
> > > Using a real Kafka/SR implementation in our tests has the benefit that we can actually do an end-to-end test from Kafka REST to Kafka/SR. In the other hand, because of the distributed nature of Kafka, the tests become non-deterministic. This causes flakiness, which oftentimes we have to resort to sleep/retry to "fix".
> > 
> > 
> > Right, so ideally we'd have a small number of tests that use the real thing and most tests would use the fake thing. Is that the plan?
> 
> Exactly. I moved this one integration test to test the waters. The idea is that most of our coverage can be achieved with the more stable fake environment, and then we can have a few "system-test" style tests that setup the real thing, possibly using test-containers.

@rigelbm , do we have any produce tests left that test against a real kafka and schema registry backend too?

kafka-rest/src/test/java/io/confluent/kafkarest/testing/fake/FakeKafkaProducer.java
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  @Override
+  public void flush() {}
+
+  @Override
+  public List<PartitionInfo> partitionsFor(String topic) {
+    return cluster.getTopic(topic).getPartitions().stream()
+        .map(partition -> partition.toPartitionInfo(cluster))
+        .collect(toImmutableList());
```
dimitarndimitrov, line 152: super-nit: Does Google Java Format like that? I find it much more unreadable than the each-link-from-the-chain-on-its-own-line style.

kafka-rest/src/test/java/io/confluent/kafkarest/testing/fake/FakeKafkaRestTestEnvironment.java
```diff
@@ -0,0 +1,130 @@
...
+    private FakeKafkaModule(FakeKafkaCluster kafkaCluster, FakeSchemaRegistry schemaRegistry) {
+      this.kafkaCluster = requireNonNull(kafkaCluster);
+      this.schemaRegistry = requireNonNull(schemaRegistry);
+    }
+
+    @Override
+    protected void configure() {
+      bind(kafkaCluster).to(FakeKafkaCluster.class).ranked(1);
+      bind(schemaRegistry).to(FakeSchemaRegistry.class).ranked(1);
+      bind(FakeKafkaRestContext.class).to(KafkaRestContext.class).ranked(1);
```
dimitarndimitrov, line 79: nit: Is the ranking needed? If yes, can you please add a short comment why?

kafka-rest/src/test/java/io/confluent/kafkarest/testing/fake/Partition.java
```diff
@@ -0,0 +1,68 @@
...
+  ImmutableList<Replica> getReplicas() {
+    return ImmutableList.<Replica>builder().add(leader).addAll(followers).build();
+  }
+
+  PartitionInfo toPartitionInfo(FakeKafkaCluster cluster) {
+    Node[] replicas =
+        getReplicas().stream()
+            .map(Replica::getBrokerId)
+            .map(cluster::getBroker)
+            .map(Broker::toNode)
```
dimitarndimitrov, line 59: nit: I think this will be both more efficient and more readable if done in a single mapping step.

kafka-rest/src/test/java/io/confluent/kafkarest/testing/fake/Topic.java
```diff
@@ -0,0 +1,43 @@
...
+package io.confluent.kafkarest.testing.fake;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
+
+final class Topic {
+  private final ConcurrentMap<Integer, Partition> partitions = new ConcurrentHashMap<>();
```
dimitarndimitrov, line 25: nit: I think it's better to use `ImmutableMap` for now - once we need the partition change functionality, we'll change this accordingly.

dimitarndimitrov (commented):
Awesome changes!
You'll find multiple comments below that I think we should address first, but the main idea is so neat and simple that I'm wondering - does it make sense to make this its own Maven module that can be exported and reused by other projects?

PR #988, opened 2022-03-03
KREST-4783 Increase slack in consumer timeout test

ehumber:
This test sets a consumer instance timeout of 1000ms, creates a consumer instance with that timeout, does some consuming, and then waits 2100ms before doing the next consume.

It expects that the consume will fail because the consume instance has timed out - it should expire after 1s and the test waits for 2.1s which should be plenty even allowing for slowness.

However, if you look in the consumer instance timeout code, there is a 1000ms wait between cleanup cycles.

So, it’s conceivable that we could wait up to 2000ms for the consumer instance to be cleared.

The current wait in the test is 2100ms (instanceTimeout + slackTime) before checking if the consumer has expired, which doesn’t account for any slowness in the machine.

I've increased the slack time


rigelbm (approved):
LGTM

PR #991, opened 2022-03-08
KREST-4825 Huge timeout in consume test

ehumber:
The consumeTimeout test is failing in the 7.1 CP tests because of huge gaps between API calls.

There is 0.5s between the POST to create the consumer group and the subscription and then over 2 seconds between the subscription create and the GET that does the consume

Before this change the instance expiry (how long the consumer group is kept for) was set to 1s, and so the consumer group was timing out before the GET happened, resulting in a 404.

This PR increases the instance timeout to 3.5 s, which unfortutately adds 2.5s to the consumeTimeout test run time, but won't impact the run time of any other tests.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ConsumerTimeoutTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ConsumerTimeoutTest.java
@@ -30,10 +30,11 @@ public class ConsumerTimeoutTest extends AbstractConsumerTest {
   private static final Integer requestTimeout = 500;
   // This is pretty large since there is sometimes significant overhead to doing a read (e.g.
   // checking topic existence in ZK)
-  private static final Integer instanceTimeout = 1000;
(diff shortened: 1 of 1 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/integration/ConsumerTimeoutTest.java
```diff
@@ -30,10 +30,11 @@
   private static final Integer requestTimeout = 500;
```
dimitarndimitrov, line 30: nit: On all of these I would add a suffix indicating that they are in millis, and would also use capitalized snake case (e.g. `REQUEST_TIMEOUT_MS`).

dimitarndimitrov (approved):
+1.

Thanks for addressing this! It's unfortunate that we have to go this route, but this should do the trick.

ehumber:
Test is a flakey test, merging.

PR #993, opened 2022-03-09
KREST-3323 add global rate limiter for produce

ehumber:
At the moment produce rate limiting is done per clusterId.  This means that if there are lots of clusterIds (as there would be in  MT cluster) then we could swamp the jvm.

This PR adds in a global count and byte rate limiter that will prevent this

The config is set to the same as the per clusterID limiters, so there is no change in behaviour for existing deployments.

I've put the new rate limiters after the existing per clusterId limiters because I think the per cluster ones are more likely to be triggered.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -169,14 +169,37 @@ public class KafkaRestConfig extends RestConfig {
   public static final ConfigDef.Range PRODUCE_MAX_REQUESTS_PER_SECOND_VALIDATOR =
       ConfigDef.Range.between(1, Integer.MAX_VALUE);
 
+  public static final String PRODUCE_MAX_REQUESTS_GLOBAL_PER_SECOND =
+      "api.v3.produce.rate.limit.max.requests.global.per.sec";
+  private static final String PRODUCE_MAX_REQUESTS_GLOBAL_PER_SECOND_DOC =
+      "Maximum number of requests per second before rate limiting is enforced, across all "
+          + "clusterIds. "
+          + "Messages produced that exceed the rate limit are discarded and a 429 is returned "
+          + "to the client.";
+  public static final String PRODUCE_MAX_REQUESTS_GLOBAL_PER_SECOND_DEFAULT = "10000";
+  public static final ConfigDef.Range PRODUCE_MAX_REQUESTS_GLOBAL_PER_SECOND_VALIDATOR =
+      ConfigDef.Range.between(1, Integer.MAX_VALUE);
+
   public static final String PRODUCE_MAX_BYTES_PER_SECOND =
       "api.v3.produce.rate.limit.max.bytes.per.sec";
   private static final String PRODUCE_MAX_BYTES_PER_SECOND_DOC =
       "Maximum number of bytes per second before rate limiting is enforced. "
+          + "The limit is enforced per clusterId, so the total rate limit will be "
+          + "number of clusters * api.v3.produce.rate.limit.max.bytes.per.sec. "
           + "Messages produced that exceed the rate limit are discarded and a 429 is returned "
           + "to the client.";
   public static final String PRODUCE_MAX_BYTES_PER_SECOND_DEFAULT = "10000000";
   public static final ConfigDef.Range PRODUCE_MAX_BYTES_PER_SECOND_VALIDATOR =
+      ConfigDef.Range.between(1, Integer.MAX_VALUE);
+
+  public static final String PRODUCE_MAX_BYTES_GLOBAL_PER_SECOND =
+      "api.v3.produce.rate.limit.max.bytes.global.per.sec";
+  private static final String PRODUCE_MAX_BYTES_GLOBAL_PER_SECOND_DOC =
+      "Maximum number of bytes per second before rate limiting is enforced, across all clusterIds. "
+          + "Messages produced that exceed the rate limit are discarded and a 429 is returned "
+          + "to the client.";
+  public static final String PRODUCE_MAX_BYTES_GLOBAL_PER_SECOND_DEFAULT = "10000000";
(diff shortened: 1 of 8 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -169,16 +169,39 @@
   public static final ConfigDef.Range PRODUCE_MAX_REQUESTS_PER_SECOND_VALIDATOR =
       ConfigDef.Range.between(1, Integer.MAX_VALUE);
 
+  public static final String PRODUCE_MAX_REQUESTS_GLOBAL_PER_SECOND =
+      "api.v3.produce.rate.limit.max.requests.global.per.sec";
```
dimitarndimitrov, line 173: nit: I think it's better to have the `global` segment as early as possible - I'd suggest putting it just before `max`, so the config becomes `api.v3.produce.rate.limit.global.max.requests.per.sec`. The same applies for the bytes config, and it might also be applied to the constant names (although we care less about these, as they are not part of our public interface).

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -169,16 +169,39 @@
...
+  private static final String PRODUCE_MAX_REQUESTS_GLOBAL_PER_SECOND_DOC =
+      "Maximum number of requests per second before rate limiting is enforced, across all "
+          + "clusterIds. "
```
dimitarndimitrov, line 176: nit: Maybe reword to "across the whole Kafka REST instance"? This way we won't have to change this if we change the facet across we apply the more granular rate-limiting.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -169,16 +169,39 @@
...
 
+  public static final String PRODUCE_MAX_BYTES_GLOBAL_PER_SECOND =
+      "api.v3.produce.rate.limit.max.bytes.global.per.sec";
+  private static final String PRODUCE_MAX_BYTES_GLOBAL_PER_SECOND_DOC =
+      "Maximum number of bytes per second before rate limiting is enforced, across all clusterIds. "
+          + "Messages produced that exceed the rate limit are discarded and a 429 is returned "
+          + "to the client.";
+  public static final String PRODUCE_MAX_BYTES_GLOBAL_PER_SECOND_DEFAULT = "10000000";
+  public static final ConfigDef.Range PRODUCE_MAX_BYTES_GLOBAL_PER_SECOND_VALIDATOR =
+      ConfigDef.Range.between(1, Integer.MAX_VALUE);
```
dimitarndimitrov, line 203: This means that we can support up to 2 GiB/s, right? While Kafka REST might be nowhere near that right now, it feels like too low of a ceiling. Can we move to a `Long` here?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
```diff
@@ -35,14 +37,21 @@
...
   public ProduceRateLimiters(
       @ProduceRateLimiterCount Provider<RequestRateLimiter> countLimiterProvider,
       @ProduceRateLimiterBytes Provider<RequestRateLimiter> bytesLimiterProvider,
+      @ProduceRateLimiterCountGlobal RequestRateLimiter countLimiterGlobal,
+      @ProduceRateLimiterBytesGlobal RequestRateLimiter bytesLimiterGlobal,
       @ProduceRateLimitEnabledConfig Boolean produceRateLimitEnabledConfig,
       @ProduceRateLimitCacheExpiryConfig Duration produceRateLimitCacheExpiryConfig) {
     this.rateLimitingEnabled = requireNonNull(produceRateLimitEnabledConfig);
+    this.countLimiterGlobal = requireNonNull(countLimiterGlobal);
+    this.bytesLimiterGlobal = requireNonNull(bytesLimiterGlobal);
```
dimitarndimitrov, line 53: Not something addressable as part of this change, but I'm starting to wonder - does it really make sense to use HK2 for the actual rate-limiter instance management (as opposed to the rate-limiter config management)? It makes setting up the testing mocks a bit easier, but that can be worked around (e.g. with a delegating ctor), and the price for it is a significant amount of factories and annotations.
Thoughts, @ehumber , @rigelbm ?

dimitarndimitrov (approved):
+1.

Some comments below.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
ehumber: I'm not keen on the HK2 usage here, in my view it makes thing more complicated than they need to be.  We also know that H2K adds performance costs (although I don't know how much of an impact that will have for this particular code)

The testing isn't actually very good with the injection - the hard bit of getting this change working was getting the injection working, and you can't mock that, you need to run an actual rest instance.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
ehumber: Yes, and I swapped the per cluster one over too.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
ehumber: As discussed, I'm going to leave this as is.

ehumber:
@dimitarndimitrov Can you take a look at the latest commit please and confirm you're happy.

ehumber:
> +1.
> 
> Yep, looks good to me - after #970 I am actually a fan of preferring this approach in our codebase.

Yes, that's a very good point, I'd forgotten about this fix.  Thanks for the quick review :)

PR #994, opened 2022-03-15
KEXP-280 new manager method to submit multiple ACLs for creation 

thomaskwscott:
This PR lays the ground work for a later PR to allow multiple ACLs to be created in a single request. 

It adds a new method createAcls that takes multiple Acl objects and creates them.

Tested with unit tests

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AclManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AclManager.java
@@ -51,6 +51,9 @@ public interface AclManager {
       Operation operation,
       Permission permission);
 
+  /** Creates a collection of Kafka {@link Acl acls}. */
+  CompletableFuture<Void> createAcls(String clusterId, List<Acl> acls);
+
   /**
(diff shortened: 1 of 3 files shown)


ehumber (approved):

rigelbm:
It would be good to have the API changes together with these changes. Otherwise, we risk having this piece of code unused. It will also be easier to see everything we need.

thomaskwscott:
> It would be good to have the API changes together with these changes. Otherwise, we risk having this piece of code unused. It will also be easier to see everything we need.

@rigelbm the api changes will follow in the other repo as they are not public. Do we need that PR in order to merge this one? It's a lot easier from a dev env perspective if we merge this one first.

rigelbm:
> > It would be good to have the API changes together with these changes. Otherwise, we risk having this piece of code unused. It will also be easier to see everything we need.
> 
> @rigelbm the api changes will follow in the other repo as they are not public. Do we need that PR in order to merge this one? It's a lot easier from a dev env perspective if we merge this one first.

I see. That's fine. Let's merge this one first.

PR #997, opened 2022-03-22
KREST-4979 reverse order of produce global rate limiters

ehumber:
Testing showed that putting the global rate limiter first results in reduced CPU load.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
@@ -66,7 +66,8 @@ public class ProduceRateLimiters {
     if (!rateLimitingEnabled) {
       return;
     }
-    // global rate limit first to reduce CPU usage under load
+    // Global rate limit first to reduce CPU usage under load
+    // https://confluentinc.atlassian.net/browse/KREST-4979
     countLimiterGlobal.get().rateLimit(1);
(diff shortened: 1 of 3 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
```diff
@@ -66,14 +66,13 @@ public void rateLimit(String clusterId, long requestSize) {
     if (!rateLimitingEnabled) {
       return;
     }
-
+    // global rate limit first to reduce CPU usage under load
```
dimitarndimitrov, line 69: nit: Can you please link our internal JIRA here as well?

dimitarndimitrov (approved):
+1.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
```diff
@@ -66,14 +66,14 @@ public void rateLimit(String clusterId, long requestSize) {
     if (!rateLimitingEnabled) {
       return;
     }
-
+    // Global rate limit first to reduce CPU usage under load
+    // https://confluentinc.atlassian.net/browse/KREST-4979
```
dimitarndimitrov, line 70: super-nit: In other places we are using just the KREST-XXX format: https://github.com/confluentinc/kafka-rest/search?q=KREST. Having said that though I just realized this comment can be linked to the JIRA anyway via `git blame` because you annotate commits just like me, so if you want just disregard my suggestion.

PR #1002, opened 2022-04-22
KREST-5385: Add error_code to produce responses.

ehumber:
Before this change, produce responses did not include an http status code if they were successful, only for error responses

This meant that code generated to match the openapi spec can't easily differentiate between a success and a failure response.

Ideally we'd have added in a status_code for success and failure cases, but there is an existing ErrorResponse that is used by all existing endpoints, and it seems overkill (and probably not the right thing to do for consistency) to put in place custom error handling for produce only, with a different structure

Also, using status_code implies real http statuses, whereas error_code gives us the freedom to use the more detailed 5 digit responses.

The response for a successful produce now looks like

{"error_code":200,"cluster_id":"nHQfD6V7Q_K29guDHiLdpw","topic_name":"summit","partition_id":0,"offset":6,"timestamp":"2022-04-22T14:48:47.650Z","value":{"type":"JSON","size":30}}

The error code is at the beginning to match what we do for the error response.

I thought about having a null error code response, but I think an explicit value is clearer.

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -2061,11 +2061,15 @@ components:
     ProduceResponse:
       type: object
       required:
+        - error_code
         - cluster_id
         - topic_name
         - partition_id
         - offset
       properties:
+        error_code:
+          type: integer
+          format: int32
         cluster_id:
           type: string
         topic_name:
@@ -3623,6 +3627,7 @@ components:
           schema:
             $ref: '#/components/schemas/ProduceResponse'
           example:
+            error_code: 200
             cluster_id: 'cluster-1'
             topic_name: 'topic-1'
             partition_id: 1
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -1271,10 +1271,10 @@ paths:
         Produce records to the given topic, returning delivery reports for each
                     record produced. This API can be used in streaming mode by setting "Transfer-Encoding:
                     chunked" header. For as long as the connection is kept open, the server will
-                    keep accepting records. For each record sent to the server, the server will
-                    asynchronously send back a delivery report, in the same order. The delivery reports 
-                    each have their own error_code. An error_code of 200 indicates success. Records are
-                    streamed to and from the server as Concatenated JSON. The HTTP status code will be HTTP 
+                    keep accepting records. Records are streamed to and from the server as Concatenated 
+                    JSON. For each record sent to the server, the server will
+                    asynchronously send back a delivery report, in the same order, each with its own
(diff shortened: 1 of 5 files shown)


api/v3/openapi.yaml
```diff
@@ -2061,11 +2061,15 @@ components:
     ProduceResponse:
       type: object
       required:
+        - error_code
```
rigelbm, line 2064: Should we add an explanation in the produce api section in `paths` that the users should *always* check the `error_code` field, and that `200` means no error?

kafka-rest/src/test/java/io/confluent/kafkarest/response/StreamingResponseTest.java
```diff
@@ -122,6 +124,7 @@ public void testWriteToChunkedOutput() throws IOException {
             .setTopicName("topicName")
             .setPartitionId(1)
             .setOffset(1L)
+            .setErrorCode(HttpStatus.ACCEPTED_202)
```
rigelbm, line 127: Why 202?

dimitarndimitrov (approved):
Looks good to me once exactly the comments Rigel has brought up are addressed.

kafka-rest/src/test/java/io/confluent/kafkarest/response/StreamingResponseTest.java
ehumber: I think this is a typo I didn't spot - was convinced I put a 200, sorry!

ehumber:
@rigelbm @dimitarndimitrov ready for a re-review

ehumber:
(Reminder to self - update ce-kafka-rest and api once the doc is changed here)

api/v3/openapi.yaml
```diff
@@ -1272,9 +1272,11 @@ paths:
                     record produced. This API can be used in streaming mode by setting "Transfer-Encoding:
                     chunked" header. For as long as the connection is kept open, the server will
                     keep accepting records. For each record sent to the server, the server will
-                    asynchronously send back a delivery report, in the same order. Records are
-                    streamed to and from the server as Concatenated JSON. Errors are reported per
-                    record. The HTTP status code will be HTTP 200 OK as long as the connection is successfully established.
+                    asynchronously send back a delivery report, in the same order. The delivery reports
```
dimitarndimitrov, line 1275: nit: For me (NB: a non-native speaker), this reads slightly better if the sentences here are joined:
_For each record sent to the server, the server will asynchronously send back a delivery report, in the same order, each with their own error_code._ and if _Records are streamed to and from..._ is put before that sentence and just after _...the server will keep accepting records._.

dimitarndimitrov (approved):
+1, a single nit below.

PR #1005, opened 2022-04-26
Remove all compiler warnings from kafka-rest.

rigelbm:
All the compiler warnings are due to either raw generic type usage or deprecated methods.

There are two changes in behaviour in kafka-rest:

 * `port` config is gone. That changes a bit how we create URLs to return back on responses (Consumer V2 API and V3 APIs). The preference now is 1) `advertised.listeners` (if same scheme as `<request scheme>`), 2) `host.name`:`<request port>` and 3) `<request host>`:`<request port>`. `port` has been deprecated since forever, so the change in behaviour is fine.
 * It used to be that we didn't block on consumer poll calls to Kafka. This behaviour is now deprecated in the consumer, so we now block at most for the remaining time of the timeout. I don't expect this to have any changes in external behaviour, but you never know.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -82,7 +82,7 @@ public class KafkaRestConfig extends RestConfig {
   // ensures poll is frequently needed and called
   public static final String MAX_POLL_RECORDS_VALUE = "30";
 
-  @Deprecated public static final String HOST_NAME_CONFIG = "host.name";
+  public static final String HOST_NAME_CONFIG = "host.name";
   private static final String HOST_NAME_DOC =
       "The host name used to generate absolute URLs in responses. If empty, the default canonical"
           + " hostname is used";
@@ -269,9 +269,7 @@ public class KafkaRestConfig extends RestConfig {
           + " Use 0 for no timeout";
   public static final String SIMPLE_CONSUMER_POOL_TIMEOUT_MS_DEFAULT = "1000";
 
-  // TODO: change this to "http://0.0.0.0:8082" when PORT_CONFIG is deleted.
-  private static final String KAFKAREST_LISTENERS_DEFAULT = "";
-  @Deprecated private static final int KAFKAREST_PORT_DEFAULT = 8082;
+  private static final String KAFKAREST_LISTENERS_DEFAULT = "http://0.0.0.0:8082";
 
   public static final String METRICS_JMX_PREFIX_DEFAULT_OVERRIDE = "kafka.rest";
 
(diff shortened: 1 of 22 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -269,9 +269,7 @@ public class KafkaRestConfig extends RestConfig {
           + " Use 0 for no timeout";
   public static final String SIMPLE_CONSUMER_POOL_TIMEOUT_MS_DEFAULT = "1000";
 
-  // TODO: change this to "http://0.0.0.0:8082" when PORT_CONFIG is deleted.
```
ehumber, line 272: What implications is setting the listener default to a value, rather than null, going to have on users?

rigelbm: There should be no effect other than `port` is completely ignored now. Before it was empty to give users to opportunity to use `port` (by setting `listeners` to empty).

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerReadTask.java
```diff
@@ -97,7 +97,11 @@ public void doPartialRead() {
         messages = new Vector<>();
       }
 
-      addRecords();
+      Instant now = clock.instant();
+      Duration elapsed = Duration.between(started, now);
+      Duration remaining = requestTimeout.minus(elapsed);
+
+      addRecords(remaining);
```
dimitarndimitrov, line 104: This seems suspicious. It ends up calling `KafkaConsumer.poll(Duration)` and without the check whether the duration is already negative that we do below it looks like this will just cause an `IllegalArgumentException` (see https://kafka.apache.org/30/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#poll(java.time.Duration)). My guess would be that this should not be done and will just HTTP 500 us.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -380,8 +381,8 @@ synchronized ConsumerRecord<KafkaKeyT, KafkaValueT> next() {
    * the existing consumer records if the records have not been fully consumed by client yet. Must
    * be invoked with the lock held, i.e. after startRead().
    */
-  private synchronized void getOrCreateConsumerRecords() {
-    ConsumerRecords<KafkaKeyT, KafkaValueT> polledRecords = consumer.poll(0);
+  private synchronized void getOrCreateConsumerRecords(Duration timeout) {
+    ConsumerRecords<KafkaKeyT, KafkaValueT> polledRecords = consumer.poll(timeout);
```
dimitarndimitrov, line 385: nit: See my previous comment.

PR #1006, opened 2022-04-28
Add produce record rate-limited metrics.

rigelbm:


kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -273,16 +269,21 @@ private static ProduceResponse toProduceResponse(
...
 
-  private void recordErrorMetrics(long latency) {
+  private void recordErrorMetrics(Duration latency) {
     producerMetrics.get().recordError();
     producerMetrics.get().recordRequestLatency(latency);
   }
 
+  private void recordRateLimitedMetrics(Duration latency) {
+    producerMetrics.get().recordRateLimited();
+    producerMetrics.get().recordRequestLatency(latency);
```
rigelbm, line 284: Do we want to include rate-limited requests in the latency metric?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -131,15 +127,16 @@ public void produce(
 
   private CompletableFuture<ProduceResponse> produce(
       String clusterId, String topicName, ProduceRequest request, ProduceController controller) {
+    Instant requestStart = Instant.now();
```
dimitarndimitrov, line 130: `Instant.now()` is not the best way to measure elapsed time on the same machine, as it ends up using a system clock that measures wall time (which is not monotonic). Let's use `System.nanoTime()` instead.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -131,15 +127,16 @@ public void produce(
...
 
     try {
       produceRateLimiters.rateLimit(clusterId, request.getOriginalSize());
     } catch (RateLimitExceededException e) {
+      recordRateLimitedMetrics(Duration.between(requestStart, Instant.now()));
```
dimitarndimitrov, line 135: > Do we want to include rate-limited requests in the latency metric?

I don't think so. Especially in the case here where we are effectively measuring solely the rate-limiting duration.

ehumber:
@rigelbm Can you link through to the jira this is for please.  We should do this anyway for all PRs, but atm this PR is missing any context as to why we are making the change as the description is also empty.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -186,8 +183,7 @@ private CompletableFuture<ProduceResponse> produce(
               ProduceResponse response =
                   toProduceResponse(
                       clusterId, topicName, keyFormat, keySchema, valueFormat, valueSchema, result);
-              long latency =
-                  Duration.between(requestInstant, result.getCompletionTimestamp()).toMillis();
+              Duration latency = Duration.between(requestStart, result.getCompletionTimestamp());
```
dimitarndimitrov, line 186: The completion timestamp that we create in the callback of `Producer.send(ProducerRecord, Callback)` should also be monotonic.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
ehumber: I think if we did want a latency metric for rate limited requests, then it should be a separate metric, in the same way you have done for the rate limited count.

I guess the question is, do we think we are going to suffer from increased latency in the rate limiting part of the flow.

I think we might, if it's really busy, so we probably do care about this metric too.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -89,6 +91,16 @@ final class ProducerMetrics {
   private static final String RECORD_ERROR_RATE_METRIC_DOC =
       "The average per-second number of record sends that resulted in errors.";
 
+  // rate-limiting
+  static final String RECORD_RATE_LIMITED_COUNT_WINDOWED_METRIC_NAME =
+      "record-rate-limited-count-windowed";
+  private static final String RECORD_RATE_LIMITED_COUNT_WINDOWED_METRIC_DOC =
+      "The total number of record sends that resulted in errors in the given window.";
```
dimitarndimitrov, line 98: This doesn't sound correct. I think it should be something like
_The total number of produce requests that were rate-limited in the given window._

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -89,6 +91,16 @@ final class ProducerMetrics {
...
+
+  static final String RECORD_RATE_LIMITED_RATE_METRIC_NAME = "record-rate-limited-rate";
+  private static final String RECORD_RATE_LIMITED_RATE_METRIC_DOC =
+      "The average per-second number of record sends that resulted in errors.";
```
dimitarndimitrov, line 102: Similar to above.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
ehumber, line 98: Description should be something like "The total number of record sends that were rate limited by the produce rate limiters"

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
ehumber, line 102: Description should be something like "The average per-second number of record sends that were rate limited by the produce rate limiters"

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
dimitarndimitrov: I commented out-of-thread on this - I don't think the measurement of the latency of the rate-limiting functionality is interesting. It doesn't take into account any sort of pipelining or queuing effects that might be telling.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
dimitarndimitrov: While it also doesn't exactly capture what's happening, I prefer using _produce requests_ over _record sends_ because when rate-limiting on Kafka REST's side kicks in, we never get to the `Produce.send` part.

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProducerMetricsTest.java
```diff
@@ -78,6 +79,7 @@ public void testRateMetrics() throws Exception {
         .forEach(
             n -> {
               metrics.recordError();
+              metrics.recordRateLimited();
```
ehumber, line 82: More of a thought than a requirement.... Is this test not very good..... As in, I

ehumber, line 82: Just a thought, rather than a requirement.  Is this test not very good any more, now that there are more metrics? 

I'm wondering if we should test the metrics one at a time.

I'm thinking about the case where the non-rate limiting accidentally triggers the rate limiting metric too. But then the rate limiting would have to not trigger for the test to pass....

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: @dimitarndimitrov, actually this metric is for our record-level rate-limiter. So records is more accurate than requests.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
rigelbm: DOne.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: Done.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: Done.

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProducerMetricsTest.java
rigelbm: Ack.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -94,25 +101,26 @@ final class ProducerMetrics {
...
   private final String requestSensorName;
   private final String requestSizeSensorName;
   private final String responseSensorName;
   private final String recordErrorSensorName;
+  private final String recordRateLimitedSensorName;
   private final String requestLatencySensorName;
 
-  ProducerMetrics(KafkaRestConfig config, Time time) {
-    this(config, time, ImmutableMap.of());
+  // TODO: Used in ce-kafka-rest. To be deleted.
```
dimitarndimitrov, line 113: nit: I personally think we should not burden the open-source code with details related to the internal Confluent source code. Maybe reword to something like "Not considered a public API, to be deleted once known usages are removed."

ehumber (approved):

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
rigelbm: Done.

PR #1008, opened 2022-05-05
KREST-5637 Add count based byte metrics and move tracking earlier

ehumber:
In our billing one pager, we have agreement that we will count billing metrics after rate limiting, but before validation of the produce request body.

This PR moves the produce metrics collection point slightly earlier to match that decision.

Currently the bytes metric is an Avg() - so if you send 10 messages of 67, the metric records 67.

I've added a summed count metric to the produce bytes sensor, which is what ce-kafka uses to record bytes as described here https://confluentinc.atlassian.net/wiki/spaces/KREST/pages/2777684595/Billing+metrics#Counting-the-bytes

Waiting on confirmation that this is correct, but I'm fairly sure it is, so raising the PR now.  It's simple to revisit later to swap to eg a Windowed Count or something else if needed.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
@@ -140,6 +140,8 @@ public final class ProduceAction {
       throw new StacklessCompletionException(e);
     }
 
+    recordRequestMetrics(request.getOriginalSize());
+
     Instant requestInstant = Instant.now();
     Optional<RegisteredSchema> keySchema =
         request.getKey().flatMap(key -> getSchema(topicName, /* isKey= */ true, key));
@@ -158,8 +160,6 @@ public final class ProduceAction {
             .orElse(request.getValue().flatMap(ProduceRequestData::getFormat));
     Optional<ByteString> serializedValue =
         serialize(topicName, valueFormat, valueSchema, request.getValue(), /* isKey= */ false);
-
-    recordRequestMetrics(request.getOriginalSize());
 
     CompletableFuture<ProduceResult> produceResult =
         controller.produce(
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
@@ -140,6 +140,8 @@ public final class ProduceAction {
       throw new StacklessCompletionException(e);
     }
 
+    // Request metrics are recorded before we check the validity of the message body, but after
+    // rate limiting, as these metrics are used for billing.
     recordRequestMetrics(request.getOriginalSize());
 
     Instant requestInstant = Instant.now();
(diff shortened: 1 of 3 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -67,6 +68,10 @@ final class ProducerMetrics {
   static final String REQUEST_SIZE_AVG_METRIC_NAME = "request-size-avg";
   private static final String REQUEST_SIZE_AVG_METRIC_DOC = "The average request size in bytes.";
 
+  static final String REQUEST_SIZE_WINDOWED_SUM_METRIC_NAME = "request-size-windowed-sum";
+  private static final String REQUEST_SIZE_WINDOWED_SUM_METRIC_DOC =
+      "The windowed sum of request sizes in bytes.";
+
   static final String REQUEST_COUNT_WINDOWED_METRIC_NAME = "request-count-windowed";
   private static final String REQUEST_COUNT_WINDOWED_METRIC_DOC =
       "The total number of requests sent within the given window.";
```
dimitarndimitrov, line 77: nit: I think a doc that's very close to the windowed request count below is a better fit - e.g. "The summed size in bytes of requests sent within the given window".

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -140,6 +140,8 @@ private CompletableFuture<ProduceResponse> produce(
       throw new StacklessCompletionException(e);
     }
 
+    recordRequestMetrics(request.getOriginalSize());
```
dimitarndimitrov, line 143: nit: I think this deserves a comment explaining the importance of its placement here - that it needs to happen after rate-limiting but before request validation and serialization.

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProducerMetricsTest.java
```diff
@@ -152,4 +152,23 @@ public void testWindowedCountMetrics() throws Exception {
       assertEquals(10.0, mBeanServer.getAttribute(beanNames.iterator().next(), metric));
     }
   }
+
+  @Test
+  public void testWindowedSumMetrics() throws Exception {
```
dimitarndimitrov, line 157: nit: It would be nice to also have a test (probably in `ProduceActionTest` that sets up messages one of which gets rate-limited and another of which causes an HTTP 400, and assert that the former doesn't bump the metrics while the latter does (as well as all the valid, non-rate-limited messages inbetween). That's entirely optional though.

dimitarndimitrov (approved):
+1.

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProducerMetricsTest.java
ehumber: I agree, but I'm going to leave it for now and try and get the main billing flow working 

I've raised the following so we don't forget about it https://confluentinc.atlassian.net/browse/KREST-5786

PR #1009, opened 2022-05-10
KREST-5732 Refactor to allow us to override producer metrics in ce-kafka-rest

ehumber:
We need the ability to set custom tags on produce metrics in ce-kafka-rest, that aren't applicable to metrics collected by just kafka-rest.

To do this we need to be able to override the ProducerMetrics class that is used.

The ProducerMetrics instance is created in the ProducerMetricsFactory provide() method in v3ResourcesModule.  

This change adds a ProducerMetrics Interface. This can then be implemented in kafka-rest and ce-kafka-rest, with different implementations of the ProducerMetrics - with a new ProducerMetricsFactory provide() method added in a new CloudV3ResourcesModule in ce-kafka-rest

The vast majority of the implementation of ProducerMetrics is identical between the two repositories. All we need at the moment is the ability to set tags on the metrics defined in ProduceMetrics.  So, I've put existing code in an abstract class and then the Impl in kafka-rest does very little in addition. (An alternative would be to make the AbstractProducerMetrics not abstract but not final either).

There are a number of updates to AbstractProducerMetrics compared to producer metrics. The summary is
 - the ImmutableMap of tags is now an empty HashMap
 - there is a new method to add entries to the HashMap
 - The creation of the metrics.metricInstance now sets the tags in the MetricsTemplate (without this Kafka barfs saying For 'request-rate', runtime-defined metric tags do not match the tags in the template. Runtime = [tenant] Template = [])
 - Some fields are no longer private

ce-kafka-rest changes here https://github.com/confluentinc/ce-kafka-rest/pull/314


kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetricsImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+  ProducerMetricsImpl(KafkaRestConfig config, Time time) {
+    this.metrics = createMetrics(config.getMetricsContext(), time);
+    this.jmxPrefix = config.getString(KafkaRestConfig.METRICS_JMX_PREFIX_CONFIG);
+    String sensorNamePrefix = jmxPrefix + ":" + GROUP_NAME + ":";
+    this.recordErrorSensorName = sensorNamePrefix + RECORD_ERROR_SENSOR_NAME;
+    this.requestSensorName = sensorNamePrefix + REQUEST_SENSOR_NAME;
+    this.requestLatencySensorName = sensorNamePrefix + REQUEST_LATENCY_SENSOR_NAME;
+    this.requestSizeSensorName = sensorNamePrefix + REQUEST_SIZE_SENSOR_NAME;
+    this.responseSensorName = sensorNamePrefix + RESPONSE_SENSOR_NAME;
```
dimitarndimitrov, line 31: This seems like common logic as well - shouldn't this also be part of the sensor setup?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AbstractProducerMetrics.java
```diff
@@ -0,0 +1,262 @@
...
+  static final String REQUEST_LATENCY_PCT_METRIC_PREFIX = "request-latency-";
+  private static final String REQUEST_LATENCY_PCT_METRIC_DOC = "Request latency percentiles.";
+
+  protected Metrics metrics;
+  protected String jmxPrefix;
+  protected String requestSensorName;
+  protected String requestSizeSensorName;
+  protected String responseSensorName;
+  protected String recordErrorSensorName;
+  protected String requestLatencySensorName;
```
dimitarndimitrov, line 112: One way to keep everything here `final` and to generally make the actually differing implementation as close to immutable as possible (if not entirely immutable) is to extract the polymorphic setup in `ProducerMetricsBuilder`, because this is more or less what we are ending up with.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AbstractProducerMetrics.java
```diff
@@ -0,0 +1,262 @@
...
+  protected static final Map<String, String> metricsTags = new HashMap<>();
+
+  protected void addMetricsTag(String tagName, String tagValue) {
+    metricsTags.put(tagName, tagValue);
+  }
```
dimitarndimitrov, line 117: This looks downright scary to me as it might end up as a leak source. Let's discuss in a bit more depth if we are forced to setup the tags this way and if there are any possible alternatives (e.g. keeping tags in an instance field as opposed to a class field).

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AbstractProducerMetrics.java
```diff
@@ -0,0 +1,262 @@
...
+    // request metrics
+    setupRequestSensor();
+    setupRequestSizeSensor();
+
+    // response metrics
+    setupResponseSensor();
+    setupRecordErrorSensor();
+    setupRequestLatencySensor();
+
+    log.info("Successfully registered kafka-rest produce metrics with JMX");
```
dimitarndimitrov, line 143: nit: Maybe move this out of this common registration point back to `ProducerMetricsImpl` ctor and add a similar log for the private implementation in the other repo.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AbstractProducerMetrics.java
ehumber: The producerMetrics are a singleton for kafka-rest so this would be fine.

However for ce-kafka-rest they are created per request.

You're right, this shouldn't be a static, it should be an instance variable, so I've removed the static

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -94,7 +94,7 @@ public ProduceAction(
       Provider<SchemaManager> schemaManagerProvider,
       Provider<RecordSerializer> recordSerializer,
       Provider<ProduceController> produceControllerProvider,
-      Provider<ProducerMetrics> producerMetrics,
+      ProducerMetrics producerMetrics,
```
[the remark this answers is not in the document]
ehumber: I'll admit I still don't 100% understand when to use Provider and when not to.

This change was based on it not working if I used a Provider, and also making our setup match what we do with producer https://github.com/confluentinc/kafka-rest/blob/5b869189dbdeb644b62560eff2233361bbb5f7b9/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ProduceControllerImpl.java#L38

I think it's because the Provider makes the object too late, but I can't see why this would be a problem in the code below.

Let's talk to Rigel in the morning about this.

ehumber:
Don't re-review yet, my unit tests are failing in a mvn package, but not when run solo.... investigating.

ehumber:
@dimitarndimitrov Ready for re-review.

dimitarndimitrov (commented):
@ehumber, I hope you can bear with me here, but now I have to wonder - do we need the separate interface if we have practically removed its purpose?

ehumber:
> 

Hmm, I thought the same.  I left it in as I was thinking we define the interface with which we call this class across repos.  But that feels a bit overkill.  

Happy to take it out :)

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProducerMetricsImplTest.java
```diff
@@ -171,4 +180,14 @@ public void testCumulativeSumMetrics() throws Exception {
       assertEquals(1230.0, mBeanServer.getAttribute(beanNames.iterator().next(), metric));
     }
   }
+
+  @Test
+  public void testTenantTag() throws Exception {
+    metrics.recordRequestSize(3);
```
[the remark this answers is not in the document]
ehumber: You're right, removed.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetricsImpl.java
ehumber: Do you still think this should be moved now we've pulled out the layers?  It's how it was before, and it is setting up all the variables for the class (like a constructor) so I think it should stay as is, but no strong feelings.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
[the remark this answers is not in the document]
ehumber: For reference, this is the exception I get when I use a provider:

standalone-cloud-kafka-rest | java.lang.IllegalStateException: Not inside a request scope.
standalone-cloud-kafka-rest |   at org.glassfish.jersey.internal.guava.Preconditions.checkState(Preconditions.java:169)
standalone-cloud-kafka-rest |   at org.glassfish.jersey.process.internal.RequestScope.current(RequestScope.java:153)
standalone-cloud-kafka-rest |   at org.glassfish.jersey.inject.hk2.RequestContext.findOrCreate(RequestContext.java:55)
standalone-cloud-kafka-rest |   at org.jvnet.hk2.internal.Utilities.createService(Utilities.java:2102)
standalone-cloud-kafka-rest |   at org.jvnet.hk2.internal.ServiceLocatorImpl.internalGetService(ServiceLocatorImpl.java:758)
standalone-cloud-kafka-rest |   at org.jvnet.hk2.internal.ServiceLocatorImpl.getUnqualifiedService(ServiceLocatorImpl.java:770)
standalone-cloud-kafka-rest |   at org.jvnet.hk2.internal.IterableProviderImpl.get(IterableProviderImpl.java:87)
standalone-cloud-kafka-rest |   at io.confluent.kafkarest.resources.v3.ProduceAction.recordResponseMetrics(ProduceAction.java:281)
standalone-cloud-kafka-rest |   at io.confluent.kafkarest.resources.v3.ProduceAction.lambda$produce$8(ProduceAction.java:194)
standalone-cloud-kafka-rest |   at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:642)
standalone-cloud-kafka-rest |   at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:478)
standalone-cloud-kafka-rest |   at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
standalone-cloud-kafka-rest |   at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
standalone-cloud-kafka-rest |   at java.base/java.lang.Thread.run(Thread.java:829)


The record metrics call is done just after we create the ProduceResponse, but before we return it, so I'm a bit surprised this is out of scope of the request.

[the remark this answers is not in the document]
rigelbm: This exception happens because when you call `Provider#get`, that's when the injection happen. For you to be able to inject something `RequestScoped`, you need to be in the request-scope, i.e. running on the request handling thread. If you follow the stack trace, you will notice that you are not in fact running on the request handling thread. You are running it from an async callback.

When do you need a `Provider`? Whenever you need to inject something that is `RequestScoped`. Since injection happens when the constructor of the resource (e.g. `new ProduceAction(...)`) is called, the request scope is not available yet. So injecting a `RequestScoped` thing without a provider in the constructor will do one of two things: 1. inject the wrong thing or 2. fail because it can't even figure out how to inject the wrong thing. That's we use the `Provider`, so that we delay injection until the resource method (e.g. `void produce(...)`) is called, since by then, the request scope is already set.

Why does not `ProducerControllerImpl` inject a Provider of the Producer? Well, because we inject a `Provider<ProducerController>` here in `ProduceAction`. By the time we instantiate `ProducerControllerImpl`, when calling `Provider<ProducerController>#get`, we are already in the request scope, so the right `Producer` is injected. In other words: When `new ProducerControllerImpl` is called, we KNOW we are in the request scope. That's not true for when `new ProducerAction` is called.

What about `ProducerMetrics`? It doesn't have any request scoped state, neither it injects anything request scoped, so it should really be a Singleton, no RequestScoped. I think by not using a Provider, you are injecting the "wrong thing", which turns out to be okay since you never needed a request scoped `ProducerMetrics` anyway.

That would be the end of it if it weren't for your changes in `ce-kafka-rest`. There, we a binding a `RequestScoped` implementation because we need to. The `ProducerMetrics` there needs request scoped information, e.g. the cluster ID, etc. If you try to inject that one here, it will fail because this information is not available. Your tests probably succeeded because in the current PR you are not actually using any request scoped information, but you intend to.

TL;DR: We still need a provider here. Make sure to call `Provider#get` in the request thread (for an example, see where we call the get() for the other providers). Then make sure the ProducerMetrics here in `kafka-rest` is bound as a `Singleton`. This one here doesn't not need to be `RequestScoped`.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -111,8 +111,14 @@ final class ProducerMetrics {
   private final String responseSensorName;
   private final String recordErrorSensorName;
   private final String requestLatencySensorName;
+  private final Map<String, String> metricsTags;
```
rigelbm, line 114: I know it's cumbersome to pass this down to `createMetrics` all the way to `getMetricName`, but this variable is not actually used after the constructor exits, so it shouldn't really be available. Please, pass it down instead. :)

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -237,26 +244,26 @@ private void addPercentiles(
 
   private MetricName getMetricName(String name, String doc) {
     return metrics.metricInstance(
-        new MetricNameTemplate(name, GROUP_NAME, doc, Collections.emptySet()), METRIC_TAGS);
+        new MetricNameTemplate(name, GROUP_NAME, doc, metricsTags.keySet()), metricsTags);
   }
 
-  void recordResponse() {
+  public void recordResponse() {
```
rigelbm, line 250: Why did this and the others below become `public`?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -111,8 +111,14 @@ final class ProducerMetrics {
   private final String responseSensorName;
   private final String recordErrorSensorName;
   private final String requestLatencySensorName;
+  private final Map<String, String> metricsTags;
 
   ProducerMetrics(KafkaRestConfig config, Time time) {
+    this(config, time, ImmutableMap.of());
+  }
+
+  ProducerMetrics(KafkaRestConfig config, Time time, Map<String, String> metricsTags) {
```
rigelbm, line 120: If all the difference between this and the one we bind in `ce-kafka-rest` is the `metricsTags`, we should just bind a different `metricsTags` instead there, and inject it here. Food for thought.

rigelbm: Never mind. Let's default to bind less things instead of more. It's not like HK2 is super nice anyway.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
ehumber: Left over from when we had the interface and various impl classes.  I thought I'd pulled these out already!

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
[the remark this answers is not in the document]
ehumber: done

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
ehumber: I'm not entirely sure I agree - metricsTags is a property of the ProduerMetrics in the same way that the Name fields are. I've typed up the change now though, so take a look.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -111,8 +111,14 @@ final class ProducerMetrics {
   private final String responseSensorName;
   private final String recordErrorSensorName;
   private final String requestLatencySensorName;
+  private final Map<String, String> metricsTags;
 
   ProducerMetrics(KafkaRestConfig config, Time time) {
+    this(config, time, ImmutableMap.of());
+  }
```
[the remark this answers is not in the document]
ehumber: Leaving as is, I prefer the two because there is a tiny risk someone is relying on the original constructor.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
```diff
@@ -84,11 +84,13 @@ public final class ProduceAction {
   private final Provider<SchemaManager> schemaManagerProvider;
   private final Provider<RecordSerializer> recordSerializerProvider;
   private final Provider<ProduceController> produceControllerProvider;
-  private final Provider<ProducerMetrics> producerMetrics;
+  private final Provider<ProducerMetrics> producerMetricsProvider;
   private final StreamingResponseFactory streamingResponseFactory;
   private final ProduceRateLimiters produceRateLimiters;
   private final ExecutorService executorService;
 
+  private ProducerMetrics metrics = null;
```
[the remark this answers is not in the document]
ehumber: I did this to start with and it barfed. But, that's because I passed down the provider, not the got ProducerMetrics, which is what I do now.

@rigelbm ready for re-review.

dimitarndimitrov (approved):
+1.

PR #1012, opened 2022-05-11
KREST-5637 Move to cumulative sum for billing metrics

ehumber:
Move to cumulative sum for billing metrics

https://confluent.slack.com/archives/C0NL3SPNK/p1652191966280159?thread_ts=1652188859.047919&cid=C0NL3SPNK

Also update the name to be something more friendly to end users, who may use this metric name to check their billing is accurate.

request-bytes is what kafka uses, so copying that

https://confluentinc.atlassian.net/wiki/spaces/CS/pages/1307607876/Billing+Metrics


dimitarndimitrov (approved):
+1.

PR #1014, opened 2022-05-13
KREST-5830: Create a lazy wrapper around MappingIterator for Produce Action.

rigelbm:
JsonParser instantiation, which happens during MappingIterator deserialization, can throw on bad request. This can cause the request scope to not be disposed properly if a `@RequestScoped` resource is injected into a filter before an async resource method is executed. For Kafka REST, this only happens for the Produce Action.

This PR wraps the MappingIterator, so that the JsonParser is only instantiated during the first hasNext/nextValue call.

The PR also adds a test that shows the aforementioned resource leak.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStream.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStream.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2022 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.response;
+
+import com.fasterxml.jackson.databind.MappingIterator;
+import com.google.common.base.Suppliers;
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.NoSuchElementException;
+import java.util.function.Supplier;
+
+/**
+ * A lazy wrapper over {@link MappingIterator} to delay {@link
+ * com.fasterxml.jackson.core.JsonParser} until the first read.
+ */
+public final class JsonStream<T> implements Closeable {
+  private final Supplier<MappingIterator<T>> delegate;
+
+  public JsonStream(Supplier<MappingIterator<T>> delegate) {
+    this.delegate = Suppliers.memoize(delegate::get);
+  }
+
+  public boolean hasNext() {
+    if (delegate.get() == null) {
+      return false;
+    }
+    return delegate.get().hasNext();
+  }
+
+  public T nextValue() throws IOException {
+    if (delegate.get() == null) {
+      throw new NoSuchElementException();
+    }
+    return delegate.get().nextValue();
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (delegate.get() != null) {
+      delegate.get().close();
+    }
+  }
+}
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2022 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest.response;
+
+import static java.util.Objects.requireNonNull;
+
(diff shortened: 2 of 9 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStream.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.NoSuchElementException;
+import java.util.function.Supplier;
+
+/**
+ * A lazy wrapper over {@link MappingIterator} to delay {@link
+ * com.fasterxml.jackson.core.JsonParser} until the first read.
+ */
+public final class JsonStream<T> implements Closeable {
```
ehumber, line 29: Should this also implement Iterator, as MappingIterator does?

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProduceActionTest.java
```diff
@@ -556,14 +565,13 @@ public void testHasNextOnNullData() throws Exception {
...
             bytesLimiterGlobalProvider);
-    MappingIterator<ProduceRequest> requests = null;
 
     FakeAsyncResponse fakeAsyncResponse = new FakeAsyncResponse();
 
     RestConstraintViolationException e =
         assertThrows(
             RestConstraintViolationException.class,
-            () -> produceAction.produce(fakeAsyncResponse, "clusterId", "topicName", requests));
+            () -> produceAction.produce(fakeAsyncResponse, "clusterId", "topicName", null));
```
ehumber, line 574: Should the null here be `new JsonStream<>(() -> null)` ?

kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
```diff
@@ -0,0 +1,71 @@
...
+  public JsonStream<?> readFrom(
+      Class<JsonStream<?>> type,
+      Type genericType,
+      Annotation[] annotations,
+      MediaType mediaType,
+      MultivaluedMap<String, String> httpHeaders,
+      InputStream entityStream)
+      throws WebApplicationException {
+    JavaType wrappedType = objectMapper.constructType(genericType).containedType(0);
+    return new JsonStream<>(
```
ehumber, line 57: I think I'm missing a concept that's stopping me from reviewing this properly.

How does the returned `() -> objectMapper.readValues` end up being the JsonStream (ie lazy wrapped MappingIterator) that ends up being injected into ProduceAction?

I get that ObjectMapper.readValues returns the right type (based on https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectMapper.html#readValues(com.fasterxml.jackson.core.JsonParser,%20java.lang.Class))

But, I don't understand how this JsonStreamMessageBodyReader overwrites the default ObjectMapper that I am theorizing would have previously returned the not-lazy MappingIterator to ProduceAction.

I know it's `config.register(new JsonStreamMessageBodyReader(getJsonMapper()));` but I don't know why.

Can you help me understand please.

kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStream.java
rigelbm: We never use is as an Iterator, or will ever do it, so I don't think so.

kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
```diff
@@ -0,0 +1,71 @@
...
+  @Override
+  public JsonStream<?> readFrom(
+      Class<JsonStream<?>> type,
+      Type genericType,
+      Annotation[] annotations,
+      MediaType mediaType,
+      MultivaluedMap<String, String> httpHeaders,
+      InputStream entityStream)
+      throws WebApplicationException {
+    JavaType wrappedType = objectMapper.constructType(genericType).containedType(0);
```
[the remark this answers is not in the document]
rigelbm: The `genericType` is always `JsonStream<T>`. `containedType(0)` returns `T`.

kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
rigelbm: > But, I don't understand how this JsonStreamMessageBodyReader overwrites the default ObjectMapper that I am theorizing would have previously returned the not-lazy MappingIterator to ProduceAction.

It does not. It specifically implements deserialization for `JsonStream` only, which is not a JSON object, therefore not handled by Jackson. The `isReadable` method tells what this `MessageBodyReader` can deserialize.

ehumber: But how does it know to use this MessageBodyReader, that can deserliaze JsonStream?

The user POSTs to  .../records to produce.

They include some bytes as the data, and the content-type application/json.

The ProduceAction.produce(...) method is called, one of the fields of which is `JsonStream<ProduceRequest> requests)`

How does the code know to inject the field as type JsonStream<ProduceRequests>, I can't see a bind of JsonStream<ProduceRequests> to anything, so I don't understand how the code knows how to make this type of object.  

I can see that the JsonStreamMessageBodyReader returns this type of object when it does a readFrom, which I'm guessing is what populates the `JsonStream<ProduceRequests>` in the ProduceAction.produce() call.  

But I can't see how the code links from the POST to ...../records, and some bytes arriving, into creating a JsonStreamMessageBodyReader (as opposed to some other type of MessageBodyReader).  And how does it know that if a @POST tagged method call contains an arbitrary type of eg `JsonStream<ProduceRequest> requests)` that it needs to use this JsonStreamMessageBodyReader to parse those bytes?

rigelbm: Jersey goes over all MessageBodyReaders registered asking isReadable for the type and content-type, at start-up time.

ehumber (approved):

PR #1016, opened 2022-05-18
KREST-5096 Fix NPE in Topics Resource when no key, or no value specified

ehumber:
We need to null check the record.getJsonKey() call before calling any methods on it.

5.3.x still passes, the jackson version there is 2.10.5, and the jackson version at 5.4.x is 2.13.2.

I can only assume something has changed between these Jackson versions that has altered how null objects are dealt with (ie really null, rather than returning a json literal null object), possibly as part of  RFC 7159.  However I've not managed to track down the change.

I also had a look to see if there was an annotation I could set on the fields in the AvroProducerRecord to change how it behaves with nulls, but no joy.

Have fixed for all the cases I could find of .isNull being called without a null check.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/entities/AvroProduceRecord.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/entities/AvroProduceRecord.java
@@ -70,4 +70,12 @@ public class AvroProduceRecord extends ProduceRecordBase<JsonNode, JsonNode> {
     result = 31 * result + (value != null ? value.hashCode() : 0);
     return result;
   }
+
+  public boolean hasKey() {
+    return this.getJsonKey() != null && !this.getJsonKey().isNull();
+  }
+
+  public boolean hasValue() {
+    return this.getJsonValue() != null && !this.getJsonValue().isNull();
+  }
 }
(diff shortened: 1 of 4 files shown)


rigelbm (approved):
LGTM

kafka-rest/src/main/java/io/confluent/kafkarest/entities/AvroProduceRecord.java
```diff
@@ -70,4 +70,12 @@ public int hashCode() {
     result = 31 * result + (value != null ? value.hashCode() : 0);
     return result;
   }
+
+  public boolean hasKey() {
+    return this.getJsonKey() != null && !this.getJsonKey().isNull();
```
dimitarndimitrov, line 75: It might be a good idea to hoist the result of `getJsonKey()` into a local variable in case that method actually has to do some costly parsing. The same applies to `getJsonValue()` below. JIT might eventually do that for you, but the default threshold for that is still pretty high.

dimitarndimitrov (approved):
+1.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/AvroProduceRecord.java
ehumber: it's just a getter and a setter on a locally set variable, there is no parsing.  I could just do key !=null and key.isNull instead, on the already existing key (and value)?

PR #832, opened 2021-04-08
KREST-46: Removed property ${confluent.maven.repo}

javabrett:
This property was used in repository/url definition.  This causes a problem with resolution of the parent project/pom, since
property-evaluation occurs in Maven after the parent project is resolved.

`Dockerfile` which can be used to test can't/can build-from-source (without external repository configuration):

```
FROM maven:3.6.3-openjdk-8

RUN useradd -m build
USER build
COPY --chown=build . /home/build/src
WORKDIR /home/build/src

RUN mvn package -U
```

Prior to this change:

```
 > [5/5] RUN mvn package -U:
#7 1.995 [INFO] Scanning for projects...
#7 2.343 Downloading from central: https://repo.maven.apache.org/maven2/io/confluent/rest-utils-parent/5.0.0/rest-utils-parent-5.0.0.pom
#7 3.749 [ERROR] [ERROR] Some problems were encountered while processing the POMs:
#7 3.749 [FATAL] Non-resolvable parent POM for io.confluent:kafka-rest-parent:5.0.0: Could not transfer artifact io.confluent:rest-utils-parent:pom:5.0.0 from/to confluent (${confluent.maven.repo}): Cannot access ${confluent.maven.repo} with type default using the available connector factories: BasicRepositoryConnectorFactory and 'parent.relativePath' points at wrong local POM @ line 7, column 13
#7 3.756  @
#7 3.757 [ERROR] The build could not read 1 project -> [Help 1]
#7 3.758 [ERROR]
#7 3.758 [ERROR]   The project io.confluent:kafka-rest-parent:5.0.0 (/home/build/src/pom.xml) has 1 error
#7 3.758 [ERROR]     Non-resolvable parent POM for io.confluent:kafka-rest-parent:5.0.0: Could not transfer artifact io.confluent:rest-utils-parent:pom:5.0.0 from/to confluent (${confluent.maven.repo}): Cannot access ${confluent.maven.repo} with type default using the available connector factories: BasicRepositoryConnectorFactory and 'parent.relativePath' points at wrong local POM @ line 7, column 13: Cannot access ${confluent.maven.repo} using the registered transporter factories: WagonTransporterFactory: Unsupported transport protocol -> [Help 2]
#7 3.759 [ERROR]
#7 3.761 [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
#7 3.762 [ERROR] Re-run Maven using the -X switch to enable full debug logging.
#7 3.762 [ERROR]
[...]

PR #1020, opened 2022-05-24
KREST-6005 Use HTTPS for public Maven repo URL

dimitarndimitrov:
Tested via @javabrett's Dockerfile from #832.

One thing I've noted though - apparently `6.1.1-post` got clobbered, and the #832 changes are again unavailable there (and the aforementioned Dockerfile can again be used to reproduce the problem Brett originally fixed).
I'm not sure how or why did that happen (my guess would be some sort of an automated update process) - I will focus specifically on the HTTP/HTTPS issue here.

@javabrett feel free to file an internal JIRA for looking into the clobbering and eventually reapplying your fix to `6.1.1-post`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
         <repository>
             <id>confluent</id>
             <name>Confluent</name>
-            <url>http://packages.confluent.io/maven/</url>
+            <url>https://packages.confluent.io/maven/</url>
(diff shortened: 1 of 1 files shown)


jonchiu (approved):
lgtm

ehumber (approved):

PR #1023, opened 2022-06-06
KREST-5637 Use a Meter as recommended by observability and the right Metrics Object

ehumber:
Our metrics weren't showing up in the telemetry pipeline, so obeservability suggested we use the same pattern for creating Metrics as the MultiTenantInterceptor does, which is a Meter.

I've basically copied the code over.

Even with this change the metrics didn't get added to the telemetry pipeline.  This was because each time a produce request arrived, we made a new Metrics object.  These are only registered with Telemetry Reporter on its startup, and so the new produce metrics were never seen by the Telemetry code.

We should be using a single shared Metrics object for the whole of kafka-rest, not making a new one each time we add a new producer (this matches Kafka, which uses a single Metrics for all the server code).

The Metrics is defined in rest-utils, so I grab it from the Application and add it to the config object (which is where we already store the metrics tags) and pass it to ProducerMetrics from there.

image


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -454,6 +455,7 @@ public class KafkaRestConfig extends RestConfig {
   private static final String RATE_LIMIT_COSTS_DEFAULT = "";
 
   private static final ConfigDef config;
+  private Metrics metrics;
```
dimitarndimitrov, line 458: I think this needs to be `volatile` as the metrics seem to be written and read in a concurrent fashion (written in the app initialization, read whenever we initialize produce metrics).

dimitarndimitrov (approved):
+1.

PR #1027, opened 2022-06-09
APPSEC-1225: Exclude org.slf4j:slf4j-reload4j to avoid multile binding in Kafka

xjin-Confluent:


ijuma:
Someone from kafka rest should review this. Please make sure this works correctly in both standalone and within confluent server.

kafka-rest/src/assembly/package.xml
```diff
@@ -47,14 +47,15 @@
             <excludes>
                 <exclude>io.confluent:rest-utils</exclude>
                 <exclude>org.slf4j:slf4j-log4j12</exclude>
```
dimitarndimitrov, line 49: not-a-comment-to-be-addressed: In theory isn't this a redundant leftover? I would expect `slf4j-log4j12` to be "extinct" from our dependencies no matter what packaging is being used. In practice it probably makes sense to keep this exclusion until it can be proven to be redundant.

dimitarndimitrov (approved):
+1.
This looks good to me. @rigelbm can you please take a look as well, as the author of the original bin/lib split?

kafka-rest/src/assembly/package.xml
xjin-Confluent: Yeah, I think it is safe to keep it here for now.

PR #1034, opened 2022-06-20
Handle authz errors from Schema Registry explicitly.

rigelbm:
Pre-7.2.x, authorization errors from Schema Registry would result in a `408 Request Timeout` response. Now we are getting `422 Unprocessable Entity`. Neither error code is correct.

If we get a authorization error from Schema Registry, we should send that to the end-user. This has been unspecified behaviour so far. This PR makes this explicit by singling out the unauthorized error from Schema Registry.


ehumber (approved):

dimitarndimitrov (approved):
+1.

_For posterity, although this might be useful all the way back to `7.0.x`, it is landing only in the `7.2.x` and above branches in order to not change the unspecified behaviour that users might have adapted to._

_In contrast, for https://github.com/confluentinc/kafka-rest/pull/963, we did change the unspecified behaviour in older branches as well, as that behaviour was considered potentially dangerous (and we did get reports of that old behaviour affecting users)._

Issue #432, opened 2018-05-10
Consumer call needs to happen twice to receive recently produced messages


aaronjwhiteside:
The following is the "on the wire" interaction between our REST client and the Kafka REST Proxy.

A summary of the operations performed is:
1. Start the various docker images in the right order (wait for them to finish starting up)
2. Produce a message to the topic "pubsub"
3. Create consumer
4. Subscribe consumer to the topic "pubsub"
5. Try and consume, but receive nothing. (Note: increase or defaulting the timeout value has no effect other than how long we wait until we try and consume again)
6. Immediately try and consume again, this time we actually receive the message.

We have tried creating and subscribing the consumer before producing the message then trying to consume, but it makes no difference, in all cases we have tried we need to try and consume the topic twice to receive the message on the second attempt.

I can provide the output from all three docker containers if required.

```
2018-05-09 20:41:33.957 DEBUG 21787 --- [           main] o.a.h.wire                               : http-outgoing-0 >> "POST /topics/pubsub HTTP/1.1[\r][\n]"
2018-05-09 20:41:33.957 DEBUG 21787 --- [           main] o.a.h.wire                               : http-outgoing-0 >> "Accept: application/vnd.kafka.v2+json[\r][\n]"
2018-05-09 20:41:33.957 DEBUG 21787 --- [           main] o.a.h.wire                               : http-outgoing-0 >> "User-Agent: KafkaRestProxyClient/1.0-SNAPSHOT Jersey/2.26 Apache-HttpClient/4.5.5 Java/1.8.0_141[\r][\n]"
2018-05-09 20:41:33.957 DEBUG 21787 --- [           main] o.a.h.wire                               : http-outgoing-0 >> "Content-Type: application/vnd.kafka.binary.v2+json[\r][\n]"
2018-05-09 20:41:33.957 DEBUG 21787 --- [           main] o.a.h.wire                               : http-outgoing-0 >> "Transfer-Encoding: chunked[\r][\n]"
2018-05-09 20:41:33.957 DEBUG 21787 --- [           main] o.a.h.wire                               : http-outgoing-0 >> "Host: localhost:54987[\r][\n]"
2018-05-09 20:41:33.958 DEBUG 21787 --- [           main] o.a.h.wire                               : http-outgoing-0 >> "Connection: Keep-Alive[\r][\n]"
[...]

PR #1035, opened 2022-06-21
replaced deprecated method

dennisgermany:
Hi, just replaced the deprecated method with the new one. Best regards :)

Originally we were working on: #432

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
@@ -381,7 +381,7 @@ public abstract class KafkaConsumerState<KafkaKeyT, KafkaValueT, ClientKeyT, Cli
    * be invoked with the lock held, i.e. after startRead().
    */
   private synchronized void getOrCreateConsumerRecords() {
-    ConsumerRecords<KafkaKeyT, KafkaValueT> polledRecords = consumer.poll(0);
+    ConsumerRecords<KafkaKeyT, KafkaValueT> polledRecords = consumer.poll(Duration.ZERO);
     // drain the iterator and buffer to list
     for (ConsumerRecord<KafkaKeyT, KafkaValueT> consumerRecord : polledRecords) {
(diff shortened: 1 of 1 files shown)


ehumber:
Hi @dennisgermany Thanks for making this update.

There are some changes of behaviour that come with the move from poll(long) to poll(Duration), as described in KIP-266

https://cwiki.apache.org/confluence/display/KAFKA/KIP-266%3A+Fix+consumer+indefinite+blocking+behavior

We'll need a little time to consider the implications of these before approving the PR.

While this review is done, would you mind sharing what testing you did for the code change.

dennisgermany:
Hi!

> There are some changes of behaviour that come with the move from poll(long) to poll(Duration), as described in KIP-266
> 
> https://cwiki.apache.org/confluence/display/KAFKA/KIP-266%3A+Fix+consumer+indefinite+blocking+behavior
> 

That's a good info. Actually we weren't aware of that matter.

> While this review is done, would you mind sharing what testing you did for the code change.

we used a local kafka-rest-proxy and connected it to a remote kafka. Afterwards we used the rest-api to check that consuming messages was still possible

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
```diff
@@ -381,7 +381,7 @@ synchronized ConsumerRecord<KafkaKeyT, KafkaValueT> next() {
    * be invoked with the lock held, i.e. after startRead().
    */
   private synchronized void getOrCreateConsumerRecords() {
-    ConsumerRecords<KafkaKeyT, KafkaValueT> polledRecords = consumer.poll(0);
+    ConsumerRecords<KafkaKeyT, KafkaValueT> polledRecords = consumer.poll(Duration.ZERO);
```
hachikuji, line 384: The only difference between these two methods is that `poll(long)` will block when the consumer group is rebalancing. It's worth checking that the code doesn't depend on this blocking in some way if you haven't already.

ehumber (approved):

PR #1042, opened 2022-07-15
MINOR: Refactor schema APIs

rayokota:


ehumber:
Hi @rayokota 

Thanks for the update.  Can I get a link to a jira (or an updated PR description) that has some context as to why the change is needed.  It seems OK, but without context it's hard to review properly.

Thanks
Emma

rayokota:
@ehumber , we'll be adding a number of new fields to schema-related objects.  This change is to proactively pass the entire schema object instead of its constituent parts.  See https://confluentinc.atlassian.net/wiki/spaces/DG/pages/2792194049/Enhancements+to+the+Confluent+Schema+Format+WIP

rayokota:
Test errors seem unrelated

PR #1043, opened 2022-07-18
KREST-2746 Reflect slow responses from Kafka back to the http client

ehumber:
This PR adds code to reflect Kafka rate limiting back to the http client.

While the produce api response from Kafka does indicate that the client is being throttled to the Kafka Produce java client, this information is not exposed to the end user of that client (ie Kafka REST), so we have to infer the throttling another way.

Kafka throttles by delaying its response back to the client.  So we can assume that if we are getting a growing backlog of requests waiting to be sent to kafka then kafka is throttling REST.

When this happens (or we have a backlog of calls to Kafka for some other reason) we are obliged to send responses back to kafka in the same order as the requests arrived with us, so all we can do is add 429 responses, once we hit the "throttle from this point "queue depth, to the end of the response queue.  The requests for these are not sent to kafka, and so should reduce the traffic reaching kafka, possibly enough to bring it back under the throttle limit.

If the queue depth doesn't shrink sufficiently, and instead grows to the max limit, then the connection is chopped after the grace period in my previous (disconnect clients nicely) PR has expired.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
@@ -73,6 +73,7 @@ public abstract class StreamingResponse<T> {
 
   private static final Logger log = LoggerFactory.getLogger(StreamingResponse.class);
   private static final int ONE_SECOND_MS = 1000;
+  private static final int MAX_CLOSE_RETRIES = 2;
 
   private static final CompositeErrorMapper EXCEPTION_MAPPER =
       new CompositeErrorMapper.Builder()
@@ -188,8 +189,8 @@ public abstract class StreamingResponse<T> {
    * <p>This method will block until all requests are read in. The responses are computed and
    * written to {@code asyncResponse} asynchronously.
    */
-  public final void resume(AsyncResponse asyncResponse) throws InterruptedException {
-    log.error("Resuming StreamingResponse - about to call new async response queue");
+  public final void resume(AsyncResponse asyncResponse) {
+    log.debug("Resuming StreamingResponse");
     AsyncResponseQueue responseQueue = new AsyncResponseQueue(chunkedOutputFactory);
     responseQueue.asyncResume(asyncResponse);
     ScheduledExecutorService executorService = null;
@@ -210,32 +211,29 @@ public abstract class StreamingResponse<T> {
         }
       }
     } catch (Exception e) {
-      log.error("Exception thrown when processing stream ", e);
+      log.debug("Exception thrown when processing stream ", e);
       responseQueue.push(
           CompletableFuture.completedFuture(
               ResultOrError.error(EXCEPTION_MAPPER.toErrorResponse(e))));
     } finally {
-      System.out.println(
-          "why isfinally triggered closing started"
-              + closingStarted); // can't call hasnext because of the mocking
       // if there are still outstanding response to send back, for example hasNext has returned
       // false, but the executorService hasn't triggered yet, give everything a chance to
-      // complete within the grace period if there are still outstanding responses.
-      System.out.println("FINALLY before tail length wait");
-      if (responseQueue.getTailLength().get() > 0) {
-        System.out.println("in tail length wait");
-        Thread.sleep(gracePeriod.toMillis()); // TODO make this smarter, grace period could be HUGE
+      // complete within the grace period plus a little bit.
+      int retries = 0;
+      while (!closingFinished && retries < MAX_CLOSE_RETRIES) {
+        retries++;
+        try {
+          Thread.sleep(gracePeriod.toMillis() / retries);
+        } catch (InterruptedException e) {
+          // do nothing
+        }
       }
-      System.out.println("FINALL calling a close " + this.getClass());
-      //if (!closingFinished) {
-        close();
-        System.out.println("FINALLY calling responseQueue close" + this.getClass());
-        responseQueue.close();
-      //}
+      close();
+      responseQueue.close();
       if (executorService != null) {
         executorService.shutdown();
         try {
-          if (!executorService.awaitTermination(ONE_SECOND_MS, TimeUnit.MILLISECONDS)) {
+          if (!executorService.awaitTermination(gracePeriod.toMillis(), TimeUnit.MILLISECONDS)) {
             executorService.shutdownNow();
           }
         } catch (InterruptedException e) {
@@ -243,6 +241,20 @@ public abstract class StreamingResponse<T> {
         }
       }
     }
+  }
+
+  private void handleOverMaxStreamDuration(
+      ScheduledExecutorService executorService, AsyncResponseQueue responseQueue) {
+    triggerDelayedClose(executorService, responseQueue);
+    next();
+    responseQueue.push(
+        CompletableFuture.completedFuture(
+            ResultOrError.error(
+                EXCEPTION_MAPPER.toErrorResponse(
+                    new StatusCodeException(
+                        Status.REQUEST_TIMEOUT,
+                        "Streaming connection open for longer than allowed",
+                        "Connection will be closed.")))));
   }
 
   private void handleResponseQueueDepthTooLarge(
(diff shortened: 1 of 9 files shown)


ehumber:
@dimitarndimitrov ignore the first commit, I'd not tidied up properly.  This one shows the hang behaviour from easymock where the close on the mappingIterator never returns:

logs look like

```
** has next delayed response
THREAD EXECUTOR TRIGGERSclass io.confluent.kafkarest.response.StreamingResponse$ComposingStreamingResponse
THREAD EXECUTOR closeAll from thread executor call through to a sub close class io.confluent.kafkarest.response.StreamingResponse$ComposingStreamingResponse
CLOSE could be from either THREAD or FINALLY ->  calls through to inputStreaming closeclass io.confluent.kafkarest.response.StreamingResponse$ComposingStreamingResponse
CLOSE could be either from THREAD or FINALLY -> calls through to inputStreaming closeclass io.confluent.kafkarest.response.StreamingResponse$InputStreamingResponse
CLOSE in json stream which calls mapping iterator close
delegate not nullclass com.fasterxml.jackson.databind.MappingIterator$$EnhancerByCGLIB$$4a178d4f

Process finished with exit code 130 (interrupted by signal 2: SIGINT)
```

ehumber:
and with the debugger running it looks like this, and passes the test.  

There is a difference (after I said there wasn't)  :) 

the debug line writing to sink. 

This only writes out if I have a breakpoint in the async writing to sink method (eg line 471), otherwise I see the same behaviour with and without the debugger 

```
THREAD EXECUTOR TRIGGERSclass io.confluent.kafkarest.response.StreamingResponse$ComposingStreamingResponse
THREAD EXECUTOR closeAll from thread executor call through to a sub close class io.confluent.kafkarest.response.StreamingResponse$ComposingStreamingResponse
CLOSE could be from either THREAD or FINALLY ->  calls through to inputStreaming closeclass io.confluent.kafkarest.response.StreamingResponse$ComposingStreamingResponse
CLOSE could be either from THREAD or FINALLY -> calls through to inputStreaming closeclass io.confluent.kafkarest.response.StreamingResponse$InputStreamingResponse
[2022-07-18 14:51:09,785] DEBUG Writing to sink (io.confluent.kafkarest.response.StreamingResponse:477)
CLOSE in json stream which calls mapping iterator close
delegate not nullclass com.fasterxml.jackson.databind.MappingIterator$$EnhancerByCGLIB$$4a178d4f
delegate now closed
THREAD EXECUTOR closeAll from thread executor calling responsequeue.close
CLOSE response queue from FINALLY or THREAD class io.confluent.kafkarest.response.StreamingResponse$AsyncResponseQueue

```

So that's a nice big clue I'm going to go investigate :)

ehumber:
@dimitarndimitrov @AndrewJSchofield This PR could do with merging before mid-November if possible, so that we can deal with kafka back pressure before increasing (or removing) any byte based rate limits for produce. 

It would be great if I could get some feedback by the end of October.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -196,12 +216,24 @@ public final void resume(AsyncResponse asyncResponse) {
           CompletableFuture.completedFuture(
               ResultOrError.error(EXCEPTION_MAPPER.toErrorResponse(e))));
     } finally {
+      // if there are still outstanding response to send back, for example hasNext has returned
```
ehumber, line 219: Remove the MAX_CLOSE_RETRIES

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -211,12 +243,47 @@ public final void resume(AsyncResponse asyncResponse) {
...
+                    new StatusCodeException(
+                        Status.TOO_MANY_REQUESTS,
+                        "Backlog of messages waiting to be sent to Kafka is too large",
+                        "Not sending to Kafka.")))));
+  }
+
+  private void triggerDelayedClose(
+      ScheduledExecutorService executorService, AsyncResponseQueue responseQueue) {
+    if (executorService == null) {
+      executorService = Executors.newSingleThreadScheduledExecutor();
```
ehumber, line 279: This is null, so it's not an object reference that we can overwrite/edit at this point.

Need to eg return the executor service and set the original object to that value

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -347,14 +438,16 @@ private void asyncResume(AsyncResponse asyncResponse) {
       asyncResponse.resume(Response.ok(sink).build());
     }
 
-    private volatile boolean sinkClosed = false;
+    volatile boolean sinkClosed = false;
```
ehumber, line 441: Why is this no longer private?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -347,14 +438,16 @@ private void asyncResume(AsyncResponse asyncResponse) {
...
+
+    private volatile AtomicInteger queueDepth = new AtomicInteger(0);
```
ehumber, line 443: Shouldn't be volatile

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProduceActionTest.java
```diff
@@ -839,7 +841,7 @@ private static ProduceAction getProduceAction(
     replay(produceControllerProvider, produceController);
 
     StreamingResponseFactory streamingResponseFactory =
-        new StreamingResponseFactory(chunkedOutputFactory, FIVE_SECONDS_MS, FIVE_SECONDS_MS);
+        new StreamingResponseFactory(chunkedOutputFactory, FIVE_SECONDS_MS, FIVE_MS, DEPTH, DEPTH);
```
ehumber, line 844: Rename these to be eg DURATION or GRACE_DURATION etc, need to go look up the values atm

kafka-rest/src/test/java/io/confluent/kafkarest/response/StreamingResponseTest.java
```diff
@@ -294,69 +300,272 @@ public void testWriteToChunkedOutputAfterTimeout() throws IOException, Interrupt
...
     expect(clock.instant())
         .andReturn(Instant.ofEpochMilli(timeout + 5)); // second message beyond timeout
-
+    expect(mockedChunkedOutput.isClosed()).andReturn(false);
     mockedChunkedOutput.write(error);
+
+    // no third message
+    expect(requestsMappingIterator.hasNext()).andReturn(false);
+
+    requestsMappingIterator.close(); // call from thread executor
```
ehumber, line 333: Can't do an expect with something that returns a void, need to do the playback of the calls one by one :'( 

Add a comment explaining this, because this looks bonkers.

kafka-rest/src/test/java/io/confluent/kafkarest/response/StreamingResponseTest.java
```diff
@@ -294,69 +300,272 @@ public void testWriteToChunkedOutputAfterTimeout() throws IOException, Interrupt
...
+    expect(clock.instant()).andReturn(Instant.ofEpochMilli(0));
+
+    // first message
+    expect(requestsMappingIterator.hasNext()).andReturn(true);
+    expect(clock.instant()).andReturn(Instant.ofEpochMilli(0));
+    expect(requestsMappingIterator.nextValue()).andReturn(request);
+    expect(clock.instant()).andReturn(Instant.ofEpochMilli(1));
+    expect(mockedChunkedOutputFactory.getChunkedOutput()).andReturn(mockedChunkedOutput);
     expect(mockedChunkedOutput.isClosed()).andReturn(false);
+    mockedChunkedOutput.write(sucessResult);
```
ehumber, line 425: Two c's in success.

AndrewJSchofield (approved):
I have reviewed the code with @ehumber and am happy to approve. I do not think this is a sensible way to handle streamed REST produce requests which overrun the KafkaProducer request limits, but this does at least prevent heap exhaustion when it happens. I think there's a significant refactor due in this code at some point.

AndrewJSchofield (approved):

ehumber:
Sorry I didn't manage to get this tidied up in time :(. 

I made a PR from a branch in confluentinc/kafka-rest, so if my original branch from my fork goes missing when I leave, hopefully you still have the code

https://github.com/confluentinc/kafka-rest/pull/1098

For this PR I recommend you don't merge it until you need to.  That will most probably be when you look at rate limiting around REST produce, and potentially remove it in REST and rely on backpressure from Kafka to keep the requests at a sensible limit.

dimitarndimitrov (commented):
Whoa, I've been amazingly delinquent on this one... Sorry @ehumber :)

PR #1044, opened 2022-07-20
APIF-2768: Workaround for "Failed to bind to 0.0.0.0/0.0.0.0:9998" error.

rigelbm:
This change will cause `JerseyTest` to run on a random port instead of the default 9998. This should work around the fact the ports are possibly leaking between tests.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/v2/PartitionsResourceTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/v2/PartitionsResourceTest.java
@@ -39,16 +39,11 @@ import org.easymock.EasyMockExtension;
 import org.easymock.Mock;
 import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.test.JerseyTest;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.TestInstance;
-import org.junit.jupiter.api.TestInstance.Lifecycle;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 @ExtendWith(EasyMockExtension.class)
-@TestInstance(Lifecycle.PER_CLASS)
 public class PartitionsResourceTest extends JerseyTest {
 
   private static final String CLUSTER_ID = "cluster-1";
@@ -93,16 +88,6 @@ public class PartitionsResourceTest extends JerseyTest {
     application.register(new PartitionsResource(() -> partitionManager));
     application.register(new JacksonMessageBodyProvider());
     return application;
-  }
-
-  @BeforeAll
-  public void before() throws Exception {
-    super.setUp();
-  }
-
-  @AfterAll
-  public void after() throws Exception {
-    super.tearDown();
   }
 
   @BeforeEach
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,7 @@
     </modules>
 
     <properties>
+        <jetty.version>9.4.48.v20220622</jetty.version>
         <confluent.maven.repo>https://packages.confluent.io/maven/</confluent.maven.repo>
         <checkstyle.suppressions.location>checkstyle/suppressions.xml</checkstyle.suppressions.location>
         <io.confluent.kafka-rest.version>7.2.1-0</io.confluent.kafka-rest.version>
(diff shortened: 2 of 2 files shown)


ehumber:
Hi @rigelbm 

I think the failed to bind error is caused by a jersey upgrade, not a jetty one - kafka-rest fails after https://github.com/confluentinc/rest-utils/pull/329 went in, and it's only the jersey tests that are failing.

Separately, if you're looking at moving jetty to 4.48, please bear in mind 
https://github.com/confluentinc/rest-utils/pull/339
and 
https://github.com/confluentinc/rest-utils/pull/330

If you're looking at the `Failed to bind to 0.0.0.0/0.0.0.0:9998` then I won't :)

rigelbm:
Oh, shoot. I was so concerned with not breaking kafka-rest in the Jetty upgrade, that I didn't realize I had broken it already. Don't worry, I will fix it. Thanks for letting me know. I will take this as high priority.

rigelbm:
Hi @ehumber, 

To expand, what I think happened here is that:

1. I did the original Jersey upgrade in 5.4.x.
2. Jenkins was green (this is only a problem in 7.2.x+).
3. I pint-merged it up.
4. 7.2.x+ started failing.

The problem being that pint-merge doesn't know how to look into downstream dependencies before merging.

The takeaway here is for me to check all downstream dependencies after a pint-merge. In the meantime, if you notice another problem like this (kafka-rest broken because for rest-utils), please notify #apif-oncall immediately.

Thanks!

rigelbm:
Also:

> and it's only the jersey tests that are failing.

In my local machine I see many more, include ClusterTestHarness ones. I assume that's because my machine can take on more parallelism (thanks AMD).

ehumber:
> Also:
> 
> > and it's only the jersey tests that are failing.
> 
> In my local machine I see many more, include ClusterTestHarness ones. I assume that's because my machine can take on more parallelism (thanks AMD).

That's interesting, laptop and the jenkins boxes only consistently fail the extends JerseyTest ones. And the failure does look to have started when the jersey upgrade was done.

I was trying to run kafka-rest with a back level jersey to prove it before letting apif know, but failing!

Thanks for lookin. Let me know if you need anything else

rigelbm:
@ehumber,

The solution above "works". I still don't understand why it started failing after the upgrade. I've been combing Jersey change from 2.34 to 2.36, but haven't found a culprit yet. It doesn't seem like the tests magically started running in parallel, since this is something that is configured in maven. My suspicion is that this has something to do with JUnit5 (https://github.com/eclipse-ee4j/jersey/commit/3d131db15333dc561ea6e1ffe7952c076e507fe7), and that somehow the port is leaking.

If that is the case, the solution above merely hides the problem. I will dig more into it, but in the meantime what do you prefer:

1. Merge the above changes in kafka-rest, which will make the tests run and pass, but ports might be leaking (which might or might not be a problem)
2. Ignore the failing tests in kafka-rest until I get to the bottom of it.

Please, let me know what workaround do you prefer until I have a final solution for the issue.

ehumber:
> @ehumber,
> 
> The solution above "works". I still don't understand why it started failing after the upgrade. I've been combing Jersey change from 2.34 to 2.36, but haven't found a culprit yet. It doesn't seem like the tests magically started running in parallel, since this is something that is configured in maven. My suspicion is that this has something to do with JUnit5 ([eclipse-ee4j/jersey@3d131db](https://github.com/eclipse-ee4j/jersey/commit/3d131db15333dc561ea6e1ffe7952c076e507fe7)), and that somehow the port is leaking.
> 
> If that is the case, the solution above merely hides the problem. I will dig more into it, but in the meantime what do you prefer:
> 
> 1. Merge the above changes in kafka-rest, which will make the tests run and pass, but ports might be leaking (which might or might not be a problem)
> 2. Ignore the failing tests in kafka-rest until I get to the bottom of it.
> 
> Please, let me know what workaround do you prefer until I have a final solution for the issue.

I'd like to get kafka-rest working asap, even if it's a workaround.

The reason is that ce-kafka-rest currently has a mismatch of jetty levels - 4.48 from rest-utils and 4.44 from the kafka clients, and that causes classpath issues that prevent any of the tests from passing.  I get a lot of test failures in my inbox :)

rest-utils has been rolled back to 4.44, but the change hasn't made it up the stack past kafka-rest because of the failing jersey tests in jenkins.

So, yes to moving forward, but I notice that this change also updates jetty to 4.48, and that's not going to enable ce-kafka-rest to pass.  Until ce-kafka clients also updates their jetty level to 4.48 (and I'm not sure what's stopping that, but @abhijeet2096-confluent  has more context).

rigelbm:
When I said "Merge the above changes", I meant "Cleanup the above changes and merge them" :). I will do that ASAP. 1min.

rigelbm:
@ehumber, please review.

ehumber (approved):

PR #1051, opened 2022-07-27
NPE when post to topics endpoint without payload

jliunyu:
When create topics without payload, a general server errors will return currently: 
```
curl -X POST http://localhost:9391/v3/clusters/kZ7pDeiQSS-9eRtOXlWSMQ/topics -H "Authorization: Basic ABC"

returns a {"error_code":500,"message":"Internal Server Error"}%
```

But this should be a user error, after the user adds payload, it should be resolved. The return message and error code should be clear to mention that.

Result for local manual test.
```
curl -X POST http://localhost:8082/v3/clusters/vD3GG4PqT0-C1UcpR1j-Ag/topics -H "Authorization: Basic ABC" -H "Content-Type:application/json"
{"error_code":42206,"message":"Payload error. Null input provided. Data is required."}% 
``` 

Result of unit test:
```
mvn -Dtest=TopicsResourceTest test
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.708 s - in io.confluent.kafkarest.resources.v2.TopicsResourceTest
[INFO] Running io.confluent.kafkarest.resources.v3.TopicsResourceTest
[INFO] Tests run: 16, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.059 s - in io.confluent.kafkarest.resources.v3.TopicsResourceTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  16.044 s
[INFO] Finished at: 2022-07-26T17:11:33-07:00
[INFO] ------------------------------------------------------------------------
jliu@C02DM32YMD6T kafka-rest % 
```

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
@@ -143,7 +143,7 @@ public final class TopicsResource {
       @Valid CreateTopicRequest request) {
 
     if (request == null) {
-      throw Errors.invalidPayloadException("Null input provided. Data is required.");
+      throw Errors.invalidPayloadException("Request body is empty. Data is required.");
     }
 
     String topicName = request.getTopicName();
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
@@ -19,6 +19,7 @@ import static java.util.Collections.emptyList;
 import static java.util.Collections.emptySet;
 import static java.util.Objects.requireNonNull;
 
+import io.confluent.kafkarest.Errors;
 import io.confluent.kafkarest.controllers.TopicManager;
 import io.confluent.kafkarest.entities.Topic;
 import io.confluent.kafkarest.entities.v3.CreateTopicRequest;
@@ -140,6 +141,11 @@ public final class TopicsResource {
       @Suspended AsyncResponse asyncResponse,
       @PathParam("clusterId") String clusterId,
       @Valid CreateTopicRequest request) {
+
+    if (request == null) {
+      throw Errors.invalidPayloadException("Null input provided. Data is required.");
+    }
+
     String topicName = request.getTopicName();
     Optional<Integer> partitionsCount = request.getPartitionsCount();
     Optional<Short> replicationFactor = request.getReplicationFactor();
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceAction.java
@@ -116,7 +116,7 @@ public final class ProduceAction {
       throws Exception {
 
     if (requests == null) {
-      throw Errors.invalidPayloadException("Null input provided. Data is required.");
+      throw Errors.invalidPayloadException("Request body is empty. Data is required.");
     }
 
     ProduceController controller = produceControllerProvider.get();
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/TopicsResourceIntegrationTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/TopicsResourceIntegrationTest.java
@@ -470,6 +470,16 @@ public class TopicsResourceIntegrationTest extends ClusterTestHarness {
   }
 
   @Test
+  public void createTopic_withoutRequestBody_returnsInvalidPayload() {
+    String clusterId = getClusterId();
+    Response response =
+        request("/v3/clusters/" + clusterId + "/topics")
+            .accept(MediaType.APPLICATION_JSON)
+            .post(Entity.entity(null, MediaType.APPLICATION_JSON));
+    assertEquals(422, response.getStatus());
+  }
+
+  @Test
   public void deleteTopic_existingTopic_deletesTopic() {
     String clusterId = getClusterId();
 
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProduceActionTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProduceActionTest.java
@@ -613,7 +613,7 @@ public class ProduceActionTest {
         assertThrows(
             RestConstraintViolationException.class,
             () -> produceAction.produce(fakeAsyncResponse, "clusterId", "topicName", null));
-    assertEquals("Payload error. Null input provided. Data is required.", e.getMessage());
+    assertEquals("Payload error. Request body is empty. Data is required.", e.getMessage());
     assertEquals(42206, e.getErrorCode());
   }
 
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/TopicsResourceTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/TopicsResourceTest.java
@@ -24,6 +24,7 @@ import static org.easymock.EasyMock.expect;
 import static org.easymock.EasyMock.replay;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.google.common.collect.ImmutableMap;
@@ -43,6 +44,7 @@ import io.confluent.kafkarest.entities.v3.TopicDataList;
 import io.confluent.kafkarest.response.CrnFactoryImpl;
 import io.confluent.kafkarest.response.FakeAsyncResponse;
 import io.confluent.kafkarest.response.FakeUrlFactory;
+import io.confluent.rest.exceptions.RestConstraintViolationException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -638,6 +640,18 @@ public class TopicsResourceTest {
             .build());
 
     assertEquals(NotFoundException.class, response.getException().getClass());
+  }
+
+  @Test
+  public void createTopic_nonData_throwsInvalidPayload() {
+    FakeAsyncResponse response = new FakeAsyncResponse();
+
+    RestConstraintViolationException e =
+        assertThrows(
+            RestConstraintViolationException.class,
+            () -> topicsResource.createTopic(response, TOPIC_1.getClusterId(), null));
+    assertEquals("Payload error. Null input provided. Data is required.", e.getMessage());
+    assertEquals(42206, e.getErrorCode());
   }
 
   @Test
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/TopicsResourceTest.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/TopicsResourceTest.java
@@ -650,7 +650,7 @@ public class TopicsResourceTest {
         assertThrows(
             RestConstraintViolationException.class,
             () -> topicsResource.createTopic(response, TOPIC_1.getClusterId(), null));
-    assertEquals("Payload error. Null input provided. Data is required.", e.getMessage());
+    assertEquals("Payload error. Request body is empty. Data is required.", e.getMessage());
     assertEquals(42206, e.getErrorCode());
   }
 


jonchiu:
Looking at the api spec, it appears that the topic name is the only required field. What happens if you submit a request body with only the partition count, what is the expected error for this? Do we have a test for this?

jliunyu:
> Looking at the api spec, it appears that the topic name is the only required field. What happens if you submit a request body with only the partition count, what is the expected error for this? Do we have a test for this?

Hi @jonchiu , thanks for reviewing. Yes, I have tested this before. The result mentions that the topic name is NULL

```
jliu@C02DM32YMD6T kafka-rest % curl -X POST http://localhost:8082/v3/clusters/vD3GG4PqT0-C1UcpR1j-Ag/topics -H "Authorization: Basic ABC" -H "Content-Type:application/json" --data {}
Cannot construct instance of `io.confluent.kafkarest.entities.v3.CreateTopicRequest`, problem: Null topicName
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 2]%        jliu@C02DM32YMD6T kafka-rest % 
```

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
```diff
@@ -140,6 +141,11 @@ public void createTopic(
       @Suspended AsyncResponse asyncResponse,
       @PathParam("clusterId") String clusterId,
       @Valid CreateTopicRequest request) {
+
+    if (request == null) {
+      throw Errors.invalidPayloadException("Null input provided. Data is required.");
```
jonchiu, line 146: I think something along the lines of "request body is empty" instead of null input.

jonchiu:
> > Looking at the api spec, it appears that the topic name is the only required field. What happens if you submit a request body with only the partition count, what is the expected error for this? Do we have a test for this?
> 
> Hi @jonchiu , thanks for reviewing. Yes, I have tested this before. The result mentions that the topic name is NULL
> 
> ```
> jliu@C02DM32YMD6T kafka-rest % curl -X POST http://localhost:8082/v3/clusters/vD3GG4PqT0-C1UcpR1j-Ag/topics -H "Authorization: Basic ABC" -H "Content-Type:application/json" --data {}
> Cannot construct instance of `io.confluent.kafkarest.entities.v3.CreateTopicRequest`, problem: Null topicName
>  at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 2]%        jliu@C02DM32YMD6T kafka-rest % 
> ```

Awesome. Ideally we return a prettier error response, but that is out of scope for this PR.

jonchiu:
> > Looking at the api spec, it appears that the topic name is the only required field. What happens if you submit a request body with only the partition count, what is the expected error for this? Do we have a test for this?
> 
> Hi @jonchiu , thanks for reviewing. Yes, I have tested this before. The result mentions that the topic name is NULL
> 
> ```
> jliu@C02DM32YMD6T kafka-rest % curl -X POST http://localhost:8082/v3/clusters/vD3GG4PqT0-C1UcpR1j-Ag/topics -H "Authorization: Basic ABC" -H "Content-Type:application/json" --data {}
> Cannot construct instance of `io.confluent.kafkarest.entities.v3.CreateTopicRequest`, problem: Null topicName
>  at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 2]%        jliu@C02DM32YMD6T kafka-rest % 
> ```

Awesome. Ideally we return a prettier error response, but that is out of scope for this PR.

ehumber:
> org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream

https://confluentinc.atlassian.net/browse/KREST-6129 covers this

checkstyle/checkstyle.xml
```diff
@@ -6,7 +6,7 @@
 <module name="Checker">
   <module name="TreeWalker">
     <module name="ImportControl">
-      <property name="file" value="checkstyle/import_control.xml"/>
```
ehumber, line 9: Did you mean to leave this changed, or was it part of trying to get the checkstyle failure to work?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
```diff
@@ -140,6 +141,10 @@ public void createTopic(
       @Suspended AsyncResponse asyncResponse,
       @PathParam("clusterId") String clusterId,
       @Valid CreateTopicRequest request) {
+
+    if (request == null) {
+      throw Errors.invalidPayloadException("Request body is empty. Data is required.");
```
ehumber, line 146: We should make the error message match what is in ProduceAction.  There it says:

```
    if (requests == null) {
      throw Errors.invalidPayloadException("Null input provided. Data is required.");
    }
```

And now I'm debating whether these two errors should return a 404 rather than a 422 (The 422 was chosen for ProduceAction as that matches the other cases of malformed data in the class).  

But as long as the two methods return the same error code I don't have too strong feelings.

Have you taken a look at any of our other POST endpoints (eg createACL) to see what they do with no data provided?

ehumber (approved):

checkstyle/checkstyle.xml
dimitarndimitrov: @jliunyu, that's really weird - what kind of errors did you get before and after this change and with what Maven command? Can you please also check whether non-test targets (e.g. `clean package` or `validate`) are behaving as expected before and after the changes?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
dimitarndimitrov: Agreed, 422 seems like a slightly worse fit, and also is less well understood 4xx response status code compared to the ubiquitous 404.

dimitarndimitrov (commented):
I think it's a good idea to also add a test in `TopicsResourceIntegrationTest` - there you can assert that the returned response is an HTTP 404/HTTP 422 response.

checkstyle/checkstyle.xml
jliunyu: This one is just for removing the checkstyle failure, I'm will remove this change.

jliunyu:
> > org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream
> 
> https://confluentinc.atlassian.net/browse/KREST-6129 covers making this type of error response nicer - we shouldn't display the jersey package/class/method in our response.

Hi @ehumber , thanks for your comment, looks this is another security thing, can we fix it together with the follow up security ticket?

kafka-rest/src/main/java/io/confluent/kafkarest/Errors.java
```diff
@@ -63,6 +63,14 @@ public static RestException leaderNotAvailableException() {
     return new RestNotFoundException(LEADER_NOT_AVAILABLE_MESSAGE, LEADER_NOT_AVAILABLE_ERROR_CODE);
   }
 
+  public static final String INVALID_PAYLOAD_MESSAGE = "Payload error. ";
+  public static final int INVALID_PAYLOAD_ERROR_CODE = 40405;
+
+  public static RestConstraintViolationException invalidPayloadException(String cause) {
```
jliunyu, line 69: I updated the error code to 404, but all 404 error finally throws `RestException`, I'm not sure we should throw a RestConstraintViolationException or RestException. @ehumber @dimitarndimitrov any suggestions for this?

ehumber:
> > > org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream
> > 
> > 
> > https://confluentinc.atlassian.net/browse/KREST-6129 covers making this type of error response nicer - we shouldn't display the jersey package/class/method in our response.
> 
> Hi @ehumber , thanks for your comment, looks this is another security thing, can we fix it together with the follow up security ticket?

Yes - can you add the jira ticket number in here please for reference

kafka-rest/src/main/java/io/confluent/kafkarest/Errors.java
ehumber: I don't think the exception type matters too much, as long as what the end user sees in the message text is useful to them.  Do they still see the "Request body is empty. Data is required." message even with the RestException?

jliunyu:
> 



> Yes - can you add the jira ticket number in here please for reference

This is the JIRA ticket: https://confluentinc.atlassian.net/browse/KREST-7107

kafka-rest/src/main/java/io/confluent/kafkarest/Errors.java
jliunyu: Yes, they will see the Request body is empty. Data is required. message

jliunyu: If we change the error code to 404, it will cause `ProduceActionTest.produceNoSchemaRegistryDefined` failed

```
14:07:29  [ERROR] Failures: 
14:07:29  [ERROR]   ProduceActionTest.produceNoSchemaRegistryDefined:171 
14:07:29    Expectation failure on verify:
14:07:29      ChunkedOutput.write(ErrorHolder{error=ErrorResponse{errorCode=404, message=Error: 40405 : Payload error. Schema Registry must be configured when using schemas.}}): expected: 1, actual: 0
14:07:29    Unexpected method calls:
14:07:29      ChunkedOutput.write(ErrorHolder{error=ErrorResponse{errorCode=422, message=Error: 40405 : Payload error. Schema Registry must be configured when using schemas.}})
14:07:29  [INFO] 
14:07:29  [ERROR] Tests run: 521, Failures: 1, Errors: 0, Skipped: 5
```
RestConstraintViolationException can only return 422 : https://github.com/confluentinc/rest-utils/blob/master/core/src/main/java/io/confluent/rest/exceptions/ConstraintViolationExceptionMapper.java#L31

But for 404, looks it's `Not found`, is it expected if no request body?

```
        PAYMENT_REQUIRED(402, "Payment Required"),
        FORBIDDEN(403, "Forbidden"),
        NOT_FOUND(404, "Not Found"),
        METHOD_NOT_ALLOWED(405, "Method Not Allowed"),
        NOT_ACCEPTABLE(406, "Not Acceptable"),
        PROXY_AUTHENTICATION_REQUIRED(407, "Proxy Authentication Required"),
        REQUEST_TIMEOUT(408, "Request Timeout"),
        CONFLICT(409, "Conflict"),
```

ehumber: So this test case has a -d payload passed in. That payload has a schema defined, but the proxy is configured to not point at a schema registry.

So it's not that there is no request body, it's that there is a request body that doesn't work for the particular configuration of the REST server.

404 not found doesn't sound quite right for that, I think 422 is better.

Can you make all the exception mappers line up to do that?  If not, shall we just get them to return the 422 like you coded this change to do to start with?

jliunyu: Thanks @ehumber , I changed it to 422.

Currently I met jenkins error, I don't think they related to my change because I also saw them on my local with a clean repo. I will also take a look to see how to fix it.
```
21:57:08  [ERROR] Failures: 
21:57:08  [ERROR]   AvroProduceConsumeTest>SchemaProduceConsumeTest.produceThenConsume_returnsExactlyProduced:181->SchemaProduceConsumeTest.assertMapEquals:220 Expected and actual are not equal. Extra: [], Missing: [1, 2, 3], Different: []
21:57:08  [ERROR]   JsonSchemaProduceConsumeTest>SchemaProduceConsumeTest.produceThenConsume_returnsExactlyProduced:181->SchemaProduceConsumeTest.assertMapEquals:220 Expected and actual are not equal. Extra: [], Missing: [1, 2, 3], Different: []
21:57:08  [ERROR]   ProtobufProduceConsumeTest>SchemaProduceConsumeTest.produceThenConsume_returnsExactlyProduced:181->SchemaProduceConsumeTest.assertMapEquals:220 Expected and actual are not equal. Extra: [], Missing: [{"key":2}, {"key":3}, {"key":1}], Different: []
21:57:08  [INFO] 
21:57:08  [ERROR] Tests run: 180, Failures: 3, Errors: 0, Skipped: 1
```

ehumber: Yup, they are some of the flakey tests.  They should pass intermittently though, and on my laptop they pass very much more than they fail.  Is it a solid failure now?

jliunyu: Yeah, solid failure now, even fail without any change.

ehumber: Did this get fixed?

jliunyu: Yeah, this is fixed.

jliunyu: I just run again on my local, let me run it on jenkins again.

ehumber (approved):

PR #1058, opened 2022-08-25
Revert replacement of poll(Duration) with poll(long)

dimitarndimitrov:
#1035 replaced a usage of the deprecated [KafkaConsumer#poll(long)](https://github.com/apache/kafka/blob/0507597597e1d2d9adaa33b87e5ea509e12fd2f0/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1167) API with the [KafkaConsumer#poll(Duration)](https://github.com/apache/kafka/blob/0507597597e1d2d9adaa33b87e5ea509e12fd2f0/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1214) API in the internals of Kafka REST's Consume v2 API. There is a slight difference in behavior between the two KafkaConsumer APIs related to blocking during consumer rebalancing.
- See [this wiki](https://cwiki.apache.org/confluence/display/KAFKA/KIP-266%3A+Fix+consumer+indefinite+blocking+behavior#KIP266:Fixconsumerindefiniteblockingbehavior-Consumer#poll) for more details.

Unfortunately although this seemed fairly reasonable, we are seeing change in behavior which breaks some tests and can affect users of the Kafka REST Consume v2 API.
Because of that we will be reverting back to the old poll(long) API until we find a better alternative.

CC @dennisgermany

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerState.java
@@ -381,7 +381,7 @@ public abstract class KafkaConsumerState<KafkaKeyT, KafkaValueT, ClientKeyT, Cli
    * be invoked with the lock held, i.e. after startRead().
    */
   private synchronized void getOrCreateConsumerRecords() {
-    ConsumerRecords<KafkaKeyT, KafkaValueT> polledRecords = consumer.poll(Duration.ZERO);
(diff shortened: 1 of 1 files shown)


dimitarndimitrov:
The CI validation has failed with unrelated downstream validation checks, but it does show the relevant failing tests succeeding now. I am rerunning it but might end up submitting anyway if I get similar (not worse) results on the new run.

dimitarndimitrov:
The second CI run failed in a similar way so like I've explained above I'll proceed with merging anyway.

PR #1059, opened 2022-09-01
KREST-7275 update openapi changes for batch acl

ehumber:
This commit
 - updates the kafka-rest openapi.yaml to include the changes for batch acl create which are required by the UI

 - It ports the code from Tom's ce-kafka-rest PR into kafka-rest, making this a new publically available API.  This feels like the right thing to do, because it's a simple API (just a wrapper around the individual create API comment), it was odd to split to very linked APIs across two repos (and this made our lives much harder wrt eg the integration test and the openapi doc).

 - Add a null check in the single create ACL to avoid a NPE and the subsequent 500 error being returned to the user.  Unfortunately this typed the cyclomatic complexity of all the valid payload checks in the createAcl class over the limit, so I had to pull them out into a separate method.

 - Add a log.error of any stack traces that are part of the failed async processing future.  Without these it's really hard to debug what goes wrong if any of our async calls fail.  Note that this change does not alter what the end user sees, just what comes out in our logs.

For batch ACL, if one of the structures within the data array is malformend  (eg an enum value doesn't match) then none of the  ACLs are created.  However, if one of the structures in the data array contains a value that the broker needs to determine is invalid, then the other structures are still applied.  For example if the user id specified in the first item in the array of ACLs does not actually exist, then that ACL is not created, but the second ACL in the array of ACLs is created (assuming it is valid)

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -102,6 +102,33 @@ paths:
         '5XX':
           $ref: '#/components/responses/ServerErrorResponse'
 
+  /clusters/{cluster_id}/acls:batch:
+    parameters:
+      - $ref: '#/components/parameters/ClusterId'
+
+    post:
+      summary: 'Batch Create ACLs'
+      operationId: batchCreateKafkaV3Acls
+      description: |-
+        Internal use only.
+
+        Creates ACLs.
+      tags:
+        - ACL (v3)
+      requestBody:
+        $ref: '#/components/requestBodies/BatchCreateAclRequest'
+      responses:
+        '201':
+          description: 'No Content'
+        '400':
+          $ref: '#/components/responses/BadRequestErrorResponse'
+        '401':
+          $ref: '#/components/responses/UnauthorizedErrorResponse'
+        '429':
+          $ref: '#/components/responses/TooManyRequestsErrorResponse'
+        '5XX':
+          $ref: '#/components/responses/ServerErrorResponse'
+
   /clusters/{cluster_id}/acls:
     parameters:
       - $ref: '#/components/parameters/ClusterId'
@@ -136,7 +163,7 @@ paths:
           $ref: '#/components/responses/ServerErrorResponse'
 
     post:
-      summary: 'Create ACLs'
+      summary: 'Create an ACL'
       operationId: createKafkaV3Acls
       description: |-
         [![Generally Available](https://img.shields.io/badge/Lifecycle%20Stage-Generally%20Available-%2345c6e8)](#section/Versioning/API-Lifecycle-Policy)
@@ -1723,6 +1750,17 @@ components:
         permission:
           $ref: '#/components/schemas/AclPermission'
 
+    CreateAclRequestDataList:
+      allOf:
+        - type: object
+          required:
+            - data
+          properties:
+            data:
+              type: array
+              items:
+                $ref: '#/components/schemas/CreateAclRequestData'
+
     CreateTopicRequestData:
       type: object
       required:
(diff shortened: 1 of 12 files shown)


api/v3/openapi.yaml
```diff
@@ -102,6 +102,33 @@ paths:
...
 
+  /clusters/{cluster_id}/acls:batch:
+    parameters:
+      - $ref: '#/components/parameters/ClusterId'
+
+    post:
+      summary: 'Batch Create ACLs'
+      operationId: batchCreateKafkaV3Acls
+      description: |-
+        Internal use only.
```
jaykimc3, line 113: remove this?

api/v3/openapi.yaml
```diff
@@ -136,7 +163,7 @@ paths:
           $ref: '#/components/responses/ServerErrorResponse'
 
     post:
-      summary: 'Create ACLs'
+      summary: 'Create an ACL'
       operationId: createKafkaV3Acls
```
jaykimc3, line 167: nit: remove `s` at the end?

jaykimc3 (approved):
thanks for tagging me on this. have some minor comments.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AclsResource.java
```diff
@@ -158,25 +159,13 @@ public void createAcl(
...
-        || request.getOperation() == Acl.Operation.UNKNOWN) {
-      throw new BadRequestException("operation cannot be ANY");
-    }
-    if (request.getPermission() == Acl.Permission.ANY
-        || request.getPermission() == Acl.Permission.UNKNOWN) {
-      throw new BadRequestException("permission cannot be ANY");
+
+    if (request == null) {
+      throw Errors.invalidPayloadException("Null input provided. Data is required.");
     }
```
dimitarndimitrov, line 165: nit: I wonder if the `@Valid` annotation doesn't preclude this check from happening. Below I've suggested adding a test that checks that case.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/AsyncResponses.java
```diff
@@ -104,8 +108,10 @@ public void asyncResume(AsyncResponse asyncResponse) {
                 asyncResponse.resume(responseBuilder.entity(entity).build());
               }
             } else if (exception instanceof CompletionException) {
+              log.error("AsyncResponse CompletionException: {}", exception);
```
dimitarndimitrov, line 111: nit: I am not sure if we would be able to link the response error to the request to which it corresponds. Could we log something additional from the async response to achieve that?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/CreateAclBatchAction.java
```diff
@@ -0,0 +1,91 @@
...
+  @Inject
+  public CreateAclBatchAction(Provider<AclManager> aclManager) {
+    this.aclManager = requireNonNull(aclManager);
+  }
+
+  @POST
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Produces(MediaType.APPLICATION_JSON)
+  @PerformanceMetric("v3.acls.create")
+  @ResourceName("api.v3.acls.create")
```
dimitarndimitrov, line 57: I think these should clearly indicate this as a separate API - WDYT about `v3.acls.batch-create`?

api/v3/openapi.yaml
ehumber: I don't know if this will have an unexpected effect downstream.  @linouk23 might know?  If not, I'm leaving the `s` on the end to avoid breaking eg the docs or the go sdk generation.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/AsyncResponses.java
ehumber: Yes - but the AsyncResponse doesn't have anything useful in it, and the entity is a generic type that extends Object, so I cant guarantee anything more than the class, hash and hope that there is a useful toString defined on the object.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/AclsResource.java
ehumber: I put this in because I manually tested that I got 500s without it, so it is needed.

dimitarndimitrov (approved):
+1, nice!

PR #1070, opened 2022-10-19
KENGINE-238 update createBrokerConfig method signatures

ehumber:
An extra value was added to this method signature in ce-kafka.  Updating kafka-rest's usage to match


ehumber:
Not tested locally as I couldn't get it to pull the latest ce-kafka :(

Need to wait and see if the CI passes before merging.

ehumber:
Ohh, this isn't going to work.  ce-kafka and apache kafka have different method signatures for the same method.

Kafka-rest pre-reqs apache kafka
ce-kafka-rest pre-reqs ce-kafka

But ce-kafka-rest uses kafka-rests version of the cluster test harness, and so barfs because kafka-rest needs a different method signature.

@delainerogers  Your original ce-kafka fix needs to make it into apache kafka I believe!  Not sure how you make that happen though.

PR #1071, opened 2022-10-24
KREST-7107 Shouldn't provide overly detailed info or stack traces in responses -…

jliunyu:
… Follow up

This fix is related to v3 version.

```curl -X POST \
     -H "Content-Type: application/json" \
     -H "Accept: application/json" \
     --data '{"records":[{"key":"abc","value":"alarm clock"},{"key":"htanaka","value":"batteries"},{"key":"awalther","value":"bookshelves"}]}' \
     "http://localhost:8082/v3/clusters/vD3GG4PqT0-C1UcpR1j-Ag/topics/tst/records"
{"error_code":400,"message":"Unrecognized field \"records\" (class io.confluent.kafkarest.entities.v3.AutoValue_ProduceRequest$Builder), not marked as ignorable (6 known properties: \"value\", \"originalSize\", \"partitionId\", \"headers\", \"key\", \"timestamp\"])\n at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 129] (through reference chain: io.confluent.kafkarest.entities.v3.AutoValue_ProduceRequest$Builder[\"records\"])"}
```
The source part shouldn't include the source part

This fix is security related, so need to start from 5.4.x, then pint merge to 7.4.x and master

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
@@ -441,7 +441,17 @@ public abstract class StreamingResponse<T> {
     @SuppressWarnings("unchecked")
     private ErrorResponse toErrorResponse(Throwable error) {
       Response response = mapper.toResponse((T) error);
-      return ErrorResponse.create(errorCode.apply(response), message.apply(response));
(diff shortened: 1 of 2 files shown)


trnguyencflt (commented):
LGTM

PR #1077, opened 2022-11-02
Remove obsolete Avro test

rayokota:
The AvroUtils.toJson has been enhanced to handle primitives, so the test here can be removed


dimitarndimitrov (approved):
+1.

The PR CI problems reporting downstream failures seem to be a problem unrelated to the changes here and seemingly not something that will break after this gets merged.

PR #1082, opened 2022-11-16
KREST-8373 KafkaRestConfig, add fields to capture broker's proxy-protocol version.

msn-tldr:
These field will be set in KafkaRestApplicationProvider.java & propagated up the REST stack along with KafkaRestConfig object.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -145,6 +145,12 @@ public class KafkaRestConfig extends RestConfig {
   public static final ConfigDef.Range PROXY_FETCH_MIN_BYTES_VALIDATOR =
       ConfigDef.Range.between(-1, PROXY_FETCH_MIN_BYTES_MAX);
 
+  public static final String BROKER_PROXY_PROTOCOL_VERSION_CONFIG = "broker.proxy.protocol.version";
+  public static final String BROKER_PROXY_PROTOCOL_VERSION_DOC =
+      "Which proxy protocol the given cluster's brokers support."
+          + "The options are 'NONE', 'V1', and 'V2'";
+  public static final String BROKER_PROXY_PROTOCOL_VERSION_DEFAULT = "NONE";
+
   @Deprecated public static final String PRODUCER_THREADS_CONFIG = "producer.threads";
 
   @Deprecated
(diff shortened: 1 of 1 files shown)


msn-tldr:
I don't this requires a UT, as public interface of the class isn't changed.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -145,6 +145,13 @@ public class KafkaRestConfig extends RestConfig {
   public static final ConfigDef.Range PROXY_FETCH_MIN_BYTES_VALIDATOR =
       ConfigDef.Range.between(-1, PROXY_FETCH_MIN_BYTES_MAX);
 
+  public static final String BROKER_PROXY_PROTOCOL_VERSION_CONFIG = "broker.proxy.protocol.version";
+  public static final String BROKER_PROXY_PROTOCOL_VERSION_DOC =
+      "Which proxy protocol the given cluster's brokers support."
+          + "The default value of \"\" overriden at runtime with more appropriate value of "
+          + "ProxyProtocol.NONE";
```
ehumber, line 152: So this is an enum in the backend code?  Do we own the enum out of interest, or is it from a ProxyProtocol library?

ProxyProtocol.NONE, ProxyProtocol.V1 and ProxyProtocol.V2 at a guess?

The other example of an enum type config we have is 

https://github.com/confluentinc/kafka-rest/blob/9aef0cf2c24f8a414aada8685100900645ef112c/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java#L431

So we should follow that pattern - state in the _DOC field what the options are, and fill in the default here with `none` so it's obvious what the value is without having to read the comment.

What do you think?

msn-tldr: > ProxyProtocol.NONE, ProxyProtocol.V1 and ProxyProtocol.V2 at a guess?
That's right, here is the [enum](https://github.com/confluentinc/ce-kafka/blob/6c2d349fda5d6ea847d100d506cd962f2ebe3363/clients/src/main/java/org/apache/kafka/common/network/ProxyProtocol.java#L9) .

> So we should follow that pattern - state in the _DOC field what the options are, and fill in the default here with none so it's obvious what the value is without having to read the comment.
I had though along the same lines, but then decided to not do it. As this would leak 'ce-kafka' details(enum & its values) into 'kafka-rest'(CCS/opensource world).
But i do see ur point, it makes sense for the config-property to have sensible defaults inline with the enum. So I would go ahead and make those changes.

Ideally there would have been 'Ce' version of KafkaRestConfig( that could extend KafkaRestConfig). CeKafkaRestConfig would have cleanly handled using enum from 'ce-kafka' IMO. What do you think?

PS - If you agree on CeKafkaRestConfig, then we should track that work separately.

msn-tldr: @ehumber updated the PR with sensible default based on the enum.
I also filed a jira to track creation CeKafkaRestConfig.java. I added a label of 'backlog', let me know if u have a better way to track these.
https://confluentinc.atlassian.net/browse/KREST-8441

PR #1084, opened 2022-11-22
KREST-3009 Add ability to update partition count on a topic

ehumber:
Call looks like

```
curl -X PATCH http://localhost:8082/v3/clusters/3W-2oFuqSYuErBzrRrTKyw/topics/bob  -H "Content-Type: application/json" -d '{"partition_count":"10"}'  | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   688    0   664  100    24   6036    218 --:--:-- --:--:-- --:--:--  6254
{
  "kind": "KafkaTopic",
  "metadata": {
    "self": "http://localhost:8082/v3/clusters/3W-2oFuqSYuErBzrRrTKyw/topics/bob",
    "resource_name": "crn:///kafka=3W-2oFuqSYuErBzrRrTKyw/topic=bob"
  },
  "cluster_id": "3W-2oFuqSYuErBzrRrTKyw",
  "topic_name": "bob",
  "is_internal": false,
  "replication_factor": 1,
  "partitions_count": 10,
  "partitions": {
    "related": "http://localhost:8082/v3/clusters/3W-2oFuqSYuErBzrRrTKyw/topics/bob/partitions"
  },
  "configs": {
    "related": "http://localhost:8082/v3/clusters/3W-2oFuqSYuErBzrRrTKyw/topics/bob/configs"
  },
  "partition_reassignments": {
    "related": "http://localhost:8082/v3/clusters/3W-2oFuqSYuErBzrRrTKyw/topics/bob/partitions/-/reassignment"
  },
  "authorized_operations": []
}

```

Error messages are shown as returned by kafka

```
{"error_code":40002,"message":"Topic already has 2 partitions."}%  
{"error_code":40002,"message":"Topic currently has 2 partitions, which is higher than the requested 1."}% 
```

I don't think we have a PATCH call anywhere else in the code.  I think it's right here - we are modifying but not fully replacing an existing topic object's information.

The flip of that could be that it should be a PUT as we are fully and idempotently replacing the whole of the topic's partition_count object, but patch felt more future safe, in case we add other things to the request body in future?

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -2448,9 +2448,9 @@ components:
     UpdatePartitionCountRequestData:
       type: object
       required:
-        - partition_count
+        - partitions_count
       properties:
-        partition_count:
+        partitions_count:
           type: integer
           format: int32
 
@@ -2675,7 +2675,7 @@ components:
             schema:
               $ref: "#/components/schemas/UpdatePartitionCountRequestData"
             example: 
-               partition_count: 10
+               partitions_count: 10
 
 
   responses:
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -898,6 +898,29 @@ paths:
         '5XX':
           $ref: '#/components/responses/ServerErrorResponse'
 
+    patch:
+      summary: 'Update partition count'
+      operationId: updatePartitionCountKafkaV3Topic
+      description: |-
+        [![Generally Available](https://img.shields.io/badge/Lifecycle%20Stage-Generally%20Available-%2345c6e8)](#section/Versioning/API-Lifecycle-Policy)
+
+        Increases number of partitions on a topic.  
+      tags:
+        - Topic (v3)
+      requestBody:
+        $ref: '#/components/requestBodies/UpdatePartitionCountRequest'
+      responses:
+        '200':
+          $ref: '#/components/responses/GetTopicResponse'
+        '400':
+          $ref: '#/components/responses/BadRequestErrorResponse'
+        '401':
+          $ref: '#/components/responses/UnauthorizedErrorResponse'
+        '429':
+          $ref: '#/components/responses/TooManyRequestsErrorResponse'
+        '5XX':
+          $ref: '#/components/responses/ServerErrorResponse'
+        
     delete:
       summary: 'Delete Topic'
       operationId: deleteKafkaV3Topic
@@ -2422,6 +2445,15 @@ components:
           type: string
           nullable: true
 
+    UpdatePartitionCountRequestData:
+      type: object
+      required:
+        - partition_count
+      properties:
+        partition_count:
+          type: integer
+          format: int32
+
   requestBodies:
     AlterBrokerConfigBatchRequest:
       description: 'The alter broker configuration parameter batch request.'
@@ -2499,7 +2531,6 @@ components:
                 host: '*'
                 operation: 'READ'
                 permission: 'ALLOW'
-
 
     CreateTopicRequest:
       description: 'The topic creation request.'
@@ -2636,6 +2667,16 @@ components:
             $ref: "#/components/schemas/UpdateConfigRequestData"
           example:
             value: 'gzip'
+
+    UpdatePartitionCountRequest:
+      description : 'The number of partitions to increase the partition count to.'
+      content:
+         application/json:
+            schema:
+              $ref: "#/components/schemas/UpdatePartitionCountRequestData"
+            example: 
+               partition_count: 10
+
 
   responses:
     CreateTopicResponse:
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
@@ -27,6 +27,7 @@ import io.confluent.kafkarest.entities.v3.CreateTopicRequest.ConfigEntry;
 import io.confluent.kafkarest.entities.v3.CreateTopicResponse;
 import io.confluent.kafkarest.entities.v3.GetTopicResponse;
 import io.confluent.kafkarest.entities.v3.ListTopicsResponse;
+import io.confluent.kafkarest.entities.v3.PartitionCountRequest;
 import io.confluent.kafkarest.entities.v3.Resource;
 import io.confluent.kafkarest.entities.v3.ResourceCollection;
 import io.confluent.kafkarest.entities.v3.TopicData;
@@ -52,6 +53,7 @@ import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.NotFoundException;
+import javax.ws.rs.PATCH;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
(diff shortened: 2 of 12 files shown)


api/v3/openapi.yaml
```diff
@@ -898,6 +898,29 @@ paths:
...
+
+        Increases number of partitions on a topic.  
+      tags:
+        - Topic (v3)
+      requestBody:
+        $ref: '#/components/requestBodies/UpdatePartitionCountRequest'
+      responses:
+        '200':
+          $ref: '#/components/responses/GetTopicResponse'
+        '400':
```
msn-tldr, line 915: Q: no-where do i see, these error-response code being generated in the code below?
How is exception handling done? What if admin-client throws?
PS: I vaguely remember from one of discussions before, that there is some common rest-utils/kafka-rest code to handle exception-handling <-> error code discussion.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
```diff
@@ -132,6 +134,37 @@ public void getTopic(
     AsyncResponses.asyncResume(asyncResponse, response);
   }
 
+  @PATCH
+  @Path("/{topicName}")
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Produces(MediaType.APPLICATION_JSON)
+  @PerformanceMetric("v3.topics.partitions")
+  @ResourceName("api.v3.topics.partitions")
+  public void updatePartitionCount(
```
msn-tldr, line 143: nit: It would be good to have a UT here, can be done as a follow-up.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/TopicManagerImplTest.java
```diff
@@ -846,6 +851,21 @@ public void deleteTopic_nonExistingCluster_throwsNotFound() throws Exception {
     }
   }
 
+  @Test
+  public void alterTopicPartitions() {
```
msn-tldr, line 855: nit: I quite like the convention of test method naming used in this file, methodNameUnderTest_WhenInputIsThis_ThenOutputIsThis.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/TopicManagerImplTest.java
```diff
@@ -451,6 +454,8 @@ public class TopicManagerImplTest {
 
   @Mock private DeleteTopicsResult deleteTopicsResult;
 
+  @Mock CreatePartitionsResult createPartitionsResult;
```
msn-tldr, line 457: should this be private?

msn-tldr:
> I don't think we have a PATCH call anywhere else in the code. I think it's right here - we are modifying but not fully replacing an existing topic object's information.

This does make sense.

msn-tldr (commented):
Minor nits.

msn-tldr (approved):

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
```diff
@@ -132,6 +134,37 @@ public void getTopic(
...
+    CompletableFuture<GetTopicResponse> response =
+        topicManager
+            .get()
+            .updateTopicPartitionCount(topicName, partitionCount.getPartitionCount())
+            .thenCompose(
+                nothing ->
+                    topicManager
+                        .get()
+                        .getTopic(clusterId, topicName)
+                        .thenApply(topic -> topic.orElseThrow(NotFoundException::new))
```
dimitarndimitrov, line 162: Is this practically guaranteed not to throw if the update operation succeeded? It should still stay there to address edge cases like racing delete and update, I just want to understand the failure path.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
```diff
@@ -132,6 +134,37 @@ public void getTopic(
     AsyncResponses.asyncResume(asyncResponse, response);
   }
 
+  @PATCH
```
dimitarndimitrov, line 137: Re: PATCH vs PUT
I really like the PATCH approach. My take is that unlike with updating topic configs, where we use a PUT and access a resource-like endpoint related to the modified topic config, here the API is designed to access the topic resource while providing a modification for part of the topic - its partition count. PUT would be plain incorrect with the model proposed here, but in addition to that PATCH seems to be a really good fit. My only worry is that PATCH is not as popular, understood, and maybe supported, but it has been out there for a long time now (RFC 5789 is 10+ years old) so this shouldn't be a problem.

dimitarndimitrov (approved):
+1.
I think we should add a couple of integration tests and we'll be golden.

api/v3/openapi.yaml
ehumber: Yes, it's handled at the rest-utils/jetty layer

eg
```
 curl -X PATCH http://localhost:8082/v3/clusters/3W-2oFuqSYuErBzrRrTKyw/topics/bob  -H "Content-Type: application/json" -d '{"partition_count123":"16"}'  | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   175    0   148  100    27    701    127 --:--:-- --:--:-- --:--:--   829
{
  "error_code": 400,
  "message": "Cannot construct instance of `io.confluent.kafkarest.entities.v3.PartitionCountRequest`, problem: Null partitionCount"
```

429 comes from the jetty dos rate limiter/filter or from the Guava/Resiliance4J rate limiters

401 comes from kafka returning a not authed and an exception handler translates this to a 401

And unexpected stuff gets translated by an exception handler to 500 errors.

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
ehumber: Ooo, I had some of them, they missed the git add.  Added now :)

ehumber:
> +1. I think we should add a couple of integration tests and we'll be golden.

Yup, totally agree.  Half written but fighting back atm.  We may still need this PR in in a hurry, so will do it as a follow up.

api/v3/openapi.yaml
```diff
@@ -898,6 +898,29 @@ paths:
         '5XX':
           $ref: '#/components/responses/ServerErrorResponse'
 
+    patch:
+      summary: 'Update partition count'
+      operationId: updatePartitionCountKafkaV3Topic
+      description: |-
+        [![Generally Available](https://img.shields.io/badge/Lifecycle%20Stage-Generally%20Available-%2345c6e8)](#section/Versioning/API-Lifecycle-Policy)
```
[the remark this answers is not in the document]
ehumber: @axisc Thoughts?  I think this is simple enough to go straight to GA, but it's safer to EA first.  Do Terraform's requirements have an impact on this decision?

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
ehumber: Yeah, I did go through that thought process, it's borrowed straight from the get topic code but I thought to leave it matching.  But you're right, there's only a really tiny timing window where the topic could be deleted.  Overkill though, so happy to take it off.

PR #1087, opened 2022-11-29
KREST-3009 Add the ability to increase num.partitions on a topic

ehumber:
Squashed commit so I can cherry pick more easily to the cloud release branch

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -2448,9 +2448,9 @@ components:
     UpdatePartitionCountRequestData:
       type: object
       required:
-        - partition_count
+        - partitions_count
       properties:
-        partition_count:
+        partitions_count:
           type: integer
           format: int32
 
@@ -2675,7 +2675,7 @@ components:
             schema:
               $ref: "#/components/schemas/UpdatePartitionCountRequestData"
             example: 
-               partition_count: 10
+               partitions_count: 10
 
 
   responses:
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -898,6 +898,29 @@ paths:
         '5XX':
           $ref: '#/components/responses/ServerErrorResponse'
 
+    patch:
+      summary: 'Update partition count'
+      operationId: updatePartitionCountKafkaV3Topic
+      description: |-
+        [![Generally Available](https://img.shields.io/badge/Lifecycle%20Stage-Generally%20Available-%2345c6e8)](#section/Versioning/API-Lifecycle-Policy)
+
+        Increases number of partitions on a topic.  
+      tags:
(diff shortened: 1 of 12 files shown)


kafka-rest/pom.xml
```diff
@@ -156,6 +156,12 @@
             <artifactId>hamcrest-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.glassfish.jersey.connectors</groupId>
+            <artifactId>jersey-apache-connector</artifactId>
+            <version>2.34</version>
+            <scope>test</scope>
+        </dependency>
```
dimitarndimitrov, line 164: That's needed for the testing of the `PATCH` method, right? Can we skip the version if we're inheriting it from the parent POM? Feel free to ignore as this is being fast-tracked, so I don't want to request unwanted changes that aren't too important.

kafka-rest/src/main/java/io/confluent/kafkarest/entities/v3/PartitionCountRequest.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2020 Confluent Inc.
```
dimitarndimitrov, line 2: nit: 2020 -> 2022

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
```diff
@@ -132,6 +134,36 @@ public void getTopic(
...
+    CompletableFuture<GetTopicResponse> response =
+        topicManager
+            .get()
+            .updateTopicPartitionCount(topicName, partitionCount.getPartitionCount())
+            .thenCompose(
+                nothing ->
+                    topicManager
+                        .get()
+                        .getTopic(clusterId, topicName)
+                        .thenApply(topic -> GetTopicResponse.create(toTopicData(topic.get()))));
```
dimitarndimitrov, line 162: nit: For the record I am OK with having a check here for the topic not existing race condition - there is a theoretical case that it would address. Feel free to ignore as this is being fast-tracked, so I don't want to request unwanted changes that aren't too important.

dimitarndimitrov (approved):
+1.

api/v3/openapi.yaml
```diff
@@ -2422,6 +2445,15 @@ components:
           type: string
           nullable: true
 
+    UpdatePartitionCountRequestData:
+      type: object
+      required:
+        - partition_count
```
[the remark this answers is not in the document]
ehumber: Good spot indeed.  I thought I'd made them match and I had not.

PR #1090, opened 2022-12-05
KREST-8518 Add dry-run/validate-only versions of some topic APIs

dimitarndimitrov:
This change introduces an additional "validate only" parameter to the requests for two important topic APIs:
- Create Topic
- Batch Alter Topic Config

When this parameter, which is a boolean flag, is set, the corresponding operation would be dry-run, i.e. its execution would be validated as if its actually run, but there would be no side effects from the operation. E.g. if run in that mode, a Create Topic request won't actually create a topic.

As this requires just using simple overloads of the corresponding Admin APIs, this change mostly focuses on introducing the new parameters in minimally disruptive way in exchange for some verbosity and maybe even redundancy.
- E.g. none of the existing tests had to be changed, including the resource test, which expect specific manager calls, and the manager impl tests, which expect specific Admin Client calls.

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -854,12 +854,17 @@ paths:
         [![Generally Available](https://img.shields.io/badge/Lifecycle%20Stage-Generally%20Available-%2345c6e8)](#section/Versioning/API-Lifecycle-Policy)
 
         Creates a topic.
-        Also supports a dry-run mode that only validates whether the topic creation would succeed.
+        Also supports a dry-run mode that only validates whether the topic creation would succeed
+        if the ``validate_only`` request property is explicitly specified and set to true.
       tags:
         - Topic (v3)
       requestBody:
         $ref: '#/components/requestBodies/CreateTopicRequest'
       responses:
+        # returned when dry-run mode is being used and a topic has not been created
+        '200':
+          $ref: '#/components/responses/CreateTopicResponse'
+        # returned in regular mode when a topic has been created
         '201':
           $ref: '#/components/responses/CreateTopicResponse'
         '400':
@@ -985,12 +990,14 @@ paths:
         [![Generally Available](https://img.shields.io/badge/Lifecycle%20Stage-Generally%20Available-%2345c6e8)](#section/Versioning/API-Lifecycle-Policy)
 
         Updates or deletes a set of topic configs.
-        Also supports a dry-run mode that only validates whether the operation would succeed.
+        Also supports a dry-run mode that only validates whether the operation would succeed if the
+        ``validate_only`` request property is explicitly specified and set to true.
       tags:
         - Configs (v3)
       requestBody:
         $ref: '#/components/requestBodies/AlterTopicConfigBatchRequest'
       responses:
+        # returned in both regular and dry-run modes
         '204':
           description: 'No Content'
         '400':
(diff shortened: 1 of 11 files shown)


brianstrauch (approved):
Thank you!

kafka-rest/src/test/java/io/confluent/kafkarest/integration/v3/TopicsResourceIntegrationTest.java
```diff
@@ -376,6 +376,76 @@ public void createTopic_nonExistingTopic_returnsCreatedTopic() {
...
+            .accept(MediaType.APPLICATION_JSON)
+            .post(
+                Entity.entity(
+                    "{\"topic_name\":\""
+                        + topicName
+                        + "\",\"partitions_count\":1,"
+                        + "\"replication_factor\":1,"
+                        + "\"validate_only\":true}",
+                    MediaType.APPLICATION_JSON));
+    assertEquals(Status.CREATED.getStatusCode(), response.getStatus());
```
ehumber, line 436: Is there anything you can do about the CREATED status code that's returned on the validated, as there is no object created here?

api/v3/openapi.yaml
```diff
@@ -1610,6 +1612,8 @@ components:
                   - SET
                   - DELETE
                 nullable: true
+        validate_only:
```
ehumber, line 1615: Reminder that you'll need to update the both the cc and cp consolidated-openapi.yaml in ce-kafka-rest for the CLI to be able to use these changes

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/AbstractConfigManager.java
```diff
@@ -188,6 +191,43 @@ final CompletableFuture<Void> safeAlterConfigs(
...
 
+  /**
+   * Atomically alter configs according to {@code commands}, checking if the configs exist first. If
+   * the {@code validateOnly} flag is set, the operation is only dry-ran (the configs do not get
+   * altered as a result).
+   */
+  // KREST-8518 A separate overload is provided instead of changing the pre-existing
+  // safeAlterConfigs method in order to minimize any risks related to external usage of that method
+  // (as this manager can be injected in projects inheriting from kafka-rest) and to minimize the
+  // amount of necessary changes (e.g. by avoiding the need to heavily refactor tests).
```
ehumber, line 202: I don't quite follow (even with the comment!) why the original safeAlterConfigs can't have its body deleted, and replaced with 

`return safeAlterConfigs(clusterId, resourceId, prototype, commands, false)`

The tests shouldn't need to change, they can still call the original method.

ehumber (approved):
Approved, but please review comments before merging

dimitarndimitrov:
The failing PR CI reports downstream build failures in `ce-kafka-rest` and `blueway`. When you look closer into it though, the `blueway` failure happens because of the `ce-kafka-rest` failure (as the former's Maven reactor fails then), and the `ce-kafka-rest` failure happens because of Checkstyle initialization problem seemingly related to the import control applied in `kafka-rest`:
```
[2022-12-05T21:08:04.975Z] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.2:check (validate) on project ce-kafka-rest-parent: Failed during checkstyle configuration: cannot initialize module TreeWalker - cannot initialize module ImportControl - illegal value 'checkstyle/import_control.xml' for property 'file': com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to find: checkstyle/import_control.xml -> [Help 1]
```

The Checkstyle problem is something that has popped up from time to time before, but we never had the time to investigate it in depth, not least because of its finicky and intermittent nature.
So all of these errors seem very much unrelated to the changes here and I'm not going to consider them blocking.
Nevertheless it is a good question to ask "What's the reason behind that Checkstyle problem"?

dimitarndimitrov:
> merge it and monitor the status of kafka-rest and its expected downstream dependencies

It turned out that the Checkstyle issue was hiding a problem, which was addressed in #1093.

P.S. We should answer that "What's the reason behind that Checkstyle problem" question soon. E.g. does it happen when a specific type of build failure occurs? Does it happen every time now?

api/v3/openapi.yaml
dimitarndimitrov: Thank you, filed KREST-8552.

trnguyencflt:
@dimitarndimitrov 

> "What's the reason behind that Checkstyle problem"

Look like this problem comes up more often now 😢 it is very hard to find the root cause locally because it fails intermittently in Jenkins

PR #1093, opened 2022-12-07
Address a coincidence causing a build breakage

dimitarndimitrov:
ce-kafka-rest happens to have a config manager that has defined on its own a validateOnly version of safeAlterConfigs, so introducing one in the base config manager (with a final modifier) causes a breakage.

For now let's avoid the breakage by renaming the base method and avoiding the unwanted override, and later we can consider whether it's better to remove the ce-kafka-rest method and use the base method instead.
- The problem could also be addressed by just removing the `final` modifier of the newly added base method so that an override does happen, but that seems slightly more invasive than just isolating the newly added base method.
- The ce-kafka-rest downstream validation will probably still fail because of a new unrelated problem that sneaked in the last hours.

Tested by local builds of kafka-rest and ce-kafka-rest.


msn-tldr:
Thanks for fixing!

PR #1095, opened 2022-12-15
Krest 8353 per lkc rate limiting admin endpoints

trnguyencflt:
This PR make sure that endpoints are rate limited based on LKC.

Tested it locally, will test this in CPD and Devel.


kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/FixedCostRateLimitFeature.java
```diff
@@ -77,4 +94,19 @@ private int getCost(ResourceInfo resourceInfo) {
     }
     return defaultCost;
   }
+
+  private static final class RequestRateLimiterCacheLoader
```
ehumber, line 98: We've got two of these now - the other one is in ProduceRateLimiters.

I wonder if we should pull it out to be in its own class.

No strong feelings though, you decide.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/FixedCostRateLimitRequestFilter.java
```diff
@@ -15,26 +15,43 @@
...
+    this.perClusterRateLimiterCache = requireNonNull(perClusterRateLimiterCache);
   }
 
   @Override
   public void filter(ContainerRequestContext requestContext) {
-    rateLimiter.rateLimit();
+    String clusterId = requestContext.getUriInfo().getPathParameters(true).getFirst("clusterId");
+    if (clusterId != null) {
+      RequestRateLimiter rateLimiter = perClusterRateLimiterCache.getUnchecked(clusterId);
+      rateLimiter.rateLimit(cost);
```
ehumber, line 51: I think we still need to generic rate limit here too.

We have eg a global total of 25 requests/s

If we have 100 different LKCs all hitting us with eg 10 requests/s.  We'll blow the global limit.  

I know the dos filter does 25 requests/s so this is a bit redundant atm, but I see the dos limiter going up to eg 100 requests/s soon, and this generic limiter limiting admin requests to 25/s and the produce limiters allowing the other 75/s

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RateLimitModule.java
```diff
@@ -37,6 +37,12 @@ protected void configure() {
         .to(RequestRateLimiter.class)
         .in(Singleton.class);
 
+    bindFactory(RequestRateLimiterGenericFactory.class)
```
ehumber, line 40: I think you'll need a new factory RequestRateLimiterPerLKCFactory here, like we do for the producer global vs per lkc, as you need to be able to set different limits on each

Eg global could be 25, but per lkc could be 10.

ehumber (changes requested):
See comments

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/FixedCostRateLimitFeature.java
trnguyencflt: @ehumber yeah, I was also thinking about that, my intention for putting this in private class because I thought the intention of the previous author is to hide this class `RequestRateLimiterCacheLoader`.

If you think it is necessary to do so to reduce code repetition, I'm happy to make it a standalone class.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/FixedCostRateLimitRequestFilter.java
trnguyencflt: this make sense 👍 I'll adjust it accordingly.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/RateLimitModule.java
trnguyencflt: That makes sense, I'll add a new factory

trnguyencflt:
@ehumber thanks for your comments, I have resolved those accordingly 😄

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/FixedCostRateLimitRequestFilter.java
msn-tldr: @trnguyencflt i think this should be documented in a comment. Better yet maybe add a wiki on latest state of rate-limiting in kREST(global-limits, per lkc limits, per ip limits, admin vs produce limits etc.), then u can link that in the comment. That wiki could also link to important places in code where this all setup. That should help with the review of this PR as well :) What do you think?

PS - I do see a [wiki](https://confluentinc.atlassian.net/wiki/spaces/KREST/pages/2939096773/Rate+Limiting) exists, maybe u can update that.

trnguyencflt: @msn-tldr thanks for the comment, that work is planned in this ticket https://confluentinc.atlassian.net/browse/KREST-8608 😄 

As this PR is not yet finalised, so I would delay the documentation update until this is merged.

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/FixedCostRateLimitFeature.java
ehumber: No strong feelings either way.  My only thought is, is this code likely to cause a problem/change/need updating, at which point there is a risk we forget to fix it in both places.  But I think that risk is low.

ehumber (approved):

kafka-rest/src/main/java/io/confluent/kafkarest/ratelimit/FixedCostRateLimitFeature.java
ehumber: I just saw, you did this already :)

PR #1097, opened 2023-01-04
KREST-8639 Only call get on topic manager provider once

ehumber:
Calling get on a topic manager twice, when running in Confluent cloud, results in a Out Of Scope Exception

I'm not sure why this is - I wonder if it runs slower in this environment and as the call is async, the scope has been lost by the time the first async call to update the partition count returns, ready for us to get the topic information in the response.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/TopicsResource.java
@@ -69,14 +69,14 @@ import javax.ws.rs.core.Response.Status;
 @ResourceName("api.v3.topics.*")
 public final class TopicsResource {
 
-  private final Provider<TopicManager> topicManager;
+  private final Provider<TopicManager> topicManagerProvider;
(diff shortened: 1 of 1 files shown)


trnguyencflt (approved):
LGTM

dimitarndimitrov:
@ehumber My guess is that this would happen not just for Confluent Cloud with the thread calling the second `get()` being the one handling the result of the Admin Client call which by then is not one of our Jetty/Jersey request handling threads that contain the original request scope.

ehumber:
> @ehumber My guess is that this would happen not just for Confluent Cloud with the thread calling the second `get()` being the one handling the result of the Admin Client call which by then is not one of our Jetty/Jersey request handling threads that contain the original request scope.

I ran embedded ce-kafka-rest locally, both as a non-docker jvm and using the docker-compose, neither saw it.

I do wonder if standalone docker-compose ce-kafka-rest pointing at a cloud backend might do.

Anyway, this fixed it in cloud, and matches the pattern of a single get call we do everywhere else.

msn-tldr:
LGTM.

PR #1098, opened 2023-01-06
Krest 2746 reflect kafka slow back to clients

ehumber:


ehumber:
Duplicate of https://github.com/confluentinc/kafka-rest/pull/1043 from a non-private repo so that it's not deleted when I leave.

PR #1101, opened 2023-01-13
Fix CVE-2022-41854 in dependency org.yaml_snakeyaml, upgrade to fixed…

trnguyencflt:
… version 1.32

PR #1102, opened 2023-01-16
Fix vulnerable transitive dependency of kafka-schema-registry

trnguyencflt:


TemirAskhat:
Isn't the code duplicating https://github.com/confluentinc/kafka-rest/pull/1101/files ?

trnguyencflt:
@TemirAskhat it is slightly different:
- in 6.0.x branch, we already have dependencyManagement block
- and there doesn't exist kafka-schema-registry-client dep

TemirAskhat (approved):

PR #1115, opened 2023-02-03
fix test

msn-tldr:
Run test as `mvn test -Dtest=SchemaManagerImplTest#errorFetchingSchemaBySchemaVersion`

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
@@ -144,7 +144,7 @@ final class SchemaManagerImpl implements SchemaManager {
           e);
     }
 
-    if (schema.getSchema() == null) {
+    if (schema == null || schema.getSchema() == null) {
       throw Errors.invalidSchemaException(
           String.format(
               "Error when fetching schema by version. subject = %s, version = %d",
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
@@ -144,6 +144,13 @@ final class SchemaManagerImpl implements SchemaManager {
           e);
     }
 
+    if (schema.getSchema() == null) {
+      throw Errors.invalidSchemaException(
+          String.format(
+              "Error when fetching schema by version. subject = %s, version = %d",
+              actualSubject, schemaVersion));
+    }
+
     SchemaProvider schemaProvider;
     try {
       schemaProvider = EmbeddedFormat.forSchemaType(schema.getSchemaType()).getSchemaProvider();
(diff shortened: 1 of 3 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -154,15 +154,22 @@ private RegisteredSchema getSchemaFromSchemaVersion(
 
     ParsedSchema parsedSchema;
     try {
+      Optional<ParsedSchema> opSchema = schemaProvider.parseSchema(schema, /* isNew= */ false);
```
msn-tldr, line 157: @rayokota Weirdly enough - 

line 158 & 163 - shows OpSchema is empty, contains null.
BUT line 162 & 166, show OpSchema is NOT empty. This is super weird!!! (**Why is this happening?**)

Since `orElseThrows`, at line 166, DOESN'T consider the optional as non-null, it doesn't throw at line 166, which earlier the test used to. And then it returns NullPointerException at line 171.

[text.txt](https://github.com/confluentinc/kafka-rest/files/10579766/text.txt)
Attaching log output from running the test.

rayokota: Ok, instead of changing the test, let's change`SchemaManagerImpl`.  On line 146, after getting the `schema`, have `SchemaManagerImpl` throw an exception if `schema.getSchema()` is null.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
```diff
@@ -144,6 +144,13 @@ private RegisteredSchema getSchemaFromSchemaVersion(
           e);
     }
 
+    if (schema.getSchema() == null) {
```
msn-tldr, line 147: @rayokota is this the expected change based on latest schema-registry change?

rayokota:
```suggestion
    if (schema == null) {
```

msn-tldr: @rayokota this wouldn't work, as the test returns a non-null schema object.
https://github.com/confluentinc/kafka-rest/blob/040e2a9ab85e3f0fe9d8c0545b03ee1474c97d4f/kafka-rest/src/test/java/io/confluent/kafkarest/controllers/SchemaManagerImplTest.java#L697

rayokota: ok, how about  `if (schema == null || schema.getSchema() == null)`

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaManagerImpl.java
msn-tldr: @rayokota its done.

rayokota (approved):
LGTM

msn-tldr:
CI Job failed, rerunning it, I believe its transient because - 

It fails with this
```
[2023-02-09T11:58:30.893Z] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.2:check (validate) on project ce-kafka-rest-parent: Failed during checkstyle configuration: cannot initialize module TreeWalker - cannot initialize module ImportControl - illegal value 'checkstyle/import_control.xml' for property 'file': com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to find: checkstyle/import_control.xml -> [Help 1]

But https://jenkins.confluent.io/job/confluentinc/job/ce-kafka-rest/job/master/ are succeeding.

TemirAskhat (approved):
LGTM

PR #1119, opened 2023-02-27
KREST-1797 - CreateTopic returns partitions and repl factor

AndrewJSchofield:
Previously the response for v3 CreateTopic only contained the partition count and replication factor if they were supplied in the request. Now, this information is returned in the response always. This mirrors the behaviour of the Kafka AdminClient.createTopics.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
@@ -65,26 +65,10 @@ public interface TopicManager {
 
   /**
    * Creates a new Kafka {@link Topic} with either partitions count and replication factor or
-   * explicitly specified partition-to-replicas assignments.
-   */
-  CompletableFuture<Void> createTopic(
-      String clusterId,
-      String topicName,
-      Optional<Integer> partitionsCount,
-      Optional<Short> replicationFactor,
-      Map<Integer, List<Integer>> replicasAssignments,
-      Map<String, Optional<String>> configs);
-
-  /**
-   * Creates a new Kafka {@link Topic} with either partitions count and replication factor or
    * explicitly specified partition-to-replicas assignments. If the {@code validateOnly} flag is
-   * set, the operation is only dry-ran (a topic does not get created as a result).
+   * set, the operation is only dry-run (a topic does not get created as a result).
    */
-  // KREST-8518 A separate overload is provided instead of changing the pre-existing createTopic
-  // method in order to minimize any risks related to external usage of that method (as TopicManager
-  // can be injected in projects inheriting from kafka-rest) and to minimize the amount of necessary
-  // changes (e.g. by avoiding the need to heavily refactor tests).
-  CompletableFuture<Void> createTopic(
+  CompletableFuture<Topic> createTopic(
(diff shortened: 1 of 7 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
```diff
@@ -63,28 +63,12 @@ CompletableFuture<Optional<Topic>> getTopic(
    */
   CompletableFuture<Optional<Topic>> getLocalTopic(String topicName);
 
-  /**
-   * Creates a new Kafka {@link Topic} with either partitions count and replication factor or
-   * explicitly specified partition-to-replicas assignments.
-   */
-  CompletableFuture<Void> createTopic(
```
msn-tldr, line 70: Is there a good reason to remove the overload? Below commit introduced the overload precisely to avoid breaking external clients.
https://github.com/confluentinc/kafka-rest/commit/3f1e7c1912d14f7175eda468b42b8072e0877a41#diff-15612e27820603f2efe1f27c21963e53492df05773f3c9b4489fc575582ced9c

msn-tldr:
Thanks for doing this, few comments above, otherwise LGTM!

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
AndrewJSchofield: It is more important to note that I changed the return type of the method. This is required in order to pass out the replication factor and number of partitions returned by the AdminClient.

So I took the hit of refactoring all the tests (which the referenced PR was avoiding), and I have also checked as well as I can that there are no references to this method outside this repo. By removing the overload and just ending up with the newer of the method signatures, we will soon be sure whether there are references that I missed.

AndrewJSchofield: @msn-tldr If you have an example of an external client using this interface, please tell me. I can see that the TopicManager methods for listing and getting topics are used outside, but I think that the createTopic is not.

msn-tldr: I don't, but its very hard to say for sure-ity no-one isn't using outside of Confluent.

Maybe mark the method as @Deprecated first, what do you think?

AndrewJSchofield: Done

jonchiu (approved):
Thanks for doing this

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManager.java
```diff
@@ -63,12 +63,25 @@ CompletableFuture<Optional<Topic>> getTopic(
    */
   CompletableFuture<Optional<Topic>> getLocalTopic(String topicName);
 
+  /**
+   * The original method to create a new Kafka {@link Topic} which was not able to pass back a
+   * result. Now superseded by {@code createTopic2}.
+   */
+  @Deprecated
```
msn-tldr, line 70: Thanks for doing this!

PR #1121, opened 2023-03-02
DOCS-18055 -  `List All Broker Configs` to `List Dynamic Broker Configs`

tobydrake7:
According to [this support ticket](https://confluent.zendesk.com/agent/tickets/122747) and this[ Docs Jira](https://confluentinc.atlassian.net/browse/DOCS-18055), GET broker configs returns only the dynamic configurations for the brokers.

The change rendered in HTML:

image

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -369,11 +369,11 @@ paths:
       - $ref: '#/components/parameters/ClusterId'
 
     get:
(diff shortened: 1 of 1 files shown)


londoncalling (approved):
LGTM, reviewing the API docs wiki page, not sure if it makes sense that these aren't building properly though, let's chat tomorrow.

jonchiu (approved):

PR #1126, opened 2023-03-09
Required query parms and errors for ACL operations

AndrewJSchofield:

PR #1127, opened 2023-03-09
Fixing build

apoorvmittal10:


kafka-rest/src/main/java/io/confluent/kafkarest/controllers/TopicManagerImpl.java
```diff
@@ -212,9 +213,7 @@ public CompletableFuture<Void> createTopic(
             replicasAssignments,
             configs,
             false)
-        .thenAccept(
-            unused -> {;
-            });
+        .thenAccept(unused -> {});
```
AndrewJSchofield, line 216: There is a fix for this in another PR (https://github.com/confluentinc/kafka-rest/pull/1126). Your code will not pass the checkstyle rules for a master build. I've been wrestling with this. :)

PR #1136, opened 2023-03-23
KREST-5711 - Non-negative partition id for produce

AndrewJSchofield:
The produce API endpoint can be used with or without specifying a partition ID. If a partition ID is specified, it must be non-negative. Before this change, a negative partition ID would result in an exception from the constructor for `ProducerRecord` which was translated into HTTP 500. But because it's a result of an incorrect request, it should be HTTP 4xx.

The change validates that a supplied partition ID is indeed non-negative before going anywhere near Kafka. The HTTP status code for a negative partition ID is 404.


jonchiu:
@AndrewJSchofield is there a test for this

trnguyencflt (approved):
LGTM

msn-tldr:
lgtm

PR #1138, opened 2023-03-27
KREST-6569 - Cope with partitions disappearing

AndrewJSchofield:
The `GET v3/clusters/{}/brokers/{}/partition-replicas` operation issues multiple calls to AdminClient. In the event that a topic is deleted between the first call and the later calls, it's possible that topic-partitions found in the first call are not found in the later calls. The appropriate action in this case is just to remove them from the response, rather than failing with a 404.


msn-tldr:
lgtm

PR #1157, opened 2023-05-02
KREST-4467 First part of reworking produce tests

AndrewJSchofield:
The v3 produce integration tests have been disabled because they prove unreliable in the CI system. Looking at the symptoms of the failing tests, the number of test cases and the complexity of the infrastructure are making it difficult to track down. This PR leaves the existing test present, but still disabled, and carves the non-schema tests and the rate-limit tests into their own integration tests. The batch tests run in about 1 second instead of nearer 60 seconds because the use of KafkaConsumer is much more efficient now.


trnguyencflt (approved):

AndrewJSchofield:
Build failed with another of the unreliable tests :)

AndrewJSchofield:
Unrelated test failure `ConsumerJsonTest.testConsumeWithMultipleParallelConsumers`.

PR #1158, opened 2023-05-04
KREST-8294 Permit charset for v3 produce

AndrewJSchofield:
This permits a Content-Type header that includes UTF-8 or ISO-8859-1 as a charset header parameters such as `Content-Type=application/json; charset=UTF-8`.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
@@ -63,7 +63,8 @@ public final class JsonStreamMessageBodyReader implements MessageBodyReader<Json
       if (charset != null) {
         if (!charset.equalsIgnoreCase("UTF-8") && !charset.equalsIgnoreCase("ISO-8859-1")) {
           throw new WebApplicationException(
-              "Unsupported charset in Content-Type header.", Status.BAD_REQUEST);
+              "Unsupported charset in Content-Type header. Supports \"UTF-8\" and \"ISO-8859-1\".",
+              Status.BAD_REQUEST);
         }
       }
     }
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
@@ -28,6 +28,7 @@ import javax.ws.rs.BadRequestException;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.ext.MessageBodyReader;
 
 /** A {@link MessageBodyReader} for {@link JsonStream}. */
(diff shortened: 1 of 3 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
```diff
@@ -41,7 +42,9 @@ public JsonStreamMessageBodyReader(ObjectMapper objectMapper) {
   @Override
   public boolean isReadable(
       Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
-    return JsonStream.class.equals(type) && MediaType.APPLICATION_JSON_TYPE.equals(mediaType);
+    return JsonStream.class.equals(type)
+        && MediaType.APPLICATION_JSON_TYPE.getType().equalsIgnoreCase(mediaType.getType())
+        && MediaType.APPLICATION_JSON_TYPE.getSubtype().equalsIgnoreCase(mediaType.getSubtype());
```
trnguyencflt, line 47: Not sure if this is correct but I think the third condition should be optional if there is subtype specified, something like
`&& (mediaType.getSubtype() == null || MediaType.APPLICATION_JSON_TYPE.getSubtype().equalsIgnoreCase(mediaType.getSubtype()))`

kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
```diff
@@ -41,7 +42,9 @@ public JsonStreamMessageBodyReader(ObjectMapper objectMapper) {
   @Override
   public boolean isReadable(
       Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
-    return JsonStream.class.equals(type) && MediaType.APPLICATION_JSON_TYPE.equals(mediaType);
+    return JsonStream.class.equals(type)
```
trnguyencflt, line 45: For my curiosity, why we don't use .equals method of MediaType class?

kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
```diff
@@ -53,6 +56,18 @@ public JsonStream<?> readFrom(
       MultivaluedMap<String, String> httpHeaders,
       InputStream entityStream)
       throws WebApplicationException {
+
+    // We know that the media type is 'application/json' but there might be an optional charset too
+    if (mediaType.getParameters() != null) {
+      String charset = mediaType.getParameters().get(MediaType.CHARSET_PARAMETER);
```
trnguyencflt, line 62: should we use `withCharset` method instead

kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStreamMessageBodyReader.java
```diff
@@ -53,6 +56,18 @@ public JsonStream<?> readFrom(
...
+      if (charset != null) {
+        if (!charset.equalsIgnoreCase("UTF-8") && !charset.equalsIgnoreCase("ISO-8859-1")) {
+          throw new WebApplicationException(
+              "Unsupported charset in Content-Type header.", Status.BAD_REQUEST);
```
trnguyencflt, line 66: Should we also provide the list of valid charsets in the exception?

AndrewJSchofield: That probably is a good point.

trnguyencflt (approved):
Approved with a very minor suggestion

PR #1159, opened 2023-05-04
KREST-10360 Remove flaky broker config check

AndrewJSchofield:
The value of the `num.network.threads` config in a test framework changed.


trnguyencflt (approved):
LGTM

PR #1161, opened 2023-05-11
Remove verbose log line

trnguyencflt:
As title said, this log line doesn't add value but adding verbosity to kafka rest because it will log every time a produce request is fired


AndrewJSchofield (approved):

msn-tldr (approved):
lgtm

PR #1163, opened 2023-05-15
KREST-10286 Example python code for produce streaming V3 API

msn-tldr:
This will use the HTTP connection in a fully duplex mode.

Following is the output of `streaming_produce_v3_main.py`, writing 5 records. It demonstrates the connection is being used in full-duplex mode, i.e. record-receipt is read for 1st record even before then 2nd record is written to the connection.
```
Establishing connection with headers: {'Content-Type': 'application/json'}
Waiting for http-connection to be established.
Sleeping for 1 second, before producing record #1
Connection established, will read responses.
Sleeping for 1 second, before producing record #2
Writing a record #1 with json b'{"value": {"type": "STRING", "data": "value_0"}, "key": {"type": "STRING", "data": "key_0"}}'
Http-stream has status-code 200
Receipt for record #1 is ******
"{\"error_code\":200,\"cluster_id\":\"EV-5o5e3SViiGP0hpgKn1g\",\"topic_name\":\"topic_1\",\"partition_id\":0,\"offset\":468,\"timestamp\":\"2023-05-16T11:02:08.322Z\",\"key\":{\"type\":\"STRING\",\"size\":5},\"value\":{\"type\":\"STRING\",\"size\":7}}"
Sleeping for 1 second, before producing record #3
Writing a record #2 with json b'{"value": {"type": "STRING", "data": "value_1"}, "key": {"type": "STRING", "data": "key_1"}}'
Receipt for record #2 is ******
"{\"error_code\":200,\"cluster_id\":\"EV-5o5e3SViiGP0hpgKn1g\",\"topic_name\":\"topic_1\",\"partition_id\":0,\"offset\":469,\"timestamp\":\"2023-05-16T11:02:09.323Z\",\"key\":{\"type\":\"STRING\",\"size\":5},\"value\":{\"type\":\"STRING\",\"size\":7}}"
Sleeping for 1 second, before producing record #4
Writing a record #3 with json b'{"value": {"type": "STRING", "data": "value_2"}, "key": {"type": "STRING", "data": "key_2"}}'
Receipt for record #3 is ******
"{\"error_code\":200,\"cluster_id\":\"EV-5o5e3SViiGP0hpgKn1g\",\"topic_name\":\"topic_1\",\"partition_id\":0,\"offset\":470,\"timestamp\":\"2023-05-16T11:02:10.334Z\",\"key\":{\"type\":\"STRING\",\"size\":5},\"value\":{\"type\":\"STRING\",\"size\":7}}"
[...]

PR #1164, opened 2023-05-16
Half duplex example code

msn-tldr:
This code shows that "traditional" http-libraries, idiomatic "request-response" model leads to entire http-request being written(with all producer-records in 1 go), and then reading http-response(i.e. record-receipts in 1 go).

Example output with 2 produce-records being written:
```
Establishing connection with headers: {'Content-Type': 'application/json'}
Sleeping for 1 second, before producing record #1
Sleeping for 1 second, before producing record #2
Writing a record #1 with json b'{"value": {"type": "STRING", "data": "value_0"}, "key": {"type": "STRING", "data": "key_0"}}'
Writing a record #2 with json b'{"value": {"type": "STRING", "data": "value_1"}, "key": {"type": "STRING", "data": "key_1"}}'
No more records to produce, exiting __record_generator
Done producing-records.
Done http-respones is available.
Http Response(with records receipts) is 
 b'{"error_code":200,"cluster_id":"EV-5o5e3SViiGP0hpgKn1g","topic_name":"topic_1","partition_id":0,"offset":456,"timestamp":"2023-05-16T10:54:15.888Z","key":{"type":"STRING","size":5},"value":{"type":"STRING","size":7}}\r\n{"error_code":200,"cluster_id":"EV-5o5e3SViiGP0hpgKn1g","topic_name":"topic_1","partition_id":0,"offset":457,"timestamp":"2023-05-16T10:54:16.892Z","key":{"type":"STRING","size":5},"value":{"type":"STRING","size":7}}\r\n'
 ```
 NOTE - i have kept different thread/queues etc. to be consistent with https://github.com/confluentinc/kafka-rest/pull/1163, otherwise it wasn't needed.

### DIFF
--- a/streaming_produce_v3_half_duplex.py
+++ b/streaming_produce_v3_half_duplex.py
@@ -1,0 +1,129 @@
+import json
+import queue
+import socket
+import time
+from threading import Thread
+import urllib3
+
+# Class that holds a records to be produced.
+class ProduceRecord:
+    def __init__(self, key, value):
+        self.key = key
+        self.value = value
+
+    def to_json(self):
+        ret = {
+            "value": {
+                "type": "STRING",
+                "data": self.value
+            },
+        }
+
+        if self.key:
+            ret["key"] = {
+                "type": "STRING",
+                "data": self.key,
+            }
+
+        return json.dumps(ret).encode("utf-8")
+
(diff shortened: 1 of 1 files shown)


trnguyencflt:
@msn-tldr what is the python version that can run this example?

PR #1166, opened 2023-05-30
KREST-10244 Fix produce-request-rate metric

msn-tldr:
Simple fix, correctly use "rate" metric to record "produce-request-rate", instead of "avg".

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
@@ -158,7 +158,7 @@ final class ProducerMetrics {
 
   private void setupRequestSensor(Map<String, String> metricsTags, String sensorTags) {
     Sensor requestSensor = createSensor(REQUEST_SENSOR_NAME, sensorTags);
-    addAvg(requestSensor, REQUEST_RATE_METRIC_NAME, REQUEST_RATE_METRIC_DOC, metricsTags);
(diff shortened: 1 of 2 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProducerMetrics.java
```diff
@@ -158,7 +158,7 @@ private void setupSensors(Map<String, String> metricsTags, String sensorTags) {
 
   private void setupRequestSensor(Map<String, String> metricsTags, String sensorTags) {
     Sensor requestSensor = createSensor(REQUEST_SENSOR_NAME, sensorTags);
-    addAvg(requestSensor, REQUEST_RATE_METRIC_NAME, REQUEST_RATE_METRIC_DOC, metricsTags);
```
trnguyencflt, line 161: @msn-tldr could we add a test for this?

trnguyencflt:
@msn-tldr good catch 😄 I'm a bit surprise that was there for a long time, could you add a test for this so that we can avoid this error in the future (suppose we create new sensor and adding wrong calculation).

kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProducerMetricsTest.java
```diff
@@ -54,6 +62,8 @@ public class ProducerMetricsTest {
   private ProducerMetrics producerMetrics;
   private KafkaRestConfig config;
 
+  private Map tags = ImmutableMap.of("tag", "value");
```
trnguyencflt, line 65: Map -> Map<String,String>

PR #1168, opened 2023-06-07
KREST-10243 Add custom-request-logging to kafka-rest, and log error-codes for when various rate-limiters are triggered

msn-tldr:
Following categories of changes are introduced in this PR.

1.  The new request-logger that will log custom metadata as part of the request-log.
* Adds `CustomLog.java`, as a new Jetty request-logger, that also logs  configured `request-attributes`.
Example:
Existing
`GET /kafka/v3/clusters/lkc-foobar/consumer-groups/foobar/lags HTTP/1.1" 429 408 "-" "Mozilla/5.0 (Windows NT; Windows NT 10.0; en-US) WindowsPowerShell/5.1.14393.5582`
Vs NOTE the new log has the error-code `429007` at the end, which is error-code defined in `RateLimitExceededException.java` class.
`GET /kafka/v3/clusters/lkc-foobar/consumer-groups/foobar/lags HTTP/1.1" 429 408 "-" "Mozilla/5.0 (Windows NT; Windows NT 10.0; en-US) WindowsPowerShell/5.1.14393.5582 429007`
2. Populate the relevant meta-data on request. This requires new `request-attribute` i.e. `REST_ERROR_CODE` to be populated when a rate-limit is being triggered. All rate-limiters, when triggered have a corresponding "error-code" in request-log, this logic is tested in `RestCustomRequestLogIntegrationTest.java`.
* `RateLimitExceededException.java` has the error-codes that will be logged when rate-limit is triggered.
* `GlobalDosFilterListener.java` & `PerConnectionDosfilterListener.java` are DosFilter.Listeners that on global & per-ip Jetty rate-limiters populate request-attribute needed for logging.
* `FixedCostRateLimitRequestFilter` has changes to log error-code when Jersey request rate-limits are triggered.
* Relevant changes to `ProduceRatelimiters` to log error-code when produce rate limits are triggered.
* Tests
3.  This new custom-logger can be toggled on/off via `use.custom.request.logging`. By default it's ON.


kafka-rest/pom.xml
```diff
@@ -178,6 +178,14 @@
 
     <build>
         <plugins>
+            <plugin>
```
msn-tldr, line 181: Exclude from checkstyle checks, as style in different in Jetty from where these are taken.

kafka-rest/src/main/java/io/confluent/kafkarest/CustomLogFields.java
```diff
@@ -0,0 +1,25 @@
...
+ * specific language governing permissions and limitations under the License
+ */
+
+package io.confluent.kafkarest;
+
+/**
+ * This class lists the fields that are added to the request-log(see RestCustomRequestLog.java).
+ * These field(& corresponding) value are stored as request-attribute.
+ */
+public class CustomLogFields {
```
trnguyencflt, line 22: This should be name something like `CustomRequestAttributeFields` I think,

also, as this is a utility static class, it should be a final class, and there should be a private constructor to prevent instantiation.

kafka-rest/src/main/java/io/confluent/kafkarest/GlobalDosFilterListener.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest;
```
trnguyencflt, line 16: could you move this into a package?

kafka-rest/src/main/java/io/confluent/kafkarest/CustomLogFields.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License
+ */
+
+package io.confluent.kafkarest;
```
trnguyencflt, line 16: could you move this into a sub package?

kafka-rest/src/main/java/io/confluent/kafkarest/PerConnectionDosFilterListener.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.kafkarest;
```
trnguyencflt, line 16: could you move this to a sub package as well

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,1067 @@
...
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License
+ */
+
+package io.confluent.kafkarest;
```
trnguyencflt, line 16: please move this to a sub package

kafka-rest/src/main/java/io/confluent/kafkarest/resources/v3/ProduceRateLimiters.java
trnguyencflt, line 1: @msn-tldr while I totally agree on logging code for admin api + dos filter, I think we might need to take a step back to evaluate logging produce rate limit.

- first, logging on produce rate limit could be very verbose, do we produce the same amount of log as if there isn't this custom logger?
- second, for produce, this is where we do have a details metrics on tenant level, so I wonder should this log be needed.
- third, what happen in the case on streaming produce, does this log on every `RateLimitExceededException` with the appropriate error code? Or one code will overwrite the other? Or just one single code is log.

Please let me know what do you think?

msn-tldr: > first, logging on produce rate limit could be very verbose, do we produce the same amount of log as if there isn't this custom logger?

So in case of produce, this would only add an extra error-code at the end of the existing request-log, so new log would like below, notice `429001` at the end.

`GET /kafka/v3/clusters/lkc-foobar/topics/topic_foobar/records HTTP/1.1" 200 712 "-" "Mozilla/5.0 (Windows NT; Windows NT 10.0; en-US) WindowsPowerShell/5.1.14393.5582 429001`

> second, for produce, this is where we do have a details metrics on tenant level, so I wonder should this log be needed.

Right it just makes the debugging easier, very evident for the oncall the issue is, especially considering logging-volumes are nearly the same.
Also note our produce metrics are `after` rate-limiting has been applied. Its hard to distinguish between a tenant "going" close the rate-limit, and tenant that has breached the limit. Also the returned http-status code is generic `429`, so at the same time if Global-jetty-dos-rate-limiter kicks in, and produce-tenant gets throttled because of produce-limit, oncall won't be able to tell if it was global-jetty limit(noisy neighbor) Vs in fact user going over the produce per-tenant limits etc. My point it just make the oncall lifes easier 😄 

> third, what happen in the case on streaming produce, does this log on every RateLimitExceededException with the appropriate error code? Or one code will overwrite the other? Or just one single code is log.

Right from the example above, the overall http status code is `200`, same as the user & consistent. But it will then log an aggregated(i.e. single) error-code 429001 to 429004, indicating which produce-rate-limit was triggered.

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,74 @@
...
+    if (requestAttributesToLog == null || requestAttributesToLog.length == 0) {
+      return;
+    }
+    for (String attr : this.requestAttributesToLog) {
+      Object attrVal = request.getAttribute(attr);
+      if (attrVal == null) {
+        continue;
+      }
+      logBuilder.append(" ");
+      logBuilder.append(attrVal);
```
trnguyencflt, line 57: I think you'll need to check if attribute is a String before appending

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,74 @@
...
+    this.formatString = formatString;
+    this.requestAttributesToLog = requestAttributesToLog;
+  }
+
+  @Override
+  public void log(final Request request, final Response response) {
+    try {
+      StringBuilder logStringBuilder = new StringBuilder();
+      RequestLog.Writer jettyLogWriter = new StringBuilderLogWriter(logStringBuilder);
+      RequestLog jettyLog = new CustomRequestLog(jettyLogWriter, formatString);
```
trnguyencflt, line 38: this approach is working for the purpose, however, I'll have a bit of concern about the performance, as for every log line, we will create a new `CustomRequestLog` and a new `StringBuilder` objects. While StringBuilder might be lightweight, `CustomRequestLog` is a big one.

What do you think?

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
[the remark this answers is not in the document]
msn-tldr: @trnguyencflt Its been struggle getting to re-use Jetty's `CustomRequestLog`. What do you think about this? 🤞

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,51 @@
...
+        request.removeAttribute(attr);
+        response.setHeader(attr, attrVal.toString());
+      }
+    }
+
+    jettyRequestLog.log(request, response);
+
+    // Remove the response-headers that were added above just for logging.
+    for (String attr : this.requestAttributesToLog) {
+      response.getHttpFields().remove(attr);
```
trnguyencflt, line 48: does this method throw if the header doesn't exist?

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,51 @@
...
+import org.eclipse.jetty.server.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * RestCustomRequestLog implements Jetty's RequestLog interface. It offers the same log-format
+ * jetty's CustomRequestLog. Additionally, it would append configured request-attributes (see
+ * requestAttributesToLog) to the end of the log.
+ */
+public class RestCustomRequestLog implements RequestLog {
```
trnguyencflt, line 15: I think we should make the name a little detailed, like `CustomRequestLogWithAttributes`

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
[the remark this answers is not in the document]
trnguyencflt: this version should work 👍 👏

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,51 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(RestCustomRequestLog.class);
+
+  private RequestLog jettyRequestLog;
```
trnguyencflt, line 19: this should be final and the name is delegate (as you are using composition)

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,51 @@
...
+      formatString += " %{" + attr + "}o";
+    }
+    this.requestAttributesToLog = requestAttributesToLog;
+    this.jettyRequestLog = new CustomRequestLog(writer, formatString);
+  }
+
+  @Override
+  public void log(Request request, Response response) {
+    // Add request-attributes to log as response-headers. Then these response-headers are logged
+    // in jetty's CustomRequestLog with format specifier "%{AttributeName}o".
```
trnguyencflt, line 35: Could you please add some comment like this:
`This is the workaround as there doesn't exist the ability in jetty's CustomRequestLog to log request attribute`

kafka-rest/pom.xml
```diff
@@ -173,6 +173,14 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <configuration>
+                    <excludes>**/RestCustomRequestLog.java,**/RestCustomRequestLogTest.java</excludes>
```
trnguyencflt, line 180: You don't this anymore

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -69,14 +86,68 @@ public KafkaRestApplication(KafkaRestConfig config, String path) {
...
     restResourceExtensions =
         config.getConfiguredInstances(
             KafkaRestConfig.KAFKA_REST_RESOURCE_EXTENSION_CONFIG, RestResourceExtension.class);
     config.setMetrics(metrics);
   }
 
+  @Override
+  public Handler configureHandler() {
+    // kafka-rest's custom-request-logging should be setup before calling super.configureHandler()
+    if (config.getBoolean(KafkaRestConfig.USE_CUSTOM_REQUEST_LOGGING_CONFIG)) {
```
trnguyencflt, line 115: This block is a bit weird, it should be in the constructor I think.

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,51 @@
...
+    for (String attr : this.requestAttributesToLog) {
+      Object attrVal = request.getAttribute(attr);
+      if (attrVal != null) {
+        request.removeAttribute(attr);
+        response.setHeader(attr, attrVal.toString());
```
trnguyencflt, line 40: maybe even better is setting the header of the request as we modify the request anyway (it can be log with `%{name}i`)

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
msn-tldr: Nope only return null

kafka-rest/pom.xml
msn-tldr: remove them

kafka-rest/src/main/java/io/confluent/kafkarest/RestCustomRequestLog.java
msn-tldr: I will rename to `CustomLog` to be concise, the logging of `request-attibutes` is implementation detail that comments will explain.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -69,14 +86,68 @@ public KafkaRestApplication(KafkaRestConfig config, String path) {
...
+      log.info(
+          "For rest-app with listener {}, configuring custom request logging", getListenerName());
+      if (this.requestLogWriter == null) {
+        Slf4jRequestLogWriter logWriter = new Slf4jRequestLogWriter();
+        logWriter.setLoggerName(config.getString(RestConfig.REQUEST_LOGGER_NAME_CONFIG));
+        try {
+          // This is required to correctly initialise Slf4jRequestLogWriter otherwise it throws NPE.
+          logWriter.start();
```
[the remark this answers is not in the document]
msn-tldr: Removed now. CustomLog needed to extend AbstractLifeCycle

msn-tldr:
@trnguyencflt the build-failure is not do with  `rest-utils` but very checkstyle issue - 
`ERROR] /home/jenkins/workspace/po_PR_builder_kafka-rest_PR-1168/kafka-rest/src/test/java/io/confluent/kafkarest/resources/v3/ProduceActionTest.java:40:8: Unused import - com.google.api.Http. [UnusedImports]`

Weirdly, the import is not in my local repo, only here. When i try to update the remote repo(even forcefully), github thinks everything is upto date, and so the unused import doesn't go away.

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -32,26 +32,42 @@
...
+import org.slf4j.LoggerFactory;
 
 /** Utilities for configuring and running an embedded Kafka server. */
 public class KafkaRestApplication extends Application<KafkaRestConfig> {
 
+  private static final Logger log = LoggerFactory.getLogger(KafkaRestApplication.class);
+
   List<RestResourceExtension> restResourceExtensions;
 
+  protected RequestLog.Writer requestLogWriter;
```
trnguyencflt, line 67: I don't think this is needed, and also this `requestLogFormat`

kafka-rest/src/main/java/io/confluent/kafkarest/requestlog/CustomLog.java
```diff
@@ -0,0 +1,94 @@
...
+ * implements LifeCycle. This in turn will correctly initialise the input RequestLog.Writer, if it
+ * also implements LifeCycle.
+ */
+public class CustomLog extends AbstractLifeCycle implements RequestLog {
+
+  private static final Logger log = LoggerFactory.getLogger(CustomLog.class);
+
+  private final CustomRequestLog delegateJettyLog;
+
+  private String[] requestAttributesToLog;
```
trnguyencflt, line 40: this could also be final

kafka-rest/src/main/java/io/confluent/kafkarest/requestlog/CustomLog.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Override
+  public void log(Request request, Response response) {
+    // Add request-attributes to log as response-headers. Then these response-headers are logged
+    // in jetty's CustomRequestLog with format specifier "%{AttributeName}o". This is workaround
+    // as the Jetty's CustomRequestLog doesn't have ability to log request-attributes. Whereas
+    // request-attributes are used to propagate custom-info to log, as that is more idiomatic. Also
+    // since request is more readily available Vs response(example - GlobalDosFilterListener), and
+    // the attributes are
```
trnguyencflt, line 71: this sentence seems incomplete, could you check this?

kafka-rest/src/main/java/io/confluent/kafkarest/requestlog/CustomLog.java
```diff
@@ -0,0 +1,94 @@
...
+      if (attrVal != null) {
+        request.removeAttribute(attr);
+        response.setHeader(attr, attrVal.toString());
+      }
+    }
+
+    try {
+      delegateJettyLog.log(request, response);
+    } catch (Exception e) {
+      log.info(
```
trnguyencflt, line 83: Should this be debug? Just in case of verbose logging.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/CustomLogIntegrationTest.java
```diff
@@ -0,0 +1,515 @@
...
+        .getDisplayName()
+        .contains("test_whenJettyGlobalRateLimitTriggered_ThenRequestLogHasRelevantInfo")) {
+      restConfigs.put("dos.filter.enabled", "true");
+      restConfigs.put("dos.filter.max.requests.per.sec", 1);
+      restConfigs.put("dos.filter.max.requests.per.connection.per.sec", 99999);
+    } else if (testInfo
+        .getDisplayName()
+        .contains("test_whenNoRateLimitTriggered_ThenRequestLogHasRelevantInfo")) {
+      // Set all rate-limits very high
+      restConfigs.put("dos.filter.enabled", "true");
```
trnguyencflt, line 102: can you use the constant from RestConfig class?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/CustomLogIntegrationTest.java
```diff
@@ -0,0 +1,515 @@
...
+        boolean added = logEntries.add(requestEntry);
+        assertTrue(added, "Failed to add entry to log, unexpected.");
+        log.info("Log #{}:{}", entryCounter.getAndIncrement(), requestEntry);
+      } catch (Exception e) {
+        e.printStackTrace();
+      }
+    }
+  }
+
+  private void hammerAtConstantRate(String path, Duration requestInterval, int totalRequests) {
```
trnguyencflt, line 234: could you move private function to the end of the class?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/CustomLogIntegrationTest.java
```diff
@@ -0,0 +1,515 @@
...
+      if (status != 200 && status != 429) {
+        fail(
+            String.format(
+                "Expected HTTP 200 or HTTP 429, but got HTTP %d instead: %s",
+                status, response.readEntity(String.class)));
+      }
+    }
+  }
+
+  private void verifyLog(
```
trnguyencflt, line 297: same here, please move to the end of the class.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/CustomLogIntegrationTest.java
```diff
@@ -0,0 +1,515 @@
...
+      int expectedNumOfEntries,
+      int errorCodeInLog,
+      int expectedRateLimitLogs,
+      boolean isProduce,
+      boolean isCustomLoggingDisabled) {
+    // Sleep so all request-logs are logged before reading and validating them here.
+    try {
+      TimeUnit.SECONDS.sleep(2);
```
trnguyencflt, line 305: this could potentially be the source of flakiness, is there another source to check that all requests have finished?

kafka-rest/src/main/java/io/confluent/kafkarest/requestlog/CustomLog.java
msn-tldr: I am now logging a simple log.error, indicating failure, which is OK since anyways would have logged "1 line per request". Followed by log.debug-ing the exception stack trace.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/CustomLogIntegrationTest.java
msn-tldr: Unfortunately constant is private, with no setter 😢

trnguyencflt: 😢

trnguyencflt (approved):
LGTM 👍

kafka-rest/src/test/java/io/confluent/kafkarest/integration/CustomLogIntegrationTest.java
trnguyencflt: That's fine

PR #1171, opened 2023-06-12
Remove the use of non-public API MockTime (in ce-kafka but not in AK)

trnguyencflt:
This PR removes MockTime dependency to avoid upstream issue.


AndrewJSchofield (approved):
I changed the pom.xml when I made the previous PR. I suggest reverting that too.

trnguyencflt:
@AndrewJSchofield reverted the change in the pom.xml

PR #1178, opened 2023-07-12
Review logging, some code tidy up

trnguyencflt:
This is as part of reviewing our logging, I have checked across the repository and confirm that we don't log configuration.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
@@ -47,7 +47,7 @@ public class DefaultKafkaRestContext implements KafkaRestContext {
   private SchemaRegistryClient schemaRegistryClient;
 
   public DefaultKafkaRestContext(KafkaRestConfig config) {
-    log.debug("Creating context with config: {}", config);
+    log.debug("Creating DefaultKafkaRestContext");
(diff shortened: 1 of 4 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/DefaultKafkaRestContext.java
```diff
@@ -47,7 +47,7 @@ public class DefaultKafkaRestContext implements KafkaRestContext {
   private SchemaRegistryClient schemaRegistryClient;
 
   public DefaultKafkaRestContext(KafkaRestConfig config) {
-    log.debug("Creating context with config: {}", config);
```
trnguyencflt, line 50: this doesn't actually log the configuration at the moment due to it use default toString() of Object, but changing this just in case.

AndrewJSchofield (approved):

kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java
```diff
@@ -113,7 +113,6 @@ private static RequestLog createRequestLog(
       KafkaRestConfig config,
       RequestLog.Writer requestLogWriter,
       String requestLogFormat,
-      Logger log,
```
trnguyencflt, line 116: log is a static object, so we don't need to pass to this function

PR #1183, opened 2023-07-20
Make sure that ErrorMessage is intepret as json type in response writer

trnguyencflt:
As the title says. The reason is that without explicit MediaType, we might encounter the below error with OPTIONS request

```
org.glassfish.jersey.message.internal.MessageBodyProviderNotFoundException: MessageBodyWriter not found for media type=text/plain, type=class io.confluent.rest.entities.ErrorMessage, genericType=class io.confluent.rest.entities.ErrorMessage
```


jonchiu (approved):
LGTM

PR #1196, opened 2023-08-15
Create custom Properties and Map configs classes with toString method that hides password properties

trnguyencflt:
This PR add two classes (inner class of `KafkaRestConfig`) that can be a direct replacement for  the two classes that are often used in creating configuration
- Properties
- Map<String, Object>

The two classes have overriden `toString` method that can hide sensitive configs. `toString` is the function that will get called in logging on the class.

This PR also uses the two new classes in all functions in `KafkaRestConfig` that return configurations so that they can be safely logged later. 

Reference JIRA: https://confluentinc.atlassian.net/browse/KREST-11068

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
@@ -1349,16 +1349,16 @@ public class KafkaRestConfig extends RestConfig {
   /**
    * This function relies on {@link AbstractConfig} object to detect whether a config is a {@link
    * Password}. Therefore, it is important for the implementation to make sure that a config is
-   * registered correctly to not accidentally leaking sensitive information. If there is config that
-   * is sensitive but not registered correctly in its corresponding {@link AbstractConfig} object,
-   * please add that config to {@link KafkaRestConfig#passwordTypeConfigs} field above.
+   * registered correctly to not accidentally leaking sensitive information. If there is a config
+   * that is sensitive but not registered correctly in its corresponding {@link AbstractConfig}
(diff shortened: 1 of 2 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java
```diff
@@ -1287,4 +1336,149 @@ private static ConfigValue create(String origin, Object value) {
...
+      Entry<K, V> entry = it.next();
+      K key = entry.getKey();
+      V value = entry.getValue();
+      if (key instanceof String) {
+        if (passwordTypeConfigs.contains(key)) {
+          toStringKeyAndValue(sb, key, Password.HIDDEN, it.hasNext());
+        } else {
+          Type type = config.typeOf((String) key);
+          if (type == Type.PASSWORD) {
+            toStringKeyAndValue(sb, key, config.getPassword((String) key), it.hasNext());
```
TemirAskhat, line 1371: So we are redacting only if and only if it's a part of "passwordTypeConfigs" set, otherwise, we are showing it as it is, is it an intended logic?

trnguyencflt: yeah, this is the logic, if our configurations are sane, then we should expect password type to be sensitive ones.

TemirAskhat: Got it!

TemirAskhat: Resolving the convo

TemirAskhat (approved):

PR #1199, opened 2023-09-12
Fixing the build error

TemirAskhat:
* TSIA 
* We can wait the jenkins build to be finished 

image


trnguyencflt:
@TemirAskhat please have a look at this https://confluent.slack.com/archives/CLQ1C50JU/p1694522162943149, I think there was a behaviour change in the serder, let's see how it goes.

TemirAskhat:
Status: waiting for the gov team, if they fix it or not on their side

PR #1200, opened 2023-09-14
KREST-10285 example rest produce v3 streaming in javascript (nodejs)

trnguyencflt:
As the title says, this is the example in javascript (nodejs) for rest produce v3 streaming


examples/produce_v3/python/streaming_produce_v3_main.py
trnguyencflt, line 1: also fix a couple of typos in python example comments.

examples/produce_v3/javascript/main.js
```diff
@@ -0,0 +1,52 @@
+// if the endpoint is https, please use https package
+const http = require('http');
+
+// handling streaming response
+var recordReceiptCounter = 1;
+callback = function(response) {
+    response.on(
+        'data',
+        function (data) {
+            let chunk = data.toString('utf8');
```
TemirAskhat, line 10: We are not changing chunk, so const?

examples/produce_v3/javascript/main.js
```diff
@@ -0,0 +1,52 @@
+// if the endpoint is https, please use https package
+const http = require('http');
+
+// handling streaming response
+var recordReceiptCounter = 1;
+callback = function(response) {
```
TemirAskhat, line 6: More of a [nit] 

I think it can be changed to the modern ECMASscript (starting from 6+), afaik, function usage is considered as an "old style" : ) 

So, something like this? 

```const callback = (response) => {
    response.on('data', (data) => {
        const chunk = data.toString('utf8');
        if (chunk.trim()) {
            console.log(`Receipt for record #${recordReceiptCounter} is`);
            console.log(JSON.parse(chunk));
            recordReceiptCounter += 1;
        }
    });

    response.on('end', () => {
        console.log("Stream is closed!");
    });
};```

**Review by msn-tldr — APPROVED** [review]
lgtm

File: examples/produce_v3/javascript/main.js [hunk 3]
```diff
@@ -0,0 +1,52 @@
...
+    });
+}
+
+async function main() {
+    // please update the below information to fit your Kafka Rest endpoint
+    const host = 'localhost';
+    const port = '8082';
+    const clusterId = 'ZXBWzl8VQ5WTB_hgEAuGeQ';
+    const topic = 'topic_1';
+    let options = {
```
TemirAskhat, line 29: Not sure about the `options` and the `request` but if it can be changed to `const`, will be better to have it immutable as possible

examples/produce_v3/python/streaming_produce_v3_main.py
TemirAskhat: Thank you! :)

examples/produce_v3/javascript/main.js
```diff
@@ -0,0 +1,52 @@
...
+        function (data) {
+            let chunk = data.toString('utf8');
+            if (chunk.trim()) {
+                console.log("Receipt for record #" + recordReceiptCounter + " is");
+                console.log(JSON.parse(chunk));
+                recordReceiptCounter += 1;
+            }
+        });
+
+    response.on('end', function () {
```
TemirAskhat, line 18: [nit]
Same as above with the "old-fashioned" way

TemirAskhat: I meant () => {}

trnguyencflt:
Unrelated failure in downstream schema-registry, will merge this.

PR #1215, opened 2023-11-16
DGS-9099 Enable Data Contract rules in Kafka REST Proxy

rayokota:
This small change to the startup class allows the jars for the Data Contract rules to be included in the classpath for the Kafka REST Proxy

The kafka-serde-tools subdir includes the client-side jars for the data quality rules and client-side field-level encryption.


msn-tldr:
@rayokota can you include why is this package needed in proxy? Is there a PR/feature that depends on `kafka-serde-tools`?

trnguyencflt (approved):

msn-tldr (approved):
lgtm

PR #1219, opened 2023-11-22
KREST-12630 Add missing dependency due to zookeeper upgrade

trnguyencflt:
Zookeeper dependency is updated in https://github.com/confluentinc/ce-kafka/pull/11994, which now requires extra dependency for running tests.


msn-tldr (approved):
lgtm

PR #1221, opened 2023-11-22
KREST-12630 Exclude logback from zookeeper dependency in tests

trnguyencflt:
Version 3.8.3 of zookeeper from [upstream](https://github.com/confluentinc/ce-kafka/pull/11994) makes tests hang, so use version 3.7.2 instead, this version also fixes the CVE-2023-44981

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,7 +110,17 @@
             <dependency>
                 <groupId>org.apache.zookeeper</groupId>
                 <artifactId>zookeeper</artifactId>
-                <version>3.7.2</version>
+                <version>${zookeeper.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>ch.qos.logback</groupId>
+                        <artifactId>logback-core</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>ch.qos.logback</groupId>
+                        <artifactId>logback-classic</artifactId>
+                    </exclusion>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -107,6 +107,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.apache.zookeeper</groupId>
+                <artifactId>zookeeper</artifactId>
+                <version>3.7.2</version>
```
ehumber, line 113: Is there a version of the zookeeper version number available from the common pom.xml?

trnguyencflt: unfortunately, the version from common is [3.8.3](https://github.com/confluentinc/common/blob/67e880a6a62c5a5f3bf5467c05ec408c77b28172/pom.xml#L99), which is causing the issue, I'll add a comment to this.

ehumber (approved):

trnguyencflt:
> How do we remember to eventually update the ZK level in the future? Is there a ZK fix coming?

@ehumber actually, this might raise a concern on any change to performance of clients, I've raised this kafka-eng channel.

trnguyencflt:
@ehumber after a bit of dependency and log digging, it is very likely that logback is the real culprit of making the tests hang, I've made a new commit to exclude logback, let's hope the build will pass now 🤞

PR #1230, opened 2023-12-06
KREST-12372 Upgrade dependency to JDK 1.8 with version 1.77 

trnguyencflt:
This PR upgrade dependencies to versions that fixe CVE-2023-33201

There will be pint merge following up after this.

### DIFF
--- a/kafka-rest/pom.xml
+++ b/kafka-rest/pom.xml
@@ -141,17 +141,17 @@
         </dependency>
         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcprov-jdk15on</artifactId>
+            <artifactId>bcprov-jdk18on</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcpkix-jdk15on</artifactId>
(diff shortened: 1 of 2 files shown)


kafka-rest/pom.xml
```diff
@@ -141,17 +141,17 @@
         </dependency>
         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcprov-jdk15on</artifactId>
```
trnguyencflt, line 144: jdk 1.5 is probably older than us, so we should use jdk 1.8 at least :)

ehumber: Do we state anywhere our minimum JRE version?  If we do and it's 8, then all good.

If we do and it's 5 (which would be a surprise) then we have a problem.

If we don't, then it's probably worth a quick readme update to add a line to say java 8 and later.

ehumber (approved):

kafka-rest/pom.xml
msn-tldr: https://docs.confluent.io/platform/current/installation/versions-interoperability.html#java

Java 8 is min, and its going to be deprecated soon.

msn-tldr (approved):

PR #1235, opened 2024-01-10
KREST-2300 Deflake KafkaConsumerManagerTest.testBackoffMsControlsPollCalls test

trnguyencflt:
This PR fixes the logic in `KafkaConsumerManager` class in the way it handles consumer backoff for readRecords

The approach that KafkaConsumerManager uses to execute backoff RunnableReadTask is:
- Each task has a fix backoff (in millis) and timeout (in millis)
- The [first run](https://github.com/confluentinc/kafka-rest/blob/6ad01a571a106f539cbfc8ed84ea900db3f367b0/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java#L336) is submitted in readRecords(...) function
- If the task is not done, then then task is [resubmitted into a DelayQueue](https://github.com/confluentinc/kafka-rest/blob/6ad01a571a106f539cbfc8ed84ea900db3f367b0/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java#L403) with [new calculated time to be executed](https://github.com/confluentinc/kafka-rest/blob/6ad01a571a106f539cbfc8ed84ea900db3f367b0/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java#L402)
- The DelayQueue is [polled](https://github.com/confluentinc/kafka-rest/blob/23130bbaa8b61e09c89559d68eb93723ba07067d/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java#L717) to get the task to be executed based on getDelay of the task.
- And so on until the timeout of the task has reach (timeout)

However, the implementation has flaws:
- First, it is in the [getDelay calculation](https://github.com/confluentinc/kafka-rest/blob/23130bbaa8b61e09c89559d68eb93723ba07067d/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java#L435), we want the task to be executed on `waitExpiration` instant, however Duration.between exclude the end instant, therefore, the task might be executed 1 millisecond earlier than expected
- Second, it is in the logic of calculating [the next time to be executed](https://github.com/confluentinc/kafka-rest/blob/23130bbaa8b61e09c89559d68eb93723ba07067d/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java#L402), basically, at the last retries, the value will always be set to requestExpiration. 

Combining the two flaws above, we can now explain why the test is flaky (to be fair, the test is correct, it is implementation's issue)
- On the last retry of the task, waitExpiration set to requestExpiration, due to the first flaw, the task would be executed 1 millisecond early.
- This task would do the polling, and other steps, as there is nothing to read, the task would finish those steps very quick (in less than 1 millisecond), and due to the second flaw, it would then be executed in the exact millisecond that the previous one ran. You can now notice the cycle will continue until that exact millisecond passes. 
- Hence, that is why we will get many retries at the last retry.

This PR fixes the above flaws by ensuring that:
- the task will be executed in the right time for the first flaw
- and it checks for the last retry condition for the second flaw

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
@@ -367,19 +367,16 @@ public class KafkaConsumerManager {
 
   class RunnableReadTask implements Runnable, Delayed {
     private final ReadTaskState taskState;
-    private final Instant started;
     private final Instant requestExpiration;
     private final Duration backoff;
     // Expiration if this task is waiting, considering both the expiration of the whole task and
     // a single backoff, if one is in progress
     private Instant waitExpiration;
-    private final Clock clock = Clock.systemUTC();
 
     public RunnableReadTask(ReadTaskState taskState, KafkaRestConfig config) {
       this.taskState = taskState;
-      this.started = clock.instant();
       this.requestExpiration =
-          this.started.plus(
+          clock.instant().plus(
               Duration.ofMillis(config.getInt(KafkaRestConfig.CONSUMER_REQUEST_TIMEOUT_MS_CONFIG)));
       this.backoff =
           Duration.ofMillis(config.getInt(KafkaRestConfig.CONSUMER_ITERATOR_BACKOFF_MS_CONFIG));
@@ -390,7 +387,10 @@ public class KafkaConsumerManager {
      * Delays for a minimum of {@code delayMs} or until the read request expires
      */
     public void delayFor(Duration delay) {
-      if (!requestExpiration.isAfter(clock.instant())) {
+      // the first condition means that the request has expired,
+      // the second one means that the request has been retried for
+      // the last time, so if it goes to this block again, it will be expired
+      if (!requestExpiration.isAfter(clock.instant()) || requestExpiration.equals(waitExpiration)) {
         // no need to delay if the request has expired
         taskState.task.finish();
         log.trace("Finished executing  consumer read task ({}) due to request expiry",
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
@@ -432,7 +432,9 @@ public class KafkaConsumerManager {
 
     @Override
     public long getDelay(TimeUnit unit) {
-      Duration delay = Duration.between(clock.instant(), waitExpiration);
+      // Note that, Duration.between excludes the end instant,
+      // so we need to add 1 millisecond so that the task will be run at the desire time
+      Duration delay = Duration.between(clock.instant(), waitExpiration.plusMillis(1));
       return unit.convert(delay.toMillis(), TimeUnit.MILLISECONDS);
     }
 
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/v2/MockConsumer.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/v2/MockConsumer.java
@@ -52,7 +52,7 @@ public class MockConsumer<K, V> extends org.apache.kafka.clients.consumer.MockCo
         SortedSet<OffsetAndTimestamp> offsets =
             offsetForTimes.computeIfAbsent(
                 new TopicPartition(topic, partition),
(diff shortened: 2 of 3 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/v2/MockConsumer.java
```diff
@@ -62,7 +62,7 @@ public synchronized Map<TopicPartition, OffsetAndTimestamp> offsetsForTimes(
         Map<TopicPartition, OffsetAndTimestamp> result = new HashMap<>();
         for (Entry<TopicPartition, Long> entry : timestampsToSearch.entrySet()) {
             SortedSet<OffsetAndTimestamp> offsets =
-                offsetForTimes.getOrDefault(entry.getKey(), new TreeSet<>());
```
trnguyencflt, line 65: This is the main change, `new TreeSet<>()` doesn't have a comparator so the order can be messed up here.

ehumber (approved):

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -367,19 +367,16 @@ protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
 
   class RunnableReadTask implements Runnable, Delayed {
     private final ReadTaskState taskState;
-    private final Instant started;
     private final Instant requestExpiration;
     private final Duration backoff;
     // Expiration if this task is waiting, considering both the expiration of the whole task and
     // a single backoff, if one is in progress
     private Instant waitExpiration;
-    private final Clock clock = Clock.systemUTC();
```
trnguyencflt, line 376: we already have a static clock on top-level class, there is no need to have another clock here.

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -390,7 +387,10 @@ public RunnableReadTask(ReadTaskState taskState, KafkaRestConfig config) {
      * Delays for a minimum of {@code delayMs} or until the read request expires
      */
     public void delayFor(Duration delay) {
-      if (!requestExpiration.isAfter(clock.instant())) {
+      // the first condition means that the request has expired,
+      // the second one means that the request has been retried for
+      // the last time, so if it goes to this block again, it will be expired
+      if (!requestExpiration.isAfter(clock.instant()) || requestExpiration.equals(waitExpiration)) {
```
trnguyencflt, line 393: this is the main fix for the flakiness

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
```diff
@@ -390,7 +387,12 @@ public RunnableReadTask(ReadTaskState taskState, KafkaRestConfig config) {
      * Delays for a minimum of {@code delayMs} or until the read request expires
      */
     public void delayFor(Duration delay) {
-      if (!requestExpiration.isAfter(clock.instant())) {
+      // the first condition means that the request has expired,
+      // the second one means that the request has been retried for
+      // the last time, so if it goes to this block again, it will be expired
+      if (!requestExpiration.isAfter(clock.instant())
+      || requestExpiration.equals(waitExpiration)
```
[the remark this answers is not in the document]
ehumber: Is that how a request is marked to say all the retries are used up, by setting the expiration time to be Instant.EPOCH?

kafka-rest/src/main/java/io/confluent/kafkarest/v2/KafkaConsumerManager.java
ehumber: Why does this fix the problem?

trnguyencflt: @ehumber I've provided the detailed explanation in the description 👍 , hope it is clear enough.

ehumber (approved):

trnguyencflt:
Jenkins failed due to unrelated issue, proceed to merge.

PR #1236, opened 2024-01-10
KREST-NNN disable flakey test

ehumber:
See https://confluent.slack.com/archives/C06CFQNK2BF/p1704845102888579 for details.  This test continues to flake, and is disabled in master.  Matching that behaviour back in the 6.2.x branch


kafka-rest/src/test/java/io/confluent/kafkarest/v2/KafkaConsumerManagerTest.java
```diff
@@ -70,6 +70,7 @@
 import io.confluent.kafkarest.entities.EmbeddedFormat;
 import io.confluent.kafkarest.entities.TopicPartitionOffset;
 import io.confluent.rest.RestConfigException;
+import org.junit.jupiter.api.Disabled;
```
trnguyencflt, line 73: could you remove this import as this branch is with Junit4 so it would be import for Ignore

trnguyencflt (approved):

PR #1241, opened 2024-01-25
KNET-13992 Fix java.lang.ClassNotFoundException: org.xerial.snappy.SnappyInputStream

trnguyencflt:
As title said. Add snappy-java test dependency to fix the error of ClassNotFoundException


msn-tldr:
confused, why is snappy dependency required now in pom?

msn-tldr:
Got it, so missing transitive dep broke us now.

msn-tldr (approved):
lgtm

PR #1249, opened 2024-02-22
Add doLog parameter for KafkaRestConfig constructor

trnguyencflt:
Ref: https://github.com/confluentinc/ce-kafka/pull/13010

For context, we don't want to log configuration values by default for transient clients in CCloud. This PRs introduces new constructors to `KafkaRestConfig` so that we can have the control over the logging behaviour. Note that, we don't use the `CONFIGS_LOGGING_ENABLED_PROP` config because it is internal, this config will be use in upstream project of kafka-rest.

Related PRs:
- [rest-utils](https://github.com/confluentinc/rest-utils/pull/464)


msn-tldr (approved):
lgtm

PR #1250, opened 2024-02-23
Allow to set KafkaAvroSerializerConfig in KafkaRestConfig (#1229)

apurushotama:
Cherry-picking to ensure avro serializer is more configurable. 
This is ` avro.use.logical.type.converters` config so that we are able to serialize logical type in avro message.


ehumber (approved):

msn-tldr (approved):
lgtm

In this case one can work-out you are cherry-picking an old commit. But generally good to add a description, even if its quick like "cherry-picking <PR> to fix <fill_issue>"

PR #1261, opened 2024-03-25
Upscale Semaphore CI instance to avoid occasionally crash

trnguyencflt:
Update semaphore instance to a bigger machine type to avoid occasionally crashes in nightly builds.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -7,7 +7,7 @@ version: v1.0
 name: build-test-release
 agent:
   machine:
-    type: s1-prod-ubuntu20-04-amd64-1
+    type: s1-prod-ubuntu20-04-amd64-2
 
 fail_fast:
   cancel:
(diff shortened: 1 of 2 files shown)


ehumber (approved):

.semaphore/semaphore.yml
```diff
@@ -7,7 +7,7 @@ version: v1.0
 name: build-test-release
 agent:
   machine:
-    type: s1-prod-ubuntu20-04-amd64-1
+    type: s1-prod-ubuntu20-04-amd64-2
```
msn-tldr, line 10: curious, is this the recommended way to scaling up in semaphore?

msn-tldr: it seems so
https://confluentinc.atlassian.net/wiki/spaces/TOOLS/pages/2820542346/Semaphore+Self+Hosted+Agents

msn-tldr (approved):

.semaphore/semaphore.yml
trnguyencflt: yeah, that is the link 👍

PR #1264, opened 2024-04-15
KNET-14986: Uncaught exception while closing input stream is causing …

akhileshchg:
…clean up failures.

InpuStream close just handles the IOException. But when the there are unused streams Jersey throws BadRequestException instead of IOException. This uncaught exception disrupts the outbound response closure leaving the producer thread behind.

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
@@ -255,7 +255,11 @@ public abstract class StreamingResponse<T> {
       try {
         inputStream.close();
       } catch (IOException e) {
-        log.error("Error when closing the request stream.", e);
+        log.error("Error when closing the request stream", e);
+      } catch (BadRequestException e) {
+        log.error("Error when closing the request stream", e.getCause() != null ? e.getCause() : e);
+      } catch (Throwable e) {
+        log.error("Unknown error when closing the request stream.", e);
       }
     }
 


AndrewJSchofield (approved):

trnguyencflt:
@akhileshchg are you able to resolve the issue with this change?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -255,7 +255,11 @@ public void close() {
       try {
         inputStream.close();
       } catch (IOException e) {
-        log.error("Error when closing the request stream.", e);
+        log.error("Error when closing the request stream", e);
+      } catch (BadRequestException e) {
```
msn-tldr, line 259: @akhileshchg thanks for fixing into this.

Can you add a comment here to explain why handling these exception are important w.r.t thread leaks?

ehumber (approved):

ehumber:
Nice work, this code is so twisty, not a nice introductory task at all!
 

As Mayank says, could you add some details on how this fixes the problem please.  

It looks like 

 - https://github.com/confluentinc/kafka-rest/blob/1683233421407a61353db7c869c8420441530012/kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java#L200 is throwing on close of the InputStream, so the close of the outbound stream, never happens, and the @Injected ProduceAction.produce() never gets chance to complete.  And so the Context level closing, that would tidy up the KafkaProducer never occurs.  Is that right?

- The exception caught by InputStream.close() isn't an IOException because it's coming from the get of the delegate, at https://github.com/confluentinc/kafka-rest/blob/1683233421407a61353db7c869c8420441530012/kafka-rest/src/main/java/io/confluent/kafkarest/response/JsonStream.java#L68 rather than the InputStream.close (which does only throw an IOException). 

- The problem is only seen on a totally unused Stream, because the delegate.get() call that fails, because the delegate hasn't been set up yet for the unused stream, can successfully return once the stream has been used.

Is that right?

msn-tldr:
@ehumber & @akhileshchg thanks for writing down the detailed break down of how thread leak happens.

PR #1265, opened 2024-04-17
CONFLUENT: Remove uses of KafkaConfig Prop constants

splett2:
The KRest build is failing due to various changes to KafkaConfig prop constant locations in AK. Currently the replication factor and interbrokerListenerName props have been moved, but since it seems plausible that a bunch of the props defined in `KafkaConfig` may change class locations, I went ahead and replaced references to KafkaConfig with fixed string constants.


apurushotama (approved):

apurushotama (approved):

trnguyencflt:
@splett2 thanks for fixing this 👍

PR #1272, opened 2024-06-05
KNET-15501 work around jackson databind string length limit

ehumber:
Implement the change from https://github.com/FasterXML/jackson-core/issues/1001#issuecomment-1527216221

I'm using maxint for the value, as we do length checks of the stream elsewhere, and this seems safest to avoid a regression (see the code comment for more details)


trnguyencflt (approved):

msn-tldr (approved):

ehumber:
The build failure is due to https://confluentinc.atlassian.net/browse/KNET-15458 so we need to wait for that to be resolved before this will run through.

ehumber:
Pint merged up from 7.7 instead, closing

PR #1300, opened 2024-09-25
[KNET-12362] Produce V3 API - Improve visibility around produce-record errors

wcheng92:
### Description
A single http requests to produce V3 API can contain many produce-records. A record receipt is returned for each record, each with its own error-code. For example see the produce example to produce a batch of records - [Kafka REST APIs for Confluent Cloud | Confluent Documentation](https://docs.confluent.io/cloud/current/kafka-rest/kafka-rest-cc.html#producing-a-batch-of-records-to-a-topic)

> {
  "error_code": 401,
  "cluster_id": lkc-vo0pz,
  "topic_name": "testTopic1",
  "partition_id": 0,
  "offset": 67217,
  "timestamp": "2021-12-13T16:29:10.951Z",
  "key": {
    "type": "BINARY",
    "size": 10
  },
  "value": {
    "type": "JSON",
    "size": 26
  }
}
{
  "error_code": 503,
  "cluster_id": lkc-vo0pz,
  "topic_name": "testTopic1",
  "partition_id": 0,
  "offset": 67218,
  "timestamp": "2021-12-13T16:29:10.951Z",
  "key": {
    "type": "BINARY",
    "size": 10
  },
  "value": {
    "type": "JSON",
    "size": 26
  }
}

Using the above example, REST log would like below with status 200(OK) - 

`10.1.1.158 - - [14/Nov/2023:19:20:40 +0000] "POST /kafka/v3/clusters/lkc-m8j07q/topics/apac.consumer.AddCalculatedAttributes.v1/records HTTP/1.1" 200 196 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36 Hutool" 503 - `

Whereas clearly the user’s request had errors(sometimes 5xx). There is no visibility about these record-level error codes, which hinders oncall ability to answer customer’s question like “did the produce API request failed/succeeded for all records, or subset of records?”. This should be improved

### Solution
Add a summarised counts of different error-codes in REST request log, `Codes=401:1,503:1` as in example section below.

### Example
Before - `10.1.1.158 - - [14/Nov/2023:19:20:40 +0000] "POST /kafka/v3/clusters/lkc-m8j07q/topics/apac.consumer.AddCalculatedAttributes.v1/records HTTP/1.1" 200 196 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36 Hutool" 503 -`

Now - `10.1.1.158 - - [14/Nov/2023:19:20:40 +0000] "POST /kafka/v3/clusters/lkc-m8j07q/topics/apac.consumer.AddCalculatedAttributes.v1/records HTTP/1.1" 200 196 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36 Hutool" 503 - Codes=401:1,503:1`


kafka-rest/src/main/java/io/confluent/kafkarest/requestlog/CustomLog.java
```diff
@@ -63,6 +69,44 @@ protected void doStop() throws Exception {
...
+      }
+      ProduceRecordErrorCounter that = (ProduceRecordErrorCounter) other;
+      return Objects.equals(produceErrorCodeCountMap, that.produceErrorCodeCountMap);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(produceErrorCodeCountMap);
+    }
+  }
```
dimitarndimitrov, line 108: While the issue might be benign, this construct contains races by the classical definition of a data race - two simultaneous accesses from different threads, where at least one access is a write, and there are no synchronization mechanisms between the accesses.
The writes are synchronized, so they are mutually exclusive with themselves, but there is no mechanism to guarantee either atomicity or visibility of the write actions to the read actions.

Furthermore, while defining `equals` and `hashCode` is generally good, here it might cause the wrong impression that this is usable as a map key - it's not really, as this is an inherently mutable type. In addition, the `Objects.hash` used in the `hashCode` implementation is a bit of a code smell - for a single element, it's actually just an obfuscated `Objects.hashCode()` (it shifts the instance hash code by 31).

dimitarndimitrov (commented):
It's great that Kafka REST continues to receive attention and improvements, especially for its newest significant API, the Produce v3!

I have one comment about the error counting structure.

kafka-rest/src/main/java/io/confluent/kafkarest/requestlog/CustomLog.java
msn-tldr: > there are no synchronization mechanisms between the accesses.

The data race between write and read of the map is unlikely. But such races are hard to track whenever they occur, so gr8 u called it out!

@wcheng92 can you make the `toString()` method synchronized?

> here it might cause the wrong impression that this is usable as a map key

A good call out again. Right the equals and hashCode are misleading.

@wcheng92 you should remove 2 methods equals() and hashCode(). But u would still need a way for easymock to compare 2 objects  `ProduceRecordErrorCounter` are logically equal(depending upon the maps contained within) Vs object equality. For this you can create a [custom-matcher](ProduceRecordErrorCounter) in easymock, which would do equality similar to equals() method.

wcheng92: Thanks for helping to review the PR! 

Here are my thoughts about addressing the potential issues you mentioned:

1. Race condition issue: Synchronize `toString`, `equals`, and `hashCode` methods to prevent data race
2. Wrong impression for `equals` and `hashCode`: Make clearer comments for the class and mention this class **should NOT** be used as a map key
3. hashCode code smell: Remove `hashCode`

Please let me know if these potential solutions work or if there's a better solution ! Thank you so much!

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -218,11 +225,15 @@ private void closeAll(AsyncResponseQueue responseQueue) {
     responseQueue.close();
   }
 
-  private ResultOrError handleNext(T result, @Nullable Throwable error) {
+  private ResultOrError handleNext(
+      T result, @Nullable Throwable error, ProduceRecordErrorCounter produceRecordErrorCounter) {
     if (error == null) {
+      produceRecordErrorCounter.incrementErrorCount(HTTP_SUCCESS_CODE);
```
ehumber, line 231: Do we want to spend the time logging success error codes?

kafka-rest/src/main/java/io/confluent/kafkarest/requestlog/CustomLog.java
ehumber: If we were to run a tool like spotbugs here, it would say that all variables that sit within a synchronized block, should sit within a synchronized block throughout the class, so it's the right thing to lock the toString here too.  

In reality, I actually don't think we care too much if the toString is slightly lagging in being correct though.  And I think it's only called when the overall produce request completes, ie no more writes are coming in, so the risk seems minimal.  And it's shouldn't concurrent modification exception either, as it's a read.  But, best to be correct and your code change looks good to me.

I see you've removed the hash and equals, thank you.

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -73,6 +74,7 @@ public abstract class StreamingResponse<T> {
 
   private static final Logger log = LoggerFactory.getLogger(StreamingResponse.class);
   private static final int ONE_SECOND_MS = 1000;
+  private static final int HTTP_SUCCESS_CODE = 200;
```
trnguyencflt, line 77: we already have constant for this from either Jetty or Jersey libraries, there is no need to redeclare

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
```diff
@@ -218,11 +225,15 @@ private void closeAll(AsyncResponseQueue responseQueue) {
...
 
-  private ResultOrError handleNext(T result, @Nullable Throwable error) {
+  private ResultOrError handleNext(
+      T result, @Nullable Throwable error, ProduceRecordErrorCounter produceRecordErrorCounter) {
     if (error == null) {
+      produceRecordErrorCounter.incrementErrorCount(HTTP_SUCCESS_CODE);
       return ResultOrError.result(result);
     } else {
       log.debug("Error processing streaming operation.", error);
+      int errorCode = EXCEPTION_MAPPER.toErrorResponse(error.getCause()).getErrorCode();
```
trnguyencflt, line 235: do we already check if `error.getCause()` is not null or the whole `EXCEPTION_MAPPER.toErrorResponse(error.getCause()).getErrorCode()` could throw exception?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/CustomLogIntegrationTest.java
```diff
@@ -193,6 +195,20 @@ public void setUp(TestInfo testInfo) throws Exception {
       restConfigs.put(KafkaRestConfig.PRODUCE_MAX_REQUESTS_PER_SECOND, 99999);
       restConfigs.put(KafkaRestConfig.PRODUCE_MAX_BYTES_GLOBAL_PER_SECOND, 99999);
       restConfigs.put(KafkaRestConfig.PRODUCE_MAX_BYTES_PER_SECOND, 1);
+    } else if (testInfo
+        .getDisplayName()
+        .contains(
+            "test_WhenProduceRequestWithMultipleRecords_ThenRequestLogHasAggregatedErrorCodes")) {
+      restConfigs.put("dos.filter.enabled", "true");
```
trnguyencflt, line 202: the constant for the configuration can be found in RestConfig

kafka-rest/src/test/java/io/confluent/kafkarest/integration/CustomLogIntegrationTest.java
```diff
@@ -398,9 +422,58 @@ public void test_whenPerTenantProduceBytesRateLimitTriggered_ThenRequestLogHasRe
...
+      // and 5 records having error code 429
+      verifyLog(
+          totalRequests,
+          ErrorCodes.PRODUCE_MAX_REQUESTS_PER_TENANT_LIMIT_EXCEEDED,
+          totalRequests - requestsWithStatusOk,
+          true,
+          false,
+          true);
+    } catch (Exception e) {
+      e.printStackTrace();
```
trnguyencflt, line 473: could you please use the logger instead?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
wcheng92: The benefit of keeping success error code is that we can know the total number of produce record too. Should I still remove this line?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
wcheng92: Thanks for the comment! I wonder how can I access to the constants within Jetty/Jersey? 
image

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
wcheng92:
```suggestion
      if (error.getCause() == null) {
        throw new IllegalArgumentException("Error cause is null", error);
      }
      int errorCode = EXCEPTION_MAPPER.toErrorResponse(error.getCause()).getErrorCode();
      produceRecordErrorCounter.incrementErrorCount(errorCode);
```

Thanks for calling this! Is this change above good?

kafka-rest/src/main/java/io/confluent/kafkarest/response/StreamingResponse.java
ehumber: yes please - I saw you did this in the airlock PR already, thank you

ehumber:
One last question from me - did you test this in our cloud environment @wcheng92 ?

kafka-rest/src/test/java/io/confluent/kafkarest/integration/CustomLogIntegrationTest.java
wcheng92: I didn't see a file named `RestConfig `. Is it `KafkaRestConfig`?

ehumber (approved):

msn-tldr (approved):
lgtm

@wcheng92 
> One last question from me - did you test this in our cloud environment @wcheng92 ?

Can you share the findings of ur ccloud testing? Merge once the Ccloud testing looks good.

PR #1329, opened 2024-12-24
Remove Zk integration tests and fix consumer group describe result

akhileshchg:


kafka-rest/src/test/java/io/confluent/kafkarest/integration/AvroProducerTest.java
```diff
@@ -199,7 +199,7 @@ protected <K, V> void testProduceToPartition(
   }
 
   @ParameterizedTest(name = TEST_WITH_PARAMETERIZED_QUORUM_NAME)
-  @ValueSource(strings = {"kraft", "zk"})
+  @ValueSource(strings = {"kraft"})
```
soondenana, line 202: Kafka need to switch default to "zk" so that we don't have to add this line. Can you check if they have already done it, its annoying to have this line before each tests.

soondenana (approved):
LGTM

PR #1330, opened 2024-12-24
Fix unit test failures after changing ConsumerGroupState to GroupState

akhileshchg:


kafka-rest/src/main/java/io/confluent/kafkarest/entities/ConsumerGroup.java
```diff
@@ -102,32 +105,51 @@ public abstract static class Builder {
...
-    public ConsumerGroupState toConsumerGroupState() {
+    public GroupState toConsumerGroupState() {
       try {
-        return ConsumerGroupState.valueOf(name());
+        return GroupState.parse(name);
       } catch (IllegalArgumentException e) {
-        return ConsumerGroupState.UNKNOWN;
+        return GroupState.UNKNOWN;
       }
     }
```
soondenana, line 148: How difficult is it to create a static map for this, instead of dynamically converting back and forth.

soondenana (approved):
LGTM

PR #1333, opened 2025-01-02
Change visibility of testInfo and quorumTestHarness to protected so they can be used by subclasses

trnguyencflt:
Change visibility of testInfo and quorumTestHarness to protected so they can be used by subclasses.

This will help with customising subclasses for ce-kafka-rest to fix issue regarding deprecating zk.

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
@@ -119,8 +119,8 @@ public abstract class ClusterTestHarness {
   private final boolean manageRest;
 
   // Quorum controller
-  private TestInfo testInfo;
-  private QuorumTestHarness quorumTestHarness;
+  protected TestInfo testInfo;
+  protected QuorumTestHarness quorumTestHarness;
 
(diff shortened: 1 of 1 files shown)


shubh-ranade (commented):
LGTM, being able to override these in subclasses would be helpful! I had a qq - does it matter if we expose getters setters instead?

trnguyencflt:
> LGTM, being able to override these in subclasses would be helpful! I had a qq - does it matter if we expose getters/setters instead?

we don't need to override these, I made them protected so that we can use them directly in subclasses.

if we expose getters/setters, that is more about for other classes/packages to be used, which we don't really need.

ijuma:
Exposing setters/getters is not helpful, but exposing only accessors (if that is all subclasses need) is a better way of doing things. Mutable fields across classes are to be avoided as much as possible.

trnguyencflt:
> Exposing setters/getters is not that helpful, but exposing only accessors (if that is all subclasses need) is a better way of doing things. Mutable fields across classes are to be avoided as much as possible.

@ijuma @shubh-ranade updated the code to expose accessors to subclasses

shubh-ranade (approved):
lgtm, thanks!

PR #1340, opened 2025-01-08
Remove kafka_2.13 dependency, only for testing and exclude log4j2 compatibility jar

trnguyencflt:
kafka-rest doesn't use kafka_2.13 other than testing so we remove the dep, also excluding log4j2 compatibility jars to make sure that we don't clash on logging library


ijuma (approved):
LGTM, thanks.

PR #1341, opened 2025-01-10
Copy createBrokerConfig from AK to kafka rest to avoid depending on u…

trnguyencflt:
…pstream

### DIFF
--- a/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
+++ b/kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
@@ -34,11 +34,13 @@ import io.confluent.kafkarest.KafkaRestApplication;
 import io.confluent.kafkarest.KafkaRestConfig;
 import io.confluent.kafkarest.common.CompletableFutures;
 import io.confluent.rest.RestConfig;
+import java.io.File;
 import java.io.IOException;
 import java.net.ServerSocket;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
```diff
@@ -424,6 +437,160 @@ protected Properties getBrokerProperties(int i) {
...
 
+  private static boolean shouldEnable(
+      Option<SecurityProtocol> interBrokerSecurityProtocol, SecurityProtocol protocol) {
+    if (interBrokerSecurityProtocol.isDefined()) {
+      return interBrokerSecurityProtocol.get() == protocol;
+    }
+    return false;
+  }
+
+  public static Properties createBrokerConfig(
```
msn-tldr, line 448: @trnguyencflt is this taken from AK? If true can you add a comment linking to the source. That way its easy to track the source of there are any divergences in the future, and need to update.

msn-tldr (approved):
lgtm, 1 comment above.

kafka-rest/src/test/java/io/confluent/kafkarest/integration/ClusterTestHarness.java
trnguyencflt: added 👍

shubh-ranade (approved):
Thanks for the fix!

shubh-ranade (approved):

PR #1348, opened 2025-02-04
Use GroupState from AK instead of deprecated api

trnguyencflt:


shubh-ranade (approved):
Thanks for the PR! 

@trnguyencflt I didn't quite understand why changing the enum to a wrapper class is better for preserving REST API response compatibility. If we had kept the enum while using `GroupState` , how is that worse than having a wrapper class?

trnguyencflt:
> If we had kept the enum while using `GroupState` , how is that worse than having a wrapper class?

If we keep the enum while using `GroupState`, we still have the issue of not able to parse the GroupState, described here https://confluent.slack.com/archives/C085WE8JZJB/p1735074529676849, that will make us depend on the change from AK to propagate.

The wrapper class ensures that we keep up with the change from GroupState.

Using `GroupState` enum directly would be the best option, however, it is not possible because it overrides toString function (used by jackson serde) to return CamelCase state name instead of SCREAMING_SNAKE_CASE, that's why I use the wrapper class.

trnguyencflt:
> > If we keep the enum while using GroupState, we still have the issue of not able to parse the GroupState, described here https://confluent.slack.com/archives/C085WE8JZJB/p1735074529676849, that will make us depend on the change from AK to propagate.
> 
> @trnguyencflt I'm curious - would we have the same issue if we used the following?
> 
> ```
> public static State fromGroupState(GroupState state) {
>       try {
>         return State.valueOf(state.toString().toUpperCase(Locale.ROOT));
>       } catch (IllegalArgumentException e) {
>         return UNKNOWN;
>       }
> }
> ```

I think that would still return unknown because

`State.valueOf()` is expecting SCREAMING_SNAKE_CASE while `state.toString().toUpperCase(Locale.ROOT)` would return UPPERCASE.
this block might work

```
public static State fromGroupState(GroupState state) {
      try {
        return State.valueOf(state.name().toUpperCase(Locale.ROOT));
      } catch (IllegalArgumentException e) {
        return UNKNOWN;
      }
}
```

however, again, it goes back to an issue that we are creating our own enum State that is sensitive to change from upstream GroupState.

shubh-ranade:
> it goes back to an issue that we are creating our own enum State that is sensitive to change from upstream GroupState.

@trnguyencflt ahh I think I understand now. The problem is that having an enum `State` means we assume that `GroupState` has certain fixed values, for example `UNKNOWN`, `PREPARING_REBALANCE`, `COMPLETING_REBALANCE` if upstream `GroupState` changes, `State` enum has to be updated manually, for example when a new state is added or string casing is changed.

But by having a static class we can dynamically pull in the latest state values from upstream `GroupState` when computing `State.NAME_TO_ENUM`.

@trnguyencflt thanks for the explanation!

PR #1349, opened 2025-02-06
[KNET-17449] Implement hashCode for ConsumerGroup.State to satisfy equals contract

trnguyencflt:
This implement `hashCode` for ConsumerGroup.State since we need to override both equals and hashCode to ensure consistency of a class.


shubh-ranade (approved):
Looks good to me, thanks for the PR!

PR #1353, opened 2025-03-12
KCOORD-966: Add protocol related fields to list consumer groups api

dongnuo123:
This patch adds two protocol related fields to `ListConsumerGroups` api.

- Group protocol (CLASSIC, CONSUMER, SHARE, UNKNOWN)

- New field to indicate whether a group with the consumer protocol does/does not have all its assigned consumers on the new protocol. The boolean is set to be false for groups using protocols other than CONSUMER.

Beside, it also updates the `state` field to have the new states of the new protocol ASSIGNING and RECONCILING.

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -2172,6 +2172,8 @@ components:
             - is_simple
             - partition_assignor
             - state
+            - type
+            - is_mixed_consumer_group
             - coordinator
             - consumers
             - lag_summary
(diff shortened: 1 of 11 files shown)


dimitarndimitrov (commented):
@dongnuo123 Can you please update the OpenAPI spec as well? We might just need to update `ConsumerGroupData` with the new fields [here](https://github.com/confluentinc/kafka-rest/blob/master/api/v3/openapi.yaml#L2165-L2194), but do double-check me on that.

silvanli-confluent (approved):

PR #1355, opened 2025-03-17
perf: add singleton to Schema Record Serializers

devhritwik:
Closing this PR in lieu of https://github.com/confluentinc/kafka-rest/pull/1356

Update -
The integ tests are failing on this PR, for reasons I couldn't find. Here's an alternative PR - https://github.com/confluentinc/kafka-rest/pull/1356

JIRA: KNET-16476

For each produce request, serializers for JSON, Avro, Protobuf were being initialized as they were always being constructed (with `new`).  I have changed them to `@Singleton` so that their instances are reused with future invocations. This has led to a stark reduction in CPU time taken. 
The controller Module will now have a singleton implementation of the SchemaRecordSerializer. 

Without Singleton - 
Screenshot 2025-03-14 at 17 47 35
Without Singleton Profiler - https://drive.google.com/file/d/1wfZ7t5-EeBeenUVUzHy_FBVJ5jA_VJWH/view?usp=sharing
With Singleton - 
Screenshot 2025-03-14 at 17 47 41
With Singleton Profiler - https://drive.google.com/file/d/1PM9pyF986P_hd3vZ3i73ShoJFcuXDNuT/view?usp=sharing

### DIFF
--- a/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
+++ b/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
@@ -41,7 +41,6 @@ import io.confluent.kafkarest.exceptions.BadRequestException;
 import java.io.IOException;
 import java.util.Map;
 import java.util.Optional;
-import javax.inject.Singleton;
 import org.apache.avro.AvroTypeException;
 import org.everit.json.schema.ValidationException;
 
@@ -57,11 +56,10 @@ final class SchemaRecordSerializerImpl implements SchemaRecordSerializer {
       @JsonschemaSerializerConfigs Map<String, Object> jsonschemaSerializerConfigs,
       @ProtobufSerializerConfigs Map<String, Object> protobufSerializerConfigs) {
     requireNonNull(schemaRegistryClient);
-    avroSerializer = AvroSerializer.getInstance(schemaRegistryClient, avroSerializerConfigs);
+    avroSerializer = new AvroSerializer(schemaRegistryClient, avroSerializerConfigs);
     jsonschemaSerializer =
-        JsonSchemaSerializer.getInstance(schemaRegistryClient, jsonschemaSerializerConfigs);
-    protobufSerializer =
-        ProtobufSerializer.getInstance(schemaRegistryClient, protobufSerializerConfigs);
+        new JsonSchemaSerializer(schemaRegistryClient, jsonschemaSerializerConfigs);
+    protobufSerializer = new ProtobufSerializer(schemaRegistryClient, protobufSerializerConfigs);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
```diff
@@ -128,40 +130,73 @@ record = (Message) ProtobufSchemaUtils.toObject(data, protobufSchema);
...
     private ProtobufSerializer(
         SchemaRegistryClient schemaRegistryClient, Map<String, Object> configs) {
       this.schemaRegistry = requireNonNull(schemaRegistryClient);
       configure(serializerConfig(configs));
     }
 
+    private static ProtobufSerializer getInstance(
+        SchemaRegistryClient schemaRegistryClient, Map<String, Object> configs) {
+      if (instance == null) {
+        instance = new ProtobufSerializer(schemaRegistryClient, configs);
```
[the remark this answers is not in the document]
devhritwik: This is a good suggestion and I will apply it in this PR

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
```diff
@@ -128,40 +130,79 @@ record = (Message) ProtobufSchemaUtils.toObject(data, protobufSchema);
...
   private static final class AvroSerializer extends AbstractKafkaAvroSerializer {
 
+    private static volatile AvroSerializer instance;
+
     private AvroSerializer(SchemaRegistryClient schemaRegistryClient, Map<String, Object> configs) {
       this.schemaRegistry = requireNonNull(schemaRegistryClient);
       configure(serializerConfig(configs));
     }
 
+    private static AvroSerializer getInstance(
```
trnguyencflt, line 143: please use double-checked locking method for improve performance, example can be found here https://github.com/confluentinc/ce-kafka/blob/d16e658ce1b1ef93cfcb36e6071c0512297c3b0a/ce-kafka-client-plugins/src/main/java/io/confluent/kafka/security/fips/provider/BcFipsProviderCreator.java#L36

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
trnguyencflt, line 1: please add tests to verify that provide() return a singleton object

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
devhritwik: Oh that's a great tip! Thanks. 
I googled and found out that `synchronized` is an expensive call (because of locking) and checking for null beforehand will help us return an already instantiated object.

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerImpl.java
devhritwik: Yes I should do that, thanks!

devhritwik: I am having difficulty setting this up. If I call the `provide` function, each time it will definitely call `new` and hence not worth testing. What I want is to directly inject it here like this https://github.com/confluentinc/kafka-rest/blob/5e12d55eb3f203f159ce038c77ddf740582c4380/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/RecordSerializerFacade.java#L31 and ensure the same object gets injected everytime. I am not able to do that really. Do you have an idea on how to test this?

Alternatively, I can implement it the way we have done for AvroSerializer, using a local instance object.

devhritwik: To test this, I can implement it the same way we have implemented Singleton for AvroSerializer, using an instance and then just returning it if it is already set.

PR #1356, opened 2025-03-17
perf: add singleton to SchemaRecordSerializerFactory

devhritwik:
JIRA: KNET-16476

Alternative cleaner implementation of this PR - https://github.com/confluentinc/kafka-rest/pull/1355

For each produce request, serializers for JSON, Avro, Protobuf were being initialized as they were always being constructed (with `new`).  I have changed them to `@Singleton` so that their instances are reused with future invocations. This has led to a stark reduction in CPU time taken. (roughly 90 percent)
The controller Module will now have a singleton implementation of the SchemaRecordSerializer. 

Without Singleton - 
Screenshot 2025-03-14 at 17 47 35
Without Singleton Profiler - https://drive.google.com/file/d/1wfZ7t5-EeBeenUVUzHy_FBVJ5jA_VJWH/view?usp=sharing
With Singleton - 
Screenshot 2025-03-14 at 17 47 41
With Singleton Profiler - https://drive.google.com/file/d/1PM9pyF986P_hd3vZ3i73ShoJFcuXDNuT/view?usp=sharing


kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
```diff
@@ -76,6 +78,7 @@ private SchemaRecordSerializerFactory(
     }
 
     @Override
+    @Singleton
```
trnguyencflt, line 81: what does this annotation mean in the context?

devhritwik: This is on the provide method of `SchemaRecordSerializer`, hence in this context it means once the provide has been initialized, any future injections of `SchemaRecordSerializer` will provide the Singleton first initialized object of this class.

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/ControllersModuleTest.java
```diff
@@ -0,0 +1,68 @@
...
+    ServiceLocatorUtilities.bind(
+        serviceLocator,
+        new AbstractBinder() {
+          @Override
+          protected void configure() {
+            Map<String, Object> dummyConfig = new HashMap<>();
+            dummyConfig.put("key", "value");
+
+            bind(dummyConfig)
+                .qualifiedBy(new ConfigModule.AvroSerializerConfigsImpl())
```
trnguyencflt, line 46: why can't we use the factory binding in the test?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
devhritwik: Yes in fact, I was unsure as well but when I remove the annotation we see test failures in `RecordSerializer.serialize`. It also helps in clarifying our intent here for future use of the class and hence we should keep this annotation here.

trnguyencflt: yeah, let's keep the annotation here

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/ControllersModuleTest.java
trnguyencflt: if that is the case, would this test be moved to the same package as AvroSerializerConfigsImpl? (we can rename the test to something like SingletonSerializerTest

devhritwik: that makes sense, in that case at least we can keep the `AvroSerializerConfigsImpl` as protected instead of public as of now. I will do that then

devhritwik: We will have to make - https://github.com/confluentinc/kafka-rest/blob/5e12d55eb3f203f159ce038c77ddf740582c4380/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializer.java#L24 public. I think making this interface public is much better than making each of the `ConfigImpl` public. What do you think?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
```diff
@@ -52,7 +52,9 @@ protected void configure() {
     bindFactory(SchemaManagerFactory.class).to(SchemaManager.class);
     bind(TopicConfigManagerImpl.class).to(TopicConfigManager.class);
     bind(TopicManagerImpl.class).to(TopicManager.class);
-    bindFactory(SchemaRecordSerializerFactory.class).to(SchemaRecordSerializer.class);
+    bindFactory(SchemaRecordSerializerFactory.class)
+        .to(SchemaRecordSerializer.class)
+        .in(Singleton.class);
```
trnguyencflt, line 57: there is a concern on this binding, suppose for some reason that schemaRegistryClient is empty, this would mean this  SchemaRecordSerializerFactory will always provide `SchemaRecordSerializerThrowing`, which is not what we want, @devhritwik could you please verify this?

kafka-rest/src/test/java/io/confluent/kafkarest/config/SingletonSerializerTest.java
```diff
@@ -0,0 +1,69 @@
...
+            install(new ControllersModule());
+          }
+        });
+  }
+
+  @Test
+  public void testSchemaRecordSerializerSingleton() {
+    SchemaRecordSerializer firstInstance = serviceLocator.getService(SchemaRecordSerializer.class);
+    SchemaRecordSerializer secondInstance = serviceLocator.getService(SchemaRecordSerializer.class);
+
```
trnguyencflt, line 65: could you please assert if the instance is not of SchemaRecordSerializerThrowing class?

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
devhritwik: I think then we would need to change some logic here. 

> suppose for some reason that schemaRegistryClient is empty

SchemaRegistryClient is provided by the `BackendsModule` [here](https://github.com/confluentinc/kafka-rest/blob/555a5849104bd67c1be67bf0d2e764f72ffb50b3/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java#L163). Now if the schemaRegistryClient, we will bind the serializer to `SchemaRecordSerializerThrowing`. However, this registration of schemaRegistryClient happens at the start of the application. 
Can this be empty and then later be initialized? If true, then we cannot keep this as singleton. However it also means that if this schemaregistryclient can keep changing, then we can never keep the serializer as a singleton and this optimisation should be discarded totally. 

I would need your help in understanding this. 

The implementation I was thinking of was to bind `SchemaRecordSerializer` as a Singleton only when SchemaRegistryClient is present. However, if the schemaRegistryClient can keep changing, then even this binding is not correct.

devhritwik: I think we can do some code clean up, all this class `SchemaRecordSerializerThrowing` does is throw an [error](https://github.com/confluentinc/kafka-rest/blob/5e12d55eb3f203f159ce038c77ddf740582c4380/kafka-rest/src/main/java/io/confluent/kafkarest/controllers/SchemaRecordSerializerThrowing.java#L25). It doesn't even create an object. Hence instead of this class, why don't we just throw an error at the provide level? The schemaRegistryClient not being present and serializer being asked to serialize is a bad state for our application and hence we can just delete this class and throw the error when asked to serialize

devhritwik: I have added a new commit where I have removed `SchemaRecordSerializerThrowing` class altogether, simplifying this further in the process

trnguyencflt: > SchemaRegistryClient is provided by the BackendsModule [here](https://github.com/confluentinc/kafka-rest/blob/555a5849104bd67c1be67bf0d2e764f72ffb50b3/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java#L163). Now if the schemaRegistryClient, we will bind the serializer to SchemaRecordSerializerThrowing. However, this registration of schemaRegistryClient happens at the start of the application.
Can this be empty and then later be initialized? If true, then we cannot keep this as singleton. However it also means that if this schemaregistryclient can keep changing, then we can never keep the serializer as a singleton and this optimisation should be discarded totally.

I don't think this is possible in the context of Confluent Platform, where kafka-rest is fixed with a single SchemaRegistry server for its lifetime, so keeping this as Singleton is good for now.

For Confluent cloud, we don't have SchemaRegistry support yet but the cardinatity of that might not be singleton, it is also the reason that we have SchemaRecordSerializerThrowing class.

Please add a comment to the singleton part, noting the assumption that SchemaRegistry is initiated at application start, so we will know how to deal with this later if we ever support schema registry in Confluent cloud.

Please also revert the change related to removing SchemaRecordSerializerThrowing, let's keep this change only about singleton SchemaRecordSerializer.

devhritwik: I will revert the change related to remove the Throwing class.

> it is also the reason that we have SchemaRecordSerializerThrowing class.

I wanted to understand. What is the role of the ‘Throwing’ class if all it does is throw an error in its initialization?
When we support SchemaRegistry for ConfluentCloud, we can reintroduce this class, or any other mechanism?

devhritwik: The Throwing class just has this in its block 

```
Throw Errors.messageSerializationException(
        "Schema Registry not defined, no Schema "
            + "Registry client available to serialize message.");
```
I think it is made with keeping future integration with Confluent Cloud SR? However, I think we can keep the current code change for now, and add this back when needed? 

Whatever you suggest, I have the revert ready if you want me to

kafka-rest/src/test/java/io/confluent/kafkarest/config/SingletonSerializerTest.java
devhritwik: Added this UT

kafka-rest/src/main/java/io/confluent/kafkarest/controllers/ControllersModule.java
trnguyencflt: Let's revert the change because it is safer to now changing many behaviour in one PR, about SchemaRecordSerializerThrowing, we will access the impact later.

devhritwik: Okay, I have reverted the changes and added a comment on top of the Singleton binding for future users.

kafka-rest/src/test/java/io/confluent/kafkarest/config/SingletonSerializerTest.java
```diff
@@ -0,0 +1,83 @@
...
+            bind(dummyConfig)
+                .qualifiedBy(new ConfigModule.JsonschemaSerializerConfigsImpl())
+                .to(new TypeLiteral<Map<String, Object>>() {});
+            bind(dummyConfig)
+                .qualifiedBy(new ConfigModule.ProtobufSerializerConfigsImpl())
+                .to(new TypeLiteral<Map<String, Object>>() {});
+
+            install(new ControllersModule());
+          }
+        });
```
trnguyencflt, line 65: [minor] please add this check

```suggestion
EasyMock.verify(mockSchemaRegistryClient);
```

kafka-rest/src/test/java/io/confluent/kafkarest/config/SingletonSerializerTest.java
```diff
@@ -0,0 +1,84 @@
...
+        serviceLocator,
+        new AbstractBinder() {
+          @Override
+          protected void configure() {
+            bind(mockSchemaRegistryClient).to(SchemaRegistryClient.class);
+            Map<String, Object> dummyConfig = new HashMap<>();
+            dummyConfig.put("schema.registry.url", "http://localhost:8081");
+
+            bind(dummyConfig)
+                .qualifiedBy(new ConfigModule.AvroSerializerConfigsImpl())
```
trnguyencflt, line 54: instead of using [Avro|Jsonschema|Protobuf]SerializerConfigsImpl classes which will require us to make change to ConfigModule, we can just create these three classes in test (as private static classes), after that you can move this file to the same package with ControllersModule, we can then avoid the change to make SchemaRecordSerializer and SchemaRecordSerializerThrowing public class.

devhritwik: That makes sense, I can do that yes

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/ControllersModuleTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.glassfish.hk2.api.AnnotationLiteral;
+import org.glassfish.hk2.api.ServiceLocator;
+import org.glassfish.hk2.api.TypeLiteral;
+import org.glassfish.hk2.utilities.ServiceLocatorUtilities;
+import org.glassfish.hk2.utilities.binding.AbstractBinder;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class ControllersModuleTest {
+  private static final class AvroSerializerConfigsImpl
```
trnguyencflt, line 22: please move theses to the end of the this test class, and we should probably name them with prefix `Test`

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/ControllersModuleTest.java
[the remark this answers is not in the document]
devhritwik: Got it, will do!

kafka-rest/src/test/java/io/confluent/kafkarest/controllers/ControllersModuleTest.java
devhritwik: Okay!

PR #1359, opened 2025-03-21
Handle null payload in POST requests

trnguyencflt:
Even though we annotate the payload with `@Valid` annotation, we might still receive null payload, this will cause the endpoint to return 500 due to NPE of the payload, this PR fixes the issue by checking the nullability of the payload and return 422 if that is the case.

The fix in this PR is for v3 resources.


aaddepalli7:
Thanks for the PR, LGTM!

aaddepalli7 (approved):

PR #1368, opened 2025-04-01
KNET-18079: Migrate reload4j (log4j1) to log4j2

shubh-ranade:
- AK upgraded to log4j2: https://github.com/apache/kafka/pull/17373
- log4j2.yaml files were generated by first converting log4j.properties to XML using [Apache log4j migration tool](https://logging.apache.org/log4j/2.x/migrate-from-log4j1.html#Log4j2ConfigurationFormat), and then converting xml to yaml
- Added an explicit dependency following https://logging.apache.org/log4j/2.x/manual/configuration.html#configuration-factories

PR #1376, opened 2025-04-14
MINOR: Fix log4j2 config property

shubh-ranade:
Use log4j2 property name. 

Follow up of #1368

### DIFF
--- a/bin/kafka-rest-run-class
+++ b/bin/kafka-rest-run-class
@@ -31,11 +31,11 @@ if [ "x$KAFKAREST_LOG4J_OPTS" = "x" ]; then
   # Test for files from dev -> packages so this will work as expected in dev if you have packages
   # installed
   if [ -e "$base_dir/config/log4j2.yaml" ]; then # Dev environment
-    KAFKAREST_LOG4J_OPTS="-Dlog4j.configuration=file:$base_dir/config/log4j2.yaml"
+    KAFKAREST_LOG4J_OPTS="-Dlog4j2.configurationFile=file:$base_dir/config/log4j2.yaml"
   elif [ -e "$base_dir/etc/kafka-rest/log4j2.yaml" ]; then # Simple zip file layout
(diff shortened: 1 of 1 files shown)


bin/kafka-rest-run-class
```diff
@@ -31,11 +31,11 @@ if [ "x$KAFKAREST_LOG4J_OPTS" = "x" ]; then
   # Test for files from dev -> packages so this will work as expected in dev if you have packages
   # installed
   if [ -e "$base_dir/config/log4j2.yaml" ]; then # Dev environment
-    KAFKAREST_LOG4J_OPTS="-Dlog4j.configuration=file:$base_dir/config/log4j2.yaml"
+    KAFKAREST_LOG4J_OPTS="-Dlog4j2.configurationFile=file:$base_dir/config/log4j2.yaml"
```
omkreddy, line 34: "file:" is not needed.

omkreddy (commented):
Pls raise the PR for 8.0.x branch.

bin/kafka-rest-run-class
devhritwik, line 34: Should we change the variable here and above to either `KAFKAREST_LOG4J2_OPTS` or use `logger` for naming consistency? Also `# logj4 settings` the comment above should read log4j2/logger settings.

shubh-ranade: AK uses `KAFKA_LOG4J_OPTS` and we're already using log4j logging. Leaving this as `KAFKAREST_LOG4J_OPTS` isn't too bad imo.

shubh-ranade: done

omkreddy (approved):

devhritwik (approved):

shubh-ranade:
Verified that logs were saved to `kafka-rest.log` when running kafka-rest server locally.

devhritwik (approved):

PR #1382, opened 2025-04-21
Remove zk notice for cp 8.0.x release

silvanli-confluent:


shubh-ranade:
Should we also remove zookeeper references in https://github.com/confluentinc/kafka-rest/blob/8.0.x/NOTICE and https://github.com/confluentinc/kafka-rest/blob/8.0.x/licenses/NOTICE.confluent-common.txt ?

PR #1394, opened 2025-05-28
[KNET-5703] remove is_internal field

xuliang2019:
For CCloud, internal topics are not exposed. Thus `is_internal` will always be false. It should be removed for (List|Get|Create) api calls.

This code change is covered by local test and integration tests:
```
curl -X GET http://localhost:8082/v3/clusters/5L6g3nShT-eMCtK--X86sw/topics

{"kind":"KafkaTopicList","metadata":{"self":"http://localhost:8082/v3/clusters/5L6g3nShT-eMCtK--X86sw/topics","next":null},
"data":[{"kind":"KafkaTopic","metadata":{"self":"http://localhost:8082/v3/clusters/5L6g3nShT-eMCtK--X86sw/topics/quickstart-events","resource_name":"crn:///kafka=5L6g3nShT-eMCtK--X86sw/topic=quickstart-events"},"cluster_id":"5L6g3nShT-eMCtK--X86sw","topic_name":"quickstart-events","replication_factor":1,"partitions_count":1,"partitions":{"related":"http://localhost:8082/v3/clusters/5L6g3nShT-eMCtK--X86sw/topics/quickstart-events/partitions"},"configs":{"related":"http://localhost:8082/v3/clusters/5L6g3nShT-eMCtK--X86sw/topics/quickstart-events/configs"},"partition_reassignments":{"related":"http://localhost:8082/v3/clusters/5L6g3nShT-eMCtK--X86sw/topics/quickstart-events/partitions/-/reassignment"},"authorized_operations":[]}]}
```

### DIFF
--- a/api/v3/openapi.yaml
+++ b/api/v3/openapi.yaml
@@ -2343,7 +2343,6 @@ components:
           required:
             - cluster_id
             - topic_name
-            - is_internal
             - replication_factor
             - partitions_count
             - partitions
@@ -2354,8 +2353,6 @@ components:
               type: string
             topic_name:
               type: string
-            is_internal:
-              type: boolean
             replication_factor:
               type: integer
             partitions_count:
@@ -2591,7 +2588,6 @@ components:
               resource_name: 'crn:///kafka=cluster-1/topic=topic-X'
             cluster_id: 'cluster-1'
             topic_name: 'topic-X'
-            is_internal: false
             replication_factor: 3
             partitions_count: 1
             partitions:
@@ -2953,7 +2949,6 @@ components:
               resource_name: 'crn:///kafka=cluster-1/topic=topic-1'
             cluster_id: 'cluster-1'
             topic_name: 'topic-1'
-            is_internal: false
             replication_factor: 3
             partitions_count: 1
             partitions:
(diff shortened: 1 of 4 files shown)


trnguyencflt:
@xuliang2019 could you please explain more the motivation of this change?

kafka-rest is open source and used for both CCloud and CP, if this is only about CCloud, then the change should be in ce-kafka-rest.

xuliang2019:
> @xuliang2019 could you please explain more the motivation of this change?
> 
> kafka-rest is open source and used for both CCloud and CP, if this is only about CCloud, then the change should be in ce-kafka-rest.

@trnguyencflt Thanks for the info. Will move the change to the `ce-kafka-rest`. One more question, I don't see the `TopicData.java` in `ce-kafka-rest`, does it mean that I need to copy it from `kafka-rest` first then do the modification?

trnguyencflt:
> > @xuliang2019 could you please explain more the motivation of this change?
> > kafka-rest is open source and used for both CCloud and CP, if this is only about CCloud, then the change should be in ce-kafka-rest.
> 
> @trnguyencflt Thanks for the info. Will move the change to the `ce-kafka-rest`. One more question, I don't see the `TopicData.java` in `ce-kafka-rest`, does it mean that I need to copy it from `kafka-rest` first then do the modification?

Unfortunately, I don't see other way to do this in ce-kafka-rest rather than copying TopicsResource (for the GET requests) and TopicData.

Other way is e can create a config in KafkaRestConfig to hide `is_internal` in response in `kafka-rest` and will set this in CCloud. 

Is there a real customer ask for this or this is just a QoL change?

